PR #1157, opened 2014-12-24
Made logging amount configurable

TagForce:
Added Log file number and max logfile size to settings - General page


fernandog:
@TagForce please re-commit using the develop files. The logger.py that you used is from master and pastebin library it's not used anymore. Dev won't be able to accept this commit.

TagForce:
I've changed it a bit, since changing the setting requires a restart.

New Pull Request: https://github.com/SiCKRAGETV/SickRage/pull/1173

PR #1162, opened 2014-12-30
Don't clean user files when GIT CLEAN

fernandog:
The "x" option ignores the .gitignore

-x
Don’t use the standard ignore rules read from .gitignore (per directory) and $GIT_DIR/info/exclude, but do still use the ignore rules given with -e options. This allows removing all untracked files, including build products. This can be used (possibly in conjunction with git reset) to create a pristine working directory to test a clean build.

http://git-scm.com/docs/git-clean


Hellowlol:
You can use the -e option to exclude what git shouldnt cleanup (like userdata etc) this way thrash like pyc will be removed.

PR #1166, opened 2015-01-02
Add Custom API notifier

Heisenberg74:
Custom API allow users to use their own API for notifications. Just
enable this option in Notifications > Devices and set the URL of your
API without forget to set parameter %message%


echel0n:
Please fix this so we can merge the PR, thanks.

PR #1168, opened 2015-01-03
Implementation for sickragetv/sickrage-issues#322 - Add extra progress-bar color for when show is fully downloaded

sn3akz:
Added a different shade of green for when and only when the show is 100% downloaded

### DIFF
--- a/gui/slick/interfaces/default/home.tmpl
+++ b/gui/slick/interfaces/default/home.tmpl
@@ -364,7 +364,10 @@ $myShowList.sort(lambda x, y: cmp(x.name, y.name))
 					}
 					if (classvalue>80) {
 						classtoadd = "progress-80"
-					}		
+					}
+					if (classvalue==100) {
+						classtoadd = "progress-100"
+					}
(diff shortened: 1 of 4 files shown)


gui/slick/interfaces/default/home.tmpl
```diff
@@ -364,7 +364,10 @@ $myShowList.sort(lambda x, y: cmp(x.name, y.name))
 					}
 					if (classvalue>80) {
```
fernandog: @sn3akz 
Here should be:
if (classvalue>80 && classvalue<100) {

sn3akz: Could be, but doesn't do much. In my code, the classtoadd variable just gets overridden when it's a 100. Nothing wrong with that.

The old code should indeed be >=. When it's 80, no color is selected. Maybe that's also the cause of issue sickrage-issues/#303

fernandog:
@sn3akz can you update the code and test the issue 303?

sn3akz:
Will do

PR #1171, opened 2015-01-06
Auto send user and password when open managetorrents page

fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/315

### DIFF
--- a/sickbeard/webserve.py
+++ b/sickbeard/webserve.py
@@ -3172,6 +3172,9 @@ class Manage(Home, WebRoot):
             else:
                 t.info_download_station = '<p>To have a better experience please set the Download Station alias as <code>download</code>, you can check this setting in the Synology DSM <b>Control Panel</b> > <b>Application Portal</b>. Make sure you allow DSM to be embedded with iFrames too in <b>Control Panel</b> > <b>DSM Settings</b> > <b>Security</b>.</p><br/><p>There is more information about this available <a href="https://github.com/midgetspy/Sick-Beard/pull/338">here</a>.</p><br/>'
 
+        if not sickbeard.TORRENT_PASSWORD == "" and not sickbeard.TORRENT_USERNAME == "":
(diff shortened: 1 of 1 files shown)


fernandog:
@abeloin is there a better way to send the password without keep it on page code?

abeloin:
@fernandog None that I'm aware of.

Hellowlol:
Isnt there a redirect you can use at the username and password is hardcoded into the url?

abeloin:
Even if you use a sickrage url then use a 302 to redirect to the final page it will still show in the browser's network developer tools.

Maybe with a tornado handler acting as a proxy could work but my knowledge of Tornado is far to little for me to answer.

fernandog:
what about a javascript function to remove/replace the url with user:pass using regex?

abeloin:
It will still show when using the show source code as the page is pulled from the server directly.

Not to mention, if you use wget or curl.

Hellowlol:
i was thinking something like this:

http://tornado.readthedocs.org/en/latest/guide/structure.html#redirection

app = tornado.web.Application([
    url(r"/app", tornado.web.RedirectHandler,
        dict(url="http://itunes.apple.com/my-app-id")),
    ])

fernandog:
@Hellowlol where do we add the user/pass?

abeloin:
@Hellowlol It will still send a 302 to the browser with the new link(e.g. http://user:pass@deluge.local) and the username and password will be shown in the browser's developer tools.

I think the user have accept the fact that if they want use this feature they better use HTTPS to secure their connection.

The best would if those webui use something like oauth or openid were we could request a token an open the seesion.

Hellowlol:
Yeah, fuck it. :P

fernandog:
=) this is a feature from mr_orange. Don't know if users really use it. I prefer use Transdroid on my phone to see the torrent downloader

Hellowlol:
I that app available for ios aswell? I just use https://github.com/Hellowlol/HTPC-Manager for stuff like that.

abeloin:
Call me old fashion, I open a new tab and type in the address to get access. ;)

In the end, it's boil down to security vs user friendliness, the more secure the less friendly it is and vice versa.

PR #1178, opened 2015-01-10
Fix issue showing 'unknown dir' when adding shows

fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/286


renini:
Nice! @echel0n still on a well deserved vacation?

PR #1182, opened 2015-01-13
Fix for https://github.com/SiCKRAGETV/sickrage-issues/issues/178

lucianot54:
Problem with the user-agent. I logged with python User-Agent and download torrent with SickRage User-Agent. I think t411 have migrated to cloudflare. When you're logged with a different User-Agent, you have a different authKey in the cookie and 2 request with 2 User-Agent don't work.


titoko:
Works for me !

t0per:
It works perfectly !

Thanks for your work ;)

Zabie:
It works for me too !

PR #1185, opened 2015-01-14
Adding a download field for the episodes

Ulrar:
Added config option download_url
Added field to configure download_url in general --> interface
Added link to download episodes in the displayShow template, using download_url

### DIFF
--- a/gui/slick/interfaces/default/config_general.tmpl
+++ b/gui/slick/interfaces/default/config_general.tmpl
@@ -301,6 +301,18 @@
 							</label>
 						</div>
 
+						<div class="field-pair">
+							<label for="download_url">
+						    		<span class="component-title">Download url</span>
+						    		<input type="text" name="download_url" id="download_url" value="$sickbeard.DOWNLOAD_URL" size="35" />
+						    	</label>
+						    	<label>
+						    		<span class="component-title">&nbsp;</span>
+								<span class="component-desc">URL where the shows can be downloaded.</span>
+						    	</label>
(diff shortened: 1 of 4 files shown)


fernandog:
@Ulrar i didn't understand. Can you explain better please?

Ulrar:
It adds a config option where you can set a base url, and then in the displayShow page it adds a columns with a download link generated from that (append the episode name to the base url).
It allow to easily retrieve the episodes from anywhere without having to use ftp or similar. Very useful when sickbeard is hosted on a dedicated server outside the home.

fernandog:
@Ulrar I see. le'ts wait the DEV to analyse this. Have you fully tested?

Ulrar:
I have been using this for a year now on the mr orange sickbeard, just decided to switch to this one this morning. So far looks like it works about the same, I don't see errors related to this.
If the idea is okay I can probably improve it, I have pieces of code to allow the download of subtitles aswell by cliking on the flag.

fernandog:
@Ulrar will it work with reverse proxy? what about accessing SR from a dyndns? how do you get the public ip?

Ulrar:
That's the point of the config option, I don't.
The user set the base url where the shows are available himself.
Personally I have a separate vhost for the episodes.

fernandog:
I see. Nice feature.

Ulrar:
Just noticed, it's not the correct branch. My bad, I'll reopen that correctly in a minute.

PR #1186, opened 2015-01-14
Adding a download field for the episodes

Ulrar:
Added config option download_url
Added field to configure download_url in general --> interface
Added link to download episodes in the displayShow template, using download_url

See #1185 for discussions.

### DIFF
--- a/gui/slick/interfaces/default/displayShow.tmpl
+++ b/gui/slick/interfaces/default/displayShow.tmpl
@@ -6,6 +6,7 @@
 #from lib import subliminal
 #import os.path, os
 #import datetime
+#import urllib
 
 #set global $title=$show.name
 ##set global $header = '<a></a>' % 
@@ -376,6 +377,7 @@
 			#end if
 				<th class="col-name">Name</th>
 				<th class="col-airdate">Airdate</th>
+				<th class="col-ep">Download</th>
 			#if $sickbeard.USE_SUBTITLES and $show.subtitles:
 				<th class="col-subtitles">Subtitles</th>
 			#end if
@@ -469,6 +471,25 @@
 		
 		<td class="col-airdate">
 			<span class="${fuzzydate}">#if int($epResult['airdate']) == 1 then 'never' else $sbdatetime.sbdatetime.sbfdate($sbdatetime.sbdatetime.convert_to_setting($network_timezones.parse_date_time($epResult['airdate'],$show.airs,$show.network)))#</span>
+		</td>
+
+		<td>
+		#if len($sickbeard.DOWNLOAD_URL) > 0 and len($epResult['location']) > 0
+			#set $filename = $epResult['location']
+		  	#for $rootDir in $sickbeard.ROOT_DIRS.split('|')
+		   		#if $rootDir.startswith('/')
+		   			#set $filename = $filename.replace($rootDir, "")
+		   		#end if
+		   	#end for
+			#set $filename = $sickbeard.DOWNLOAD_URL + $urllib.quote($filename.encode('utf8'))
+			<center><a href="$filename">Download</a></center>
+		#else
+			#if $epLoc and $show._location and $epLoc.lower().startswith($show._location.lower()):
+				$epLoc[len($show._location)+1:]
+			#elif $epLoc and (not $epLoc.lower().startswith($show._location.lower()) or not $show._location):
+		        	$epLoc
+			#end if
+		#end if
 		</td>
 		
 		#if $sickbeard.USE_SUBTITLES and $show.subtitles:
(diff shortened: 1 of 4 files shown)


gui/slick/interfaces/default/displayShow.tmpl
```diff
@@ -376,6 +377,9 @@
 			#end if
 				<th class="col-name">Name</th>
 				<th class="col-airdate">Airdate</th>
+			#if len($sickbeard.DOWNLOAD_URL) > 0
+				<th class="col-ep">Download</th>
+			#end if
 			#if $sickbeard.USE_SUBTITLES and $show.subtitles:
```
fernandog: if $sickbeard.DOWNLOAD_URL

gui/slick/interfaces/default/displayShow.tmpl
```diff
@@ -470,6 +474,21 @@
 		<td class="col-airdate">
 			<span class="${fuzzydate}">#if int($epResult['airdate']) == 1 then 'never' else $sbdatetime.sbdatetime.sbfdate($sbdatetime.sbdatetime.convert_to_setting($network_timezones.parse_date_time($epResult['airdate'],$show.airs,$show.network)))#</span>
 		</td>
+
+		#if len($sickbeard.DOWNLOAD_URL) > 0 and len($epResult['location']) > 0
```
fernandog: #if $sickbeard.DOWNLOAD_URL and $epResult['location']

fernandog:
@Ulrar you should edit the field hint to user add the "http://" and etc

it's not working for me:
http://fernandog.url-removed/Show/Season%201/Show.S01E01-ECI.mkv

ERR_CONNECTION_TIMED_OUT

should I add any port to my router?

Ulrar:
You need a separate vhost for the download, pointing to your sickbeard root dir.
For example if your rootdir is /mnt/shows, a nginx configuration would look like this : 

```
server
{
        listen                          [::]:443 ssl spdy;
        listen                          443 ssl;
        server_name                     my_shows.domain.tld;

        root                            /mnt/shows/;
}
```
I guess it could be possible to use the same vhost as sickbeard by adding some code to serve the files but I don't know how sickbeard works well enough to implement that as of yet.

I'm currently thinking of a way to support multiple rootdir configurations, for the time being I'm using some trickery in my vhost to handle it but it'd be nice to be able to put a different download url for each root dir. I have rootdirs on multiple servers using sshfs and ftpfs, it's fine for storing the files once but downloading through that is horribly slow, I think it's better to be able to generate different url for each rootdir, allowing the user to download the file directly from the server storing it rather than through slowish network filesystems. Don't know if I'm clear, I'm sorry, english isn't my native language.

fernandog:
@Ulrar isn't this too complicated to a common user? He would expect just enabled this setting and start download. 
I think this setting could be hidden from UI and become a config.ini setting only.
What do you think?

Ulrar:
If you know how to configure a reverse proxy I'd think you can configure that, but I supposed some people use sickbeard without a reverse proxy ..
I'm not opposed to remove it from the gui. Actually makes it easier to support the multi rootdir I had in mind :)

fernandog:
@Ulrar or you can make a tutorial and the DEV could post here in GIT so others can use this feature.  Let's see what @echel0n thinks of it.

fernandog:
@Ulrar can you make a wiki (create as an issue) on how to make this work?

also, for this to work is it necessary to enable reverse proxy?
if so, we need to change .tmpl code to only enable "Download url" input field if reverse proxy is enabled

abeloin:
@fernandog No, you shouldn't have to use reverse proxy for serving files

fernandog:
@abeloin at least we need to add a note pointing to a Wiki explaining how to make it work

Ulrar:
Sure, I will make an entry for that during the week :). Added on my todo !
You don't need to reverse proxy or anything since sickbeard isn't actually
serving the files itself, it's just a base url where the files are
available to be served by whatever you want, probably apache or nginx
directly.

2015-03-23 19:06 GMT+01:00 Alexandre Beloin <notifications@github.com>:

> @fernandog <https://github.com/fernandog> No, you shouldn't have to use
> reverse proxy for serving files
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/SiCKRAGETV/SickRage/pull/1186#issuecomment-85126888>.
>



-- 
Kévin Lemonnier

abeloin:
@fernandog Sorry for the confusion, I was not replying to the wiki part but for the reverse proxy question.

fernandog:
@Ulrar create a wiki using gist.github.com and report back. I will add in the wiki and edit the .tmpl code pointing to the wiki

Ulrar:
Hi,

Completely forgot about that, sorry.
Here is the gist : https://gist.github.com/Ulrar/7434eb6c45e09c5ad0f5
Don't hesitate to tell me if I'm not clear or just point out spelling or syntax mistakes, English isn't my native language :)

PR #1190, opened 2015-01-15
Added support for Plex Media Center updates with Auth Token.

vandalon:
This because PMS does not allow connections without token when using Plex Home

### DIFF
--- a/gui/slick/interfaces/default/config_notifications.tmpl
+++ b/gui/slick/interfaces/default/config_notifications.tmpl
@@ -221,6 +221,20 @@
                                 </label>
                             </div>
                             <div class="field-pair">
+                                <label for="plex_server_token">
+                                    <span class="component-title">Plex Media Server Auth Token</span>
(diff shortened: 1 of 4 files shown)


Hellowlol:
Why not fetch the auth token automatically for the user? See https://github.com/Hellowlol/HTPC-Manager/blob/master2/modules/plex.py#L615 for a example

vandalon:
I will add a way to fetch the auth token later. I don't like storing user/pass if not needed. I'll build something to fetch the auth token once via a web-gui and store the auth-token in the config.

Hellowlol:
Well storing username and password is the same as storing the auth token/api keys in my eyes, this pr will have users jump thru hoops to get it working when the username and password option already is in the config. But then again im not making this pr. Thanks for solving the issue for plexhome users. :)

vandalon:
user tent to use the same/user pass for different sites/places. This token is only unique to plex :)

PR #1191, opened 2015-01-15
Remove "aoeu" div from the default "add new shows" template

fl0yd:
first commit removed the div, the second removed a blank line that I should have deleted with the first commit.

### DIFF
--- a/gui/slick/interfaces/default/home_newShow.tmpl
+++ b/gui/slick/interfaces/default/home_newShow.tmpl
@@ -32,7 +32,6 @@
 
 	<div id="core-component-group1" class="tab-pane active component-group">
 				
-
 	<form id="addShowForm" method="post" action="$sbRoot/home/addShows/addNewShow" accept-charset="utf-8">
 
 	<fieldset class="sectionwrap">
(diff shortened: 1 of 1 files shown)


fernandog:
@fl0yd I think this div is used to show other information like the directory of the new created tv-show

fl0yd:
If that's the case, I could at least remove the "aoeu" from the page

fernandog:
@fl0yd can you test adding a show and check if the div is used? I'm on mobile.

fl0yd:
The test succeeded without issue, below is a screen grab of the tab after the edit.

image

fernandog:
I see. bur try this:
select the parent folder, typing a tv-show and the DIV will be filled with the information

abeloin:
@fl0yd Pull request #1192 should fix your issue without having to remove the "aoeu"

fernandog:
this DIV must be hidden until choosing the tv-show. Then will show:

"Adding show Show-name into /path/to/folder/Show-name/"

fl0yd:
Thanks!

PR #1200, opened 2015-01-19
Fix Safari next show button not working.

abeloin:
See SiCKRAGETV/sickrage-issues#436 for more info


udochrist:
This fixes the issue with the "next show" button

fernandog:
@abeloin 

all .attr must be converted to .prop or only this one?
I ssh inside JS directorty and there's lots of files with .attr"

```
grep -Ril ").attr("
```

abeloin:
@fernandog Only those two, in this case.

PR #1218, opened 2015-01-21
Optional custom transmission RPC URL

grrr2:
Add option to specify a custom transmission client's RPC URL.


fernandog:
@grrr2 will this commit break this:
https://github.com/SiCKRAGETV/SickRage/pull/1171

grrr2:
Most likely not. If you follow the recommended host setup (ie. http://<address>:<port>) manageTorrents will directly connect to this configured port which is ok as in this case the transmission instance itself will redirect to http://<address>:<port>/<rpcurl>/web .

grrr2:
Found some problems till it fixed pull request is closed. Thanks.

fernandog:
@grrr2 any news on this one? please report your problems. we can help.

sickbeard/__init__.py
```diff
@@ -286,6 +286,7 @@
 TORRENT_LABEL = ''
 TORRENT_LABEL_ANIME = ''
 TORRENT_VERIFY_CERT = False
+TORRENT_RPCURL = '/transmission"
```
fernandog: @grrr2 I think you shouldn't hardcode this. But set as "transmission" as the default of "torrent_rpcurl" in the UI

fernandog:
@grrr2 I don't think there are errors on this. Also don't affect the https://github.com/SiCKRAGETV/SickRage/pull/1171

@abeloin what do you think?

abeloin:
@fernandog At first glance, it's missing torrent_rpcurl=None in saveSearch method parameters definition.
Let me check the code further.
Also, I don't use transmission so I won't be able to test if it's connecting properly.

fernandog:
@RackerJack can you test it when we ask you?

RackerJack:
@fernandog sure I'm on my way home at the moment

fernandog:
@RackerJack not now =) maybe in a day or two or three =)

RackerJack:
haha ok np ill keep an eye out :)

abeloin:
@fernandog Here the commit with the updated code https://github.com/abeloin/SickRage/commit/37ab43adbc32f5b4dbd892ff29bab5defad8eb14

@RackerJack Can you test the commit?

If everything is ok, I'll submit a pull request.

fernandog:
@RackerJack can you also confirm that https://github.com/SiCKRAGETV/SickRage/pull/1171 is still working? 
do you use password to access transmission?

RackerJack:
@abeloin i have made the changes from your fork but still shows 
`Error: Unable to get Transmission Authentication, check your config!`
`2015-01-26 13:17:40 SEARCHQUEUE-MANUAL-194031 :: Transmission: Authentication Failed`
@fernandog i do not use a password to access transmission

abeloin:
@RackerJack Can you try changing the rpc url from ```/transmission``` to ```transmission```

RackerJack:
@abeloin unfortunately this would break other functionality on the server im unable to do this

i run a nodejs app that uses the transmission rpc and in all my wiseness i rote the routes to use /torrent https://developer.rackerjack.uk/rackerjack/blackc.at/src/master/routes/index.js  

in settings.json on transmission it has `"rpc-url": "/torrent/", ` 

i haven't encountered any issues using transmission with couchpotato with this custom rpc

abeloin:
@RackerJack I meant in Sickrage

As looking at the original code:
```
self.url = self.host + 'transmission/rpc'
```
the / is already included in the host and this is now with the patch:
```
self.url = self.host + sickbeard.TORRENT_RPCURL + '/rpc'
```
so it should only contain the path without any /, in your case: ```torrent```

abeloin:
If it still doesn't work, try adding the following line into sickbeard/logger.py line 72 with other getLogger:
```
logging.getLogger("requests.packages.urllib3"),
```
This should add in your log the requests made to the server.

RackerJack:
@abeloin IT WORKS!! 

I changed 

Transmission host:port
to `http://127.0.0.1:9091`  removed the trailing slash
and Torrent RPC URL to `torrent` and it now shows `Success: Connected and Authenticated`

thank you so much for your help !!

fernandog:
nice!

abeloin:
Good I'll make the changes in the code and submit a pull request

fernandog:
Isnt possible to use os.path.join so does not matter if user insert the slash or not?

abeloin:
We could use urljoin but it as some limitation, see http://stackoverflow.com/questions/10893374/python-confusions-with-urljoin for more info

fernandog:
or a simple if host.endswith("/") then host = host[:-1]

abeloin:
@RackerJack Can you retest with the last 3 commits of this branch: https://github.com/abeloin/SickRage/commits/patch-transmission_rpc

RackerJack:
@abeloin Hi these changes work even with
http://127.0.0.1:9091/ in the `Transmission host:port` section

thank you again

abeloin:
New pull request available: SiCKRAGETV/SickRage#1256

grrr2:
Whoa, you did what i've been working since found bug like slashes, just wanted to commit but you were quicker :-) Thanks.

PR #1219, opened 2015-01-22
Select All checkbox on Preview Rename page

TagForce:
- Added "Select All" check box under new "All Seasons" table
- Duplicated "Rename Selected" and "Cancel Rename" buttons to be
available underneath the "Select All" checkbox.

### DIFF
--- a/TODO.txt
+++ b/TODO.txt
@@ -12,7 +12,7 @@
 -Change the hardcoded global ignore words to optional
 
 2014-10-08
--Add login page for http auth as opposed to browser dialog box
+VAdd login page for http auth as opposed to browser dialog box
 
 2014-10-13
 -Fix broken backlog


gui/slick/interfaces/default/testRename.tmpl
```diff
@@ -43,13 +63,13 @@
     <thead>
         <tr class="seasonheader" id="season-$cur_ep_obj.season">
             <td colspan="4">
-                <br/>
+            	<br />
```
fernandog: is this right? space between br and / ?

TagForce: Yup. That was already there.

fernandog: no. your commit is adding the space.
-                <br/>
+            	<br />

PR #1224, opened 2015-01-22
Fix accent issue on system with locale like POSIX.

abeloin:
Network image télé-québec.png issue with locale like POSIX
See SiCKRAGETV/sickrage-issues#427 for more info


maxcore:
Thx merci :)

fernandog:
@abeloin should we merge this or force user to use UTF-8?

abeloin:
@fernandog I think we should merge it. Sadly, not all systems(mostly embedded) have the proper locale. Ultimately the way network logos works right now should be changed to a table lookup that retrieve the proper logo as the current way can't cope with all the languages(e.g: Mandarin, Russian, ...)

fernandog:
Ok!

PR #1240, opened 2015-01-25
Ignore partially synced files from BTSync 2.0 Alpha

Tawmmm:
Ignore partially synced files from BTSync 2.0 Alpha

### DIFF
--- a/sickbeard/helpers.py
+++ b/sickbeard/helpers.py
@@ -127,7 +127,7 @@ def replaceExtension(filename, newExt):
 
 def isSyncFile(filename):
     extension = filename.rpartition(".")[2].lower()
-    if extension == '!sync' or extension == 'lftp-pget-status' or extension == 'part':
+    if extension == '!sync' or extension == 'lftp-pget-status' or extension == 'part' or extension == 'bts':
         return True
     else:
         return False


fernandog:
@Tawmmm "!sync" its not used anymore?

Tawmmm:
Nope, it uses '.bts' now, atleast the 2 alpha versions currently available do.

fernandog:
@Tawmmm so when alpha is "release" we can remove it. Can you do that when it happens?

TagForce:
@fernandog @Tawmmm Maybe we should make this list editable by the user?
Seeing as it is now hardcoded, and there's bound to be changes to the list that we need to exclude it might be easier to just make the whole thing do it using a list saved in the config.ini which the user can change.

I'm not sure whether or not the .bts extension will be removed once it comes out of alpha, and it's not doing any harm, unless there's a media file with that extension in the future, so I'd leave it there.

Tawmmm:
@fernandog I intend to keep using BTSync, so if it changes from .bts after alpha I can modify it.

@TagForce Completely agree that a user-editable list would be the ideal, unfortunately this is beyond my skills at the moment!

TagForce:
@fernandog @Tawmmm When @echel0n comes back and gets around to merging all the PRs that are piling up, I'll work on making this list editable and saveable from the config files. Not going to do that right now as there's a million other changes being merged, and I don't want to create PRs that will give echel0n a lot of conflicts.

fernandog:
@TagForce hope all PR get's merged! that is a LOT. he will be surprised.

echel0n:
oh im back ;)

fernandog:
Nice! Welcome back. We tried to keep things together here!

fernandog:
@Tawmmm  can you make ignore hidden folders?

https://github.com/SiCKRAGETV/sickrage-issues/issues/554

PR #1244, opened 2015-01-26
Rtorrent client improvement

abeloin:
Changed the http transport adapter to a new one which add support to https and digest auth. The old one only allowed http connection and basic auth.


echel0n:
plz fix so we can merge

abeloin:
@echel0n Done

PR #1246, opened 2015-01-26
Show Filesize in Episode List

TagForce:
- Show the filesize in MB in the episode list if the episode has a
filesize (is Downloaded)

- Feature requested here: https://github.com/SiCKRAGETV/sickrage-issues/issues/487

### DIFF
--- a/gui/slick/interfaces/default/displayShow.tmpl
+++ b/gui/slick/interfaces/default/displayShow.tmpl
@@ -376,6 +376,7 @@
 				<th class="col-ep">Scene Absolute</th>
 			#end if
 				<th class="col-name">Name</th>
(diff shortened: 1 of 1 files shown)


fernandog:
@TagForce will this PR make any field wraped in not so large screens?

TagForce:
Ah, uhhh, yes... I'll fix that...
It wraps the Name field as it is...

fernandog:
@TagForce I would add  this as an option.

TagForce:
@fernandog Yeah, I'll add an option to the Interface page to show filesizes.
I'll also fix the wrapping. Crude way of doing it would be to add whitespace: nowrap; to the col-name class, but I'm trying to fix this in a nicer way.

PR #1247, opened 2015-01-26
Fix to #231 - TypeError: object of type 'NoneType' has no len()

fernandog:
deque "None" value verification was in the end of the code.

SiCKRAGETV/sickrage-issues#231

### DIFF
--- a/tornado/iostream.py
+++ b/tornado/iostream.py
@@ -1379,6 +1379,8 @@ def _merge_prefix(deque, size):
     >>> _merge_prefix(d, 100); print(d)
     deque(['abcdefghij'])
     """
+    if not deque:
+        deque.appendleft(b"")
     if len(deque) == 1 and len(deque[0]) <= size:
         return
     prefix = []
@@ -1395,9 +1397,6 @@ def _merge_prefix(deque, size):
     # so do the merge based on the type of data that's actually present.
     if prefix:
         deque.appendleft(type(prefix[0])().join(prefix))
-    if not deque:
-        deque.appendleft(b"")
-
 
 def doctests():
     import doctest


tornado/iostream.py
```diff
@@ -1379,6 +1379,8 @@ def _merge_prefix(deque, size):
     >>> _merge_prefix(d, 100); print(d)
     deque(['abcdefghij'])
     """
+    if not deque:
```
Hellowlol: Isnt this a bandaid and not a fix?

fernandog: if you look at the code without knowing the issue, Isn't wrong check if the variable is None in the end and not in the beginning? I'm not an expert but this could fix it.

Hellowlol:
Dont get me wrong, i dont think this is a bad solution. But Ech/someone needs to find the None floating around the the program :P

echel0n:
been away as i started up a new job and training hours are odd so will alot some more time for SR soon enough

fernandog:
@echel0n Do you think there is a better solution for this one? Despite the none floating the above code is wrong isnt?

fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/528#issuecomment-72376359

fernandog:
@abeloin what you think about this? regardless issue, this code is wrong isn't? the PR is ok?

Hellowlol:
It would be fine if the **kwargs to async_call was logged on errors. I think it would be a lot buster to debug, I have a button on "illegal" kwargs passed from the template

fernandog:
@Hellowlol didn't get this "I have a button on "illegal" kwargs passed from the template"

do you know how to "**kwargs to async_call ogged on errors" ?

Hellowlol:
Yeah, add it to the exception in async_call and too the writeerror function just to be sure. I'm on my phone to cba with line numbers

abeloin:
@fernandog  It's look like its failling with the redirect from showposter.

fernandog:
@abeloin is there something we can do like @Hellowlol to debug better and make a fix for this one?

abeloin:
Not much, the kwargs will show that it's not always the same poster failing it a weird corruption of the ioloop

fernandog:
@abeloin maybe related to a tv-shows with tvrage indexer (no poster, no banners?)

abeloin:
The banner exist

fernandog:
we can close this. cant we?

PR #1252, opened 2015-01-27
Add support for parsing TVRage specials

drglove:
https://github.com/SiCKRAGETV/sickrage-issues/issues/292

I've added support for parsing specials from the TVRage API. There are a few limitations due to the results returned from TVRage:
- Each episode listed has a season listed corresponding to the season during which the special was released
- There are no episode numbers associated with each episode

Right now I use the season information to put the special in the correct season, but I must come up with the episode number myself. I simply keep a counter for each season and increment it (starting from 1) every time a special is seen, and use that as the episode number. This is dependent upon the order in which TVRage supplies us the episodes, but I believe this is what they do themselves as I get the same results, which is independent of air date.
Because I use the season information, and I don't want to overwrite any episodes, I use 'S<special_num>' as the episode number. This isn't very desirable and I'd love to hear alternative ways to get around this, especially if there is some reason this must be an integer. I don't think stuffing all the episodes into Season 0 (i.e. Specials) is the way to go considering TVRage gives us season information, but I'm not picky on this point.

I've tested this on Top Gear, and get results agreeing with TVRage.

- [x] Pull metadata from TVRage
- [ ] Modify search behaviour for each provider to ignore episode number and search by name


drglove:
This currently won't work when fetching the episodes, I had stupidly forgotten that the episode number must be an integer to properly search for the show.
Any suggestions on how to handle missing episode numbers?

fernandog:
@drglove absolute numbering like thetvdb?

for example: http://thetvdb.com/?tab=season&seriesid=78804&seasonid=26260&lid=7

echel0n:
is this working ?>

drglove:
@echel0n It's not complete, no. Currently this just pulls the metadata from TVRage and adds an entry for each special into its corresponding season, but I have to make up my own episode numbers as this isn't provided by TVRage on specials. Until a proper episode number is found, I don't think that anything will be able to search it as is.

fernandog:
@drglove so we can't merge it?

drglove:
@fernandog Don't merge it, but I don't see any reason to close the pull request just because it's not ready to be merged.

dmarcelino:
Hi @drglove, I agree it shouldn't be closed and I suggest adding a tasks lists (example below) to the first post as it will add a progress bar to this pull request in the PR list view. That will give maintainers a better a idea of where the PR stands. Just my 2 cents :)

- [ ] task 1
- [x] task 2

PR List View example:
Example

drglove:
@dmarcelino I'm rather new to Github, thanks!
- [x] Pull metadata from TVRage
- [ ] Find way to generate unique special episode numbers

dmarcelino:
No worries @drglove. You'll have to edit the original PR message or it won't work :wink:

fernandog:
@drglove why don't we pull the specials info from thetvdb (and banners) and the rest of the data we get from tvrage?

fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/750

fernandog:
@drglove 

tvrage has it own numbering for specials. you can use the sequencial number below.

check this:
http://www.tvrage.com/DoctorWho_2005/episode_list/all

```
Special
#	Airdate	Title
S01 - #1	18/Nov/2005	Born Again : Children in Need	8
S01 - #2	25/Dec/2005	The Christmas Invasion	8.5
S02 - #3	25/Dec/2005	Attack of the Graske	9.5
S02 - #4	25/Dec/2006	The Runaway Bride	8.6
S03 - #5	30/Jun/2007	The Infinite Quest	9
S03 - #6	16/Nov/2007	Time Crash : Children In Need	8.7
S03 - #7	25/Dec/2007	Voyage of the Damned	9.4
S04 - #8	27/Jul/2008	Music Of The Spheres	8.4
S04 - #9	25/Dec/2008	Top 5 Christmas Moments	8.1
S04 - #10	25/Dec/2008	The Next Doctor	8.3
```

drglove:
@fernandog 
I'm effectively using the XML item number as the episode number right now as I increment a counter as I go through the results in order.
I'm putting the episode with its corresponding season instead of in "Season 0" so conflicts between the episode numbers can be a problem.
The problem with this approach is that I don't think any indexers will be using this numbering scheme, so you could probably have this sitting fine inside Sickrage's database, but you'd never be able to fetch an episode.

The more useful thing to do would probably to poll TVDB for specials exclusively, but this could confuse the end user when there's a mismatch between the two services, and then we don't have TVRage separated from TVDB anymore in the codebase.

fernandog:
@abeloin "probably to poll TVDB for specials exclusively" OR "use the XML item number as the episode number". what do you think?

I prefer use the xml item as the episode number

abeloin:
I would use the xml which resemble the solution I've been tinkering:

* TvRage Special are assigned S00E00 in the database.
* Internally special referred by there xml id(same as tvdb episode)
* Adding a regex to capture tvrage special only.
* Modify the code so that rescan, adding, removing episodes treat s00e00 as special when indexer is tvrage. 
* In post process, the file need to be rename to show s00e00 title ext to work. 
* Filename special must contain S00E00 or 0x0 or 0 to be detected as tvrage special by regex and must contain the episode name.
* Change displayShow page an other page if tvrage and season 0
* Modify the searching to search for show and title when indexer is tvrage and season 0.

As you can see it would involve a lot of modifications, but I think would represent a stable solution for TVRage specials.

What do you think?

fernandog:
@drglove

drglove:
This sounds like it would work, I like avoiding using TVDB. I see that TVDB doesn't care about episode number for specials also http://thetvdb.com/wiki/index.php/Special_Episodes
I don't know the code very well: is it possible to store multiple episodes with the same season+episode number in the database? Is there an easier way to tag an episode as a TVRage special than by assigning S00E00?
I won't have much time over the next week to work on this, but I'll try to put together something in the not too distant future unless someone beats me to it.

fernandog:
@drglove 

"Is there an easier way to tag an episode as a TVRage special than by assigning S00E00?"
- I don't think so. At least KODI/XBMC uses the S00 way:

"Special episodes, if supported by your scraper, should be marked as being season 0."
http://kodi.wiki/view/Naming_video_files/TV_shows#Special_episodes

we must consider compatibility with KODI.

drglove:
@fernandog
I'm referring to the E00 part directly. I understand using S00.

fernandog:
@drglove it's the same. to be compatible with KODI we must use E00

abeloin:
Not only that but sickrage db need a season and episode number

fernandog:
@drglove any news?

drglove:
@fernandog Not yet, I have been quite busy the past two weeks. I will try to get at this by the end of the week.

drglove:
Sorry for the delay, still very busy. Just a not that I have not forgotten about this.

fernandog:
@drglove ok. we are waiting! =)

fernandog:
@drglove still waiting... =)

drglove:
@fernandog Finally got some time away from grad school, we'll see what I can pump out over the next day or two.

drglove:
@fernandog Finally got some time away from grad school, we'll see what I can pump out over the next day or two.

drglove:
@abeloin I'm currently working on the last component you've listed:
"Modify the searching to search for show and title when indexer is tvrage and season 0."
I don't know the codebase very well, but I'm poking around. Is this something that can be done before handing off the search to each provider, or do we need to do this on a per-provider level? As far as I can tell, this will require going to each provider and modifying their search behaviour.

abeloin:
@drglove Sadly, it is per provider

drglove:
I spent some time mucking about with this, but I haven't made any progress with the providers. This looks like a bigger job than anticipated and I won't be around to work on this for the next month at least. If someone else would like to take a crack at this, feel free to have at it.

PR #1253, opened 2015-01-27
Fix for errors on setting episodes to failed

TagForce:
- Added an index to search_queue.segment as it is a list of 1 segment,
instead of just a segment. epObj would therefor become the entire
segment instead of just the epObj beneath it, resulting in object type
list has no attribute 'prettyName' or 'episode' errors.


TagForce:
Fixes https://github.com/SiCKRAGETV/sickrage-issues/issues/499

TagForce:
There's an easier fix. I'll implement that and send a new PR.

PR #1254, opened 2015-01-27
Added Episode check to PP history check

TagForce:
- Fixes issue when Post-Processing a file with a generic filename in a
subfolder resulting in "Already Processed" errors

- Added a Show, Season, Episode parse to the dirname, and if that
results in usable data it adds it to the history check so this is
performed on that specific episode only


TagForce:
https://github.com/SiCKRAGETV/sickrage-issues/issues/491

PR #1259, opened 2015-01-30
Added RARBG provider

CoRpO:
Basic functions working (Rss & search)


fernandog:
Nice!

CoRpO:
Rarbg is "protected" with a cookie, which I hardcoded for now (hasn't changed since october)

Will update later this PR with cookie value detection, and add the ability to set minimum seeders.

Miss an account to provide login possiblity

fernandog:
is this still working?
is it possible to add the login settings instead of the cookie? I assume that you are using your personal cookie. Am I correct?

Also if you are in the mood, can you add https://publichd.to in the future ?
It needs login too.

CoRpO:
@fernandog yes, it's still working. it's no personnal cookie, there is javascript on the pages to set a cookie to "block" bots, and it's value hasn't changed yet and is not IP dependant.

For the login part, I can't test it because I don't have an account there and registrations are closed.

I'll have a look at publichd when I have some time, I do have an account there

fernandog:
@CoRpO ok! so we don't need login/password for this provider. Don't bother add it's not necessary right now.

abeloin:
@CoRpO Please rebase so it can be merged.

CoRpO:
What do you call a rebase ? (I'm new to github)
A commit against the current version ?

fernandog:
Yes. its an alternative

abeloin:
@CoRpO Here is more info on how to rebase a pull request: https://github.com/edx/edx-platform/wiki/How-to-Rebase-a-Pull-Request

CoRpO:
Thank you @abeloin , I'll look into it, did google about that but found nothing that I understood ^^

PR #1266, opened 2015-01-31
Add apparmor profile

d1b:
You may not want to merge this, at least not in its current state, but this pull request contains the [apparmor](https://en.wikipedia.org/wiki/AppArmor) profile that I use with sickrage.


abeloin:
@d1b Can you remove the kickass and upstart commit from this pull request?

PR #1270, opened 2015-01-31
Separate providers for oldpiratebay and thepiratebay

mwarnaar:
As TPB went online again as of today (Jan. 31), I added a separate provider for TPB and oldpiratebay. Currently there is a problem where old torrents seem to have invalid magnet links (the btih part seems to be missing), so it might be the case that this will only work for torrents that were uploaded from Jan. 31 on.


PxlBuzzard:
:+1:

PR #1277, opened 2015-02-02
changed subcategories order

djoole:
so that category "tv series" is used for cache building


fernandog:
@djoole can you explain the subcategories and what are they?

djoole:
433 = Série TV
637 = Animation Série
455 = Animation

These are 3 of several sub categories in T411 search engine, for the category Videos

When I added the provider to SickRage, it was using only 433. Someone changed it to use the 3 subcats,  but it seems the cache building is done only with the first one.

That's why I want to put 433 in first.

PR #1280, opened 2015-02-05
Set the default Pushbullet target to all devices

Arcanemagus:
Sets the default device to "All devices" (string choice may need changing). A user may still select a specific device if desired. Fixes https://github.com/SiCKRAGETV/sickrage-issues/issues/511.


Arcanemagus:
Looks like all pull requests are supposed to go to the develop branch, closing this and opening a new one there.

PR #1281, opened 2015-02-05
Provider hounddawgs: Changes to fit new site layout

SickrageDK:


fernandog:
@SickrageDK 

"NET::ERR_CERT_COMMON_NAME_INVALID" - the server cert is unsecure.

may broke the script. Did you test?

abeloin:
Using https://hounddawgs.org/ would solve the invalid cert

SickrageDK:
The reason for using the IP is because initially the login would not work using the hostname.Maybe because of some redirections. Could try to change it to the hostname and do some testing

SickrageDK:
Also the reason for update is that they put some new tables on the site, so the provider needed update

fernandog:
@SickrageDK any updates? Is it working with hostname or with IP? did you tested? can you send some logs of this PR working?

abeloin:
@SickrageDK Any news?

SickrageDK:
I have updated and done some initial testing. Seems to work with hostname.

PR #1286, opened 2015-02-06
Added option to set ShowUpdate time

TagForce:
- Changed hardcoded value from 3am to SHOWUPDATE_HOUR
- Added field in Config General to set SHOWUPDATE_HOUR
- Default value is still 3 (am), added check if value over 23 or under 0
set to 0 (12am).

This will help people that turn off their rig for the night and, in doing so, miss showupdates in SR causing downloads to not work.


TagForce:
https://github.com/SiCKRAGETV/sickrage-issues/issues/6

PR #1287, opened 2015-02-09
Added detection of fs not implemented link creation

abeloin:
Function acquire will run in infinite loop if the fs doesn't support
link creation(OSError: (38, 'Function not implemented'))
E.g.: Unraid shares


abeloin:
Ref: https://github.com/SiCKRAGETV/sickrage-issues/issues/302

PR #1296, opened 2015-02-11
Added new regex 'itunes'

Deddiekoel:
Almost identical to no_season regex but omitted the "of joiner line" to make sure that iTunes maintained TV Show folders can still be scanned.


abeloin:
Ref: SiCKRAGETV/sickrage-issues#581

PR #1300, opened 2015-02-12
12 feb

bckwltn:


fernandog:
@bckwltn please close this PR. you can't PR to master

fernandog:
@djack42 is menu bar still broken for you?

djack42:
Yes, the bootstrap has been overwritten again a while back. But my tablet should be able to display it entirely, i have 1200px wide and the menu can't display one one line because of the minimum density in chrome for android. Overriding the minimum density could be a better solution than collapsing the menu...
Or replacing some text menu by a picture like configuration

fernandog:
@djack42 how to fix it? I couldn't identify what changed on this commit. There's A LOT of commits

PR #1301, opened 2015-02-12
Fix for issue #425: Replace TVTorrents Provider with Shazbat.tv

CerebralXor:


fernandog:
As discussed here: https://github.com/SiCKRAGETV/sickrage-issues/issues/425

PR #1303, opened 2015-02-12
Fix unknown quality

abeloin:
See SiCKRAGETV/sickrage-issues#521 for more info

### DIFF
--- a/gui/slick/interfaces/default/manage_backlogOverview.tmpl
+++ b/gui/slick/interfaces/default/manage_backlogOverview.tmpl
@@ -46,22 +46,25 @@
 #end if
 #set $totalWanted = 0
 #set $totalQual = 0
+#set $totalSnatched = 0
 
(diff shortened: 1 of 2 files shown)


fernandog:
@abeloin shouldn't we revert the commit to manage_backlogOverview.tmpl ? some users in the forum are question that change. Let's go back to previous version of this file (but keep the change about unknown in tv.py). After stable master we overhall the backlog overview. what do you think?

TagForce:
@fernandog I think they're having different issues with downloads not being completed (they're saying the status never changes from snatched to downloaded).

We can, obviously, revert every change we made to the quality settings for the backlog overview, since we've basically decided that the backlog overview page needs to be rewritten (almost) completely. I'm not sure we have to, though (revert all the changes).

abeloin:
@fernandog I think we should wait until we have more info about the issue before reverting.

PR #1308, opened 2015-02-13
Monkeypatch shutil.copyfile

abeloin:
Replace shutil.copyfile for a faster copy/move. Replace PR https://github.com/SiCKRAGETV/SickRage/pull/1193 and https://github.com/SiCKRAGETV/SickRage/pull/1262.
See https://github.com/SiCKRAGETV/sickrage-issues/issues/545 for more info.

### DIFF
--- a/SickBeard.py
+++ b/SickBeard.py
@@ -26,9 +26,13 @@ codecs.register(lambda name: codecs.lookup('utf-8') if name == 'cp65001' else No
 import time
 import signal
 import sys
-import shutil
 import subprocess
 import traceback
+
+import shutil
+import lib.shutil_custom
+
+shutil.copyfile = lib.shutil_custom.copyfile_custom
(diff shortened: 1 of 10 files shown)


fernandog:
@abeloin this PR will change the way PP copy/moves and maybe check permissions and warn user about they? related to all those issues about PP couldn't find the directory.

abeloin:
Nope, this restore the usage of shutil to transfer file and fix the low speed some user experienced on Mac when move failed and use copy instead

abeloin:
I'm waiting until we push the current changes from develop to master before introducing this patch.

Thraxis:
https://github.com/SiCKRAGETV/sickrage-issues/issues/844

Note my solution I believe the import shutil from the \__init__.py file was removed inadvertently

PR #1311, opened 2015-02-13
Added TNTVillage provider

gborri:
Added tntvillage (http://www.tntvillage.scambioetico.org/index.php) provider, an Italian torrent


abeloin:
@fernandog This shouldn't be merge as it is.

@gborri 
The following should be reviewed:
* The added configuration are specific to the provider but doesn't hide it for other providers(configProviders.js)
* The option category doesn't give any example. In the config it's declared as an integer, so only one category? Also is it necessary to have a custom category, why not search with cats 29, 8, 7 instead?
* The number of "page" shouldn't exist. It should automatically retrieve between 25 to 75 results max.
* sickbeard/providers/generic.py shouldn't be touched.
* Logo must be 16x16

gborri:
Hi Alexandre,

I will fix as per your suggestion . form the category only one category
could be used in URL.

Giovanni
Il 14/feb/2015 00:22 "Alexandre Beloin" <notifications@github.com> ha
scritto:

gborri:
Hello Alexandre,

I was looking to make number o torrent result to only 75, but is not
possible, only 20 per page. to get 80 you need to get 8 pages.
why is not possible to configure the parameter as per your suggestione?

thanks
Giovanni

abeloin:
@gborri Ok then set it to 3 or 4 pages. Don't forget that most of the time, sickrage will search for the specific episode which will never yield that many results. That is why I was saying you don't really need that options or does TNTVillage provides more than 80 versions of the same episode?

gborri:
if you see my code, the limit is set but if you receive less torrent file
you don't ask for more page.
I have some problem to understand the change i have to made on
configProviders.js

Giovanni

abeloin:
Sorry for the confusion, wrong file I meant tmpl fileconfig_providers.tmpl.
Check the provider:
```#if $hasattr($curTorrentProvider, 'page'):```
to something like this
```#if $hasattr($curTorrentProvider, 'page') and ${curNewznabProvider.getID()} == 'TNTVillage':```

gborri:
hello Alexandre,

i have a doubt. But if i don't make the changes you have suggested on other
providers configuration i can't see the parameter unless i declare the
variable in the provider script. Is that True? why I have to make a check
in order to hide it? in the future someone could need it.

thanks
Giovanni

gborri:
hello Alexandre,

another question: you have said that the list of torrent to download is not
more than 75. In TNTvillage i have found that some tv show for some series
have more than 75 entries (in same case 120 in fact there is SD, 720p,
1o80p version and in same case whe have single episode or multiple episode
and also season episode) in this case i will loose some episode. In this
case i will like to have the possibilities to extend the numer of episode
to download. I'm used to set the page parameter but would be useful to have
it configurable in order to manage those case without modifing the code.

Is it possible to mantain this parameters?

thanks
Giovanni

gborri:
hello Alexandre,

I have changed the logo to 16x16
Changed something in tntvillage.py to manage correctly quality recognition.
For category there only the possibility to specify one category or all (category 0).

For generic.py i have no idea why is changed, but i don't know how to fix it in this pull request.
I'll wait for your Answer to my previous question. 

Thanks
Giovanni

fernandog:
@gborri can you send a print of these? I don't think there's that much of episodes. Maybe there's a lot of fakes. Usually there's at most 10 results for every show (a good provider with no fakes)

gborri:
here it is (the only error was that the releases are 102 and not 120, i
thought 120 because i have seen 6 page)
No fake.
it's simple: 22 episode (SD, HD, FULLHD) and in the future there would be a
BD version.

Of course we are speaking about the season search.

[image: Inline image 2]

gborri:
image

fernandog:
searching just for S01 sure you will lots of results
but searching only for S01EXX you will only get a few.

gborri:
I don't know how it works with other provider but with tntvillage you often
find release as for example S02E03-06, if you doesen't search by season you
never find that release and in this case more then often you cannot find
single episode.
That's the point. if you don't mind i think that have configurable number
of page is useful for tntvillage for sure, for other provider i don't know.

abeloin:
In reply to your comment https://github.com/SiCKRAGETV/SickRage/pull/1311#issuecomment-74634929:
If other torrent provider decide to use "select torrent also without italian language" then they will add themselves in the if condition but I doubt there will be many. That particular setting is only good with TNTVillage so I don't see why it should be enable for all the provider by default.

gborri:
@abeloin is this fine for you?

abeloin:
@gborri In reply to your comment https://github.com/SiCKRAGETV/SickRage/pull/1311#issuecomment-74636578:
In this case, I would recommend adding a ```if mode == 'Season':``` to your search code. This way you can search 1 or 2 pages when in episode mode and, let say, 10 pages when you are in Season mode.

The end result is to make it your provider the most user friendly as possible.

abeloin:
@gborri Can you rename the variable subtitle to subtitle_italian? As the setting is Italian specific.

gborri:
@abeloin is not needed to have different approch for season and for episode.
if there is only 2 pages of result the plugin stop to retrieve data on second page.

for renaming the parameter is not a problem.

abeloin:
@gborri Then set it to 10(or the number you judge appropriate) pages which is 250 results and remove the setting from the configuration page. 

The goal here is to automate the number of page pulled, not to have the user specify the number of page it want retrieved and then complain he couldn't find a release because he set the page to 1.

abeloin:
@gborri Looking at the website, I think there's a way not to specify the categories:
*Search with cat 0.
* In the parsing, look for icon at the beginning.
* The category is included in the icon name, extract the value.
* Compare with self.category_dict

This way you can search for all the categories: 29, 8, 7, 14 if you have anime and tv series in sickrage instead of having to switch manually each time.

What do you think?

gborri:
if you see on category_dict there is also the option 'ALL' : 0.
I prefer to have the possibility to select only one category because the
number of release will grow and most of the time i use only catogory TV
Series. if you don't mind i'll prefer to have the cat parameters.

for page paraeter, ok. i will remove it.

gborri:
@abeloin i'm trying to migrate subtitle parameter into subtitle_italian parameter but i don't know the structure of the providers dict.
i have tried this but i think that's not correct.
could you help me please?

in config.py:

    def _migrate_v7(self):
        """
        Reads in the old naming settings from your config and generates a new config template from them.
        """
        # get the old settings from the file and store them in the new variable names
        sickbeard.providers.TNTVILLAGE.SUBTITLE_ITALIAN = check_setting_str(self.config_obj, 'TNTVILLAGE', 'tntvillage_subtitle', '')

abeloin:
@gborri Why do you want to migrate it?
As far as I know the plugin don't exist in master nor develop so you can change the all the references in the files modified.

gborri:
Because there is some people that are using my repo and i would like to not
changing their config due to migration.

Giovanni
Il 18/feb/2015 00:38 "Alexandre Beloin" <notifications@github.com> ha
scritto:

abeloin:
@gborri Ok I see your point, mmm I'm not sure that will work as I don't think you can access the provider like that.

Alternative: is it possible to change the "select torrent also without italian language" to "select torrent with Italian subtitle" this way the subtitle tag can be reused for other providers with specific language.

gborri:
@abeloin got your suggestion on subtitle.
for page: If i will set 10 pages for search, it will set also for daily search and in this way i will always get 10 pages.... that's huge.

gborri:
@abeloin sorry i got a solution.

in doSearch

        if mode = 'RSS':
            self.page = 2

gborri:
@abeloin last question: subtitle should be hide for provider other then tntvillage or not?

abeloin:
@gborri Yes it should be hidden for any other provider.

gborri:
ok. it is.
i have merged with las commit on develop.
if you don't have any other comment, could you please merge it?

thanks
Giovanni

On Wed, Feb 18, 2015 at 12:28 PM, Alexandre Beloin <notifications@github.com

abeloin:
@gborri Merged

gborri:
thanks.

Giovanni

On Thu, Feb 19, 2015 at 12:29 AM, Alexandre Beloin <notifications@github.com

PR #1322, opened 2015-02-16
 Fix subtitles display issues.

abeloin:
When multiSubtitles is not activated:
-Download history show flag of the last 2 chars before ext.
-Subtitle lang was always und.
-Keep lang value between refresh


abeloin:
Ref: https://github.com/SiCKRAGETV/sickrage-issues/issues/88

PR #1325, opened 2015-02-16
Updated network logos.

neoatomic:
Updated network logos.


Thraxis:
I am a bit curious about the TMF logo removal.  yes the channel may have shut down in 2011, but people could still add shows that had broadcast from that channel prior to 2011 to their sickrage.

neoatomic:
I understand your point.
However i'm familiar with this channel, otherwise i would not have risked to removed it.
TMF was a Dutch music channel that did only broadcast music clips and not series like (for example) MTV does.

PR #1326, opened 2015-02-16
Improved logging to detect CloudFlare blocking

flightlevel:
Improved logging to detect when CloudFlare is blocking via javascript challenge.
Relates to issue
https://github.com/SiCKRAGETV/sickrage-issues/issues/504

### DIFF
--- a/sickbeard/providers/freshontv.py
+++ b/sickbeard/providers/freshontv.py
@@ -111,6 +111,10 @@ class FreshOnTVProvider(generic.TorrentProvider):
             if re.search('Username does not exist in the userbase or the account is not confirmed yet.', response.text):
                 logger.log(u'Invalid username or password for ' + self.name + ' Check your settings', logger.ERROR)
                 return False
+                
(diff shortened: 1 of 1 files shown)


fernandog:
@flightlevel what do you think about do IF username - ELSE IF ddos  else message error that couldn't login. To catch a 503 error and everything else.

flightlevel:
@fernandog 
Don't really like the except pass line 128, maybe we could catch that error and log
or we could change the logic a bit
From the response check if we actually logged in
`If logged in`
`        grab cookie with try/except`
`else`
`       log if username/pass is wrong`
`else`
`       log if cloudflare`
`else`
 `      log unknown response`

?????

abeloin:
Why not let requests.session handle the cookies? Move L103 out of _doLogin to the __init__ then simply check(L91) if self.session.cookies contain data instead of checking the cookiesjar. This way you remove the try/pass.

Instead of checking for error message first why not look for something specific that exist only when logged. Then if it fail, check if the message is user/pass related. Check ddos cloudfare protection. Else unknown response.

PR #1344, opened 2015-02-18
Download single file from multiepisode torrent

gborri:
is only available on transmission. I have no knowledge on other clients.

### DIFF
--- a/sickbeard/providers/tntvillage.py
+++ b/sickbeard/providers/tntvillage.py
@@ -19,10 +19,9 @@
 import re
 import traceback
 import datetime
-import urlparse
 import sickbeard
 import generic
-from sickbeard.common import Quality, cpu_presets
+from sickbeard.common import Quality
 from sickbeard import logger
 from sickbeard import tvcache
 from sickbeard import db
@@ -32,7 +31,7 @@ from sickbeard import show_name_helpers
 from sickbeard.exceptions import ex, AuthException
 from sickbeard import clients
 from lib import requests
-from lib.requests import exceptions
+from lib.requests.exceptions import RequestException
 from sickbeard.bs4_parser import BS4Parser
 from lib.unidecode import unidecode
 from sickbeard.helpers import sanitizeSceneName
@@ -152,7 +151,7 @@ class TNTVillageProvider(generic.TorrentProvider):
 
         try:
             response = self.session.post(self.urls['login'], data=login_params, timeout=30, verify=False)
-        except (requests.exceptions.ConnectionError, requests.exceptions.HTTPError), e:
+        except RequestException as e:
             logger.log(u'Unable to connect to ' + self.name + ' provider: ' + ex(e), logger.ERROR)
             return False
 
@@ -174,10 +173,7 @@ class TNTVillageProvider(generic.TorrentProvider):
                 ep_string = show_name + ' ' + "%d" % ep_obj.scene_absolute_number
             else:
                 ep_string = show_name + ' S%02d' % int(ep_obj.scene_season)  #1) showName SXX
-
             search_string['Season'].append(ep_string)
-
-
         return [search_string]
 
     def _get_episode_search_strings(self, ep_obj, add_string=''):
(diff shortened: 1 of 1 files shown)


dmarcelino:
This is a great feature! Synology's Download Station should also support this as it's possible to do it from the web ui (and, I believe, Download Station is based on transmission).

gborri:
@abeloin @fernandog wait a moment to merge it. 
@dmarcelino for now it's working only on first episode of seasonpack or multiepisode pack.
need to make sickrage get episode list fron season string SiCKRAGETV/sickrage-issues#588

sickbeard/clients/generic.py
```diff
@@ -1,241 +1,275 @@
...
+    def _get_torrent_hash(self, result):
+
+        if result.url.startswith('magnet'):
+            result.hash = re.findall('urn:btih:([\w]{32,40})', result.url)[0]
+            if len(result.hash) == 32:
+                result.hash = b16encode(b32decode(result.hash)).lower()
+        else:
+            if not result.content:
+                logger.log('Torrent without content, try to geti it', logger.ERROR)
+                result.content = result.provider.getURL(result.url)
```
gborri: @abeloin i've added this peace of code because it happens that result.content is empty not because the torrent is corrupt. let me know if it's fine

gborri:
will fix SiCKRAGETV/SickRage#588

sickbeard/providers/generic.py
```diff
@@ -274,13 +274,13 @@ def findSearchResults(self, show, episodes, search_mode, manualSearch=False):
                 continue
 
             # skip if season already searched
-            if len(episodes) > 1 and searched_scene_season == epObj.scene_season:
+            if len(episodes) > 1 and searched_scene_season == epObj.scene_season and not(episodes and sickbeard.TORRENT_METHOD == 'transmission'):
```
gborri: @abeloin i've used this method (sickbeard.TORRENT_METHOD == 'transmission') to know which client is used that is hard coded. There is other solution through sinlge_episode_enable procedure in clients.
the second way i think is better in order to have other clients to support it but we have to instantiate the client in orther to have this information. The first solution probably is better for performance. let me know.

fernandog:
@gborri with this, you can get a torrent that have multiple files and select just one? Is that it ?

if only Transmission support it, let's use it only with transmission.

gborri:
@fernandog yes that'ss the case. It's only supported transmission because I have started to have knowledge on it. I Think that all clients can support for it but we need people that have know how on other clients to get it work.

gborri:
@fernandog you don't have to chose which file to download from torrent. When you select (manually or via backlog) an episode and on torrent is only available a multiepisode, the client select for download only the episode that you have chosen. I'ave added also the fallback manage also on single episode search. I forgot to change the description on provider, i will do it.

gborri:
@fernandog need to be tested with other provider because I have tested it only with TNTVillage. But I think that this change is not provider dependent.

fernandog:
@gborri Search mode should be sponly to work? but if sponly is selected user usually wants full season. How to handle that?

gborri:
@fernandog i think that the point is: if you want full season you have to change to wanted all episode of that season.

gborri:
@fernandog the onnly way for which you will find torrent like S03-06 and so on is with sponly.
You can use as default or you can try with ep only and have fallback to sponly.

fernandog:
Ok. Imagine I only want one episode and this single episode is only available in  season pack. With your PR I will be able to only download that ep from the season pack isnt? But ep search dont get the season pack. If searchs SxxExx. So this wont work.

gborri:
Will work. Try it. The point is that if you fall back to season search it
will find it or you can define sponly  also for episode search.

Giovanni
Il 18/mar/2015 16:27 "Fernando" <notifications@github.com> ha scritto:

abeloin:
@gborri All changes in sickbeard/providers/generic.py
 need to be backported into any provider overriding a function.

gborri:
@abeloin i've make some mistake and needed to reset hard back to previous commit.
This probably erased the link with this PR. I have to create a new PR?

gborri:
@abeloin never mind. i have lost all stuff and i need to restart.

PR #1349, opened 2015-02-19
Fix self.hdtext declared as list instead of dict

abeloin:
Python 2.6 report syntax error


abeloin:
@gborri FYI

abeloin:
Ref PR https://github.com/SiCKRAGETV/SickRage/pull/1311
Ref issue https://github.com/SiCKRAGETV/sickrage-issues/issues/664

PR #1350, opened 2015-02-19
Fix color in displayShow when manually searching.

abeloin:
Changed function getManualSearchStatus to include overview.
Changed all attr to prop

### DIFF
--- a/sickbeard/webserve.py
+++ b/sickbeard/webserve.py
@@ -1809,6 +1809,7 @@ class Home(WebRoot):
     def getManualSearchStatus(self, show=None, season=None):
         def getEpisodes(searchThread, searchstatus):
             results = []
+            showObj = sickbeard.helpers.findCertainShow(sickbeard.showList, int(show))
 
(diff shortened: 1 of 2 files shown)


sickbeard/webserve.py
```diff
@@ -1809,6 +1809,7 @@ def searchEpisode(self, show=None, season=None, episode=None):
     def getManualSearchStatus(self, show=None, season=None):
         def getEpisodes(searchThread, searchstatus):
             results = []
+            showObj = sickbeard.helpers.findCertainShow(sickbeard.showList, int(show))
```
fernandog: is this related? 
https://github.com/SiCKRAGETV/sickrage-issues/issues/684

fernandog:
@abeloin sorry reverting this. Couldnt fix it.

abeloin:
@fernandog I'll redo the patch. It's mainly broken because the epAjaxSearch.js sends the word undefined instead of show integer when the values doesn't exists.

PR #1351, opened 2015-02-19
Fix navbar not collapsing properly

abeloin:
Re-applied commit f2bcc72 because commit 0da960f erased it

```
From f2bcc7217c391ea823f0e0ed99ababaef4b3622d Thu, 19 Feb 2015 01:49:30-0500
From: markheloking <markheloking@live.nl>
Date: Mon, 8 Dec 2014 15:14:52 +0100
Subject: [PATCH] Fixed bootstrap @grid-float-breakpoint


http://getbootstrap.com/components/
Look in that page for: "Changing the collapsed mobile navbar breakpoint"

Fixed collapse fixed width from 768px to the width of our current
navbar: 1010px
```

### DIFF
--- a/gui/slick/css/lib/bootstrap.css
+++ b/gui/slick/css/lib/bootstrap.css
@@ -1111,7 +1111,7 @@ p {
   font-weight: 300;
   line-height: 1.4;
 }
-@media (min-width: 768px) {
+@media (min-width: 1010px) {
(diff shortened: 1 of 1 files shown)


abeloin:
Ref https://github.com/SiCKRAGETV/sickrage-issues/issues/643

PR #1353, opened 2015-02-19
Check if restore directory is empty to not restore

fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/665

### DIFF
--- a/SickBeard.py
+++ b/SickBeard.py
@@ -263,7 +263,7 @@ class SickRage(object):
 
         # Check if we need to perform a restore first
         restoreDir = os.path.join(sickbeard.DATA_DIR, 'restore')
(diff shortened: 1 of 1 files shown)


abeloin:
@fernandog This commit makes no sense you just pull the content of that folder and do nothing with it

fernandog:
@abeloin what if user has an empty "restore" folder?  Restore process will fail

abeloin:
Better use try...except instead.

fernandog:
will you add this to your commit? I'm closing mines.

abeloin:
It's already in https://github.com/SiCKRAGETV/SickRage/pull/1358 . Each of the 2 steps are now enclosed with a try...except.

PR #1355, opened 2015-02-19
Check if restore has cache folder (new backup/restore version)

fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/665#issuecomment-75041585

### DIFF
--- a/sickbeard/__init__.py
+++ b/sickbeard/__init__.py
@@ -634,7 +634,7 @@ def initialize(consoleLogging=True):
 
         # Check if we need to perform a restore of the cache folder
         restoreDir = os.path.join(DATA_DIR, 'restore')
-        if os.path.exists(restoreDir):
+        if os.path.exists(restoreDir) and os.path.exists(srcDir):
             def restoreCache(srcDir, dstDir):
                 import ntpath
                 import shutil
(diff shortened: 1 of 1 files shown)


fernandog:
@abeloin is this ok? the srcDir is already defined there?

sickbeard/__init__.py
```diff
@@ -634,7 +634,7 @@ def initialize(consoleLogging=True):
 
         # Check if we need to perform a restore of the cache folder
         restoreDir = os.path.join(DATA_DIR, 'restore')
-        if os.path.exists(restoreDir):
+        if os.path.exists(restoreDir) and os.path.exists(srcDir):
```
abeloin: This wont work as you intend to as the srcDir is define in the function and not accesible from outside.

It should be like this:
```
if os.path.exists(restoreDir) and os.path.exists(os.path.join(restoreDir, 'cache')):
```

fernandog:
is it ok now or should I do just on commit?

abeloin:
That should be fine

fernandog:
keep this close?

abeloin:
Yes it's included in #1358

PR #1357, opened 2015-02-19
Check if there's already a "restore" folder before make a new restore

fernandog:


sickbeard/helpers.py
```diff
@@ -1081,7 +1081,9 @@ def backupConfigZip(fileList, archive, arcname = None):
 
 def restoreConfigZip(archive, targetDir):
     try:
-        if not os.path.exists(targetDir):
+        if os.path.exists(targetDir):
+            shutil.move(targetDir,os.path.join(targetDir, '{0}.bak-{1}'))
```
abeloin: This wont work as you intended. You need to extract the leaf from the targetdir, then append the string you want(bak-date) then append it to the base of targetDir.

sickbeard/helpers.py
```diff
@@ -1081,7 +1081,9 @@ def backupConfigZip(fileList, archive, arcname = None):
 
 def restoreConfigZip(archive, targetDir):
     try:
-        if not os.path.exists(targetDir):
+        if os.path.exists(targetDir):
+            shutil.move(targetDir,os.path.join(targetDir, '{0}.bak-{1}'))        
+        elif not os.path.exists(targetDir):
```
abeloin: There's no need to check the value again, a simple else is enough.

fernandog:
how to edit the code without creating a new branch and PR again?

abeloin:
@fernandog Checkout the branch fernandog:override_restore, make the change then commit. After push the new commit to git hub.

PR #1359, opened 2015-02-19
TPB: Suppressed torrent list not found error message

mwarnaar:
TPB's ajax_details_filelist.php call seems to only return 
> File list not available. 

so season search mode will not work using TPB at this moment.
I suppressed the error message as it's annoying (see SiCKRAGETV/sickrage-issues#631 and SiCKRAGETV/sickrage-issues#641) and also added a missing `return None` when the file list cannot be retrieved (so always atm.)

### DIFF
--- a/sickbeard/providers/thepiratebay.py
+++ b/sickbeard/providers/thepiratebay.py
@@ -122,7 +122,9 @@ class ThePirateBayProvider(generic.TorrentProvider):
         filesList = re.findall('<td.+>(.*?)</td>', data)
 
         if not filesList:
-            logger.log(u"Unable to get the torrent file list for " + title, logger.ERROR)
+            # disabled errormsg for now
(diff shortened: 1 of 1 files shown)


fernandog:
@mwarnaar shouldn't we add a warning or try to disable season mode for TPB using javascript?

mwarnaar:
Yeah that's also an option, but it should work once TPB fixes it on their side.

yaymuffins:
I'd like to add something I didn't know before.

Apparently it only affect torrents before the takedown, look:

https://thepiratebay.se/torrent/11711573/Alexander_and_the_Terrible_%282014%291080P_DTS-AC3_5.1_NedSubs_TBS > Uploaded after the takedown > file list is working

https://thepiratebay.se/torrent/9758468/The_Simpsons_Complete_Season_6_DVD_RIP_(1994_ndash_95) > Uploaded before the take down > "File list not available."

PR #1361, opened 2015-02-19
Update trakt error checking

fernandog:
The goal is not to show any tracebacks

### DIFF
--- a/lib/trakt/exceptions.py
+++ b/lib/trakt/exceptions.py
@@ -1,8 +1,0 @@
-class traktException(Exception):
-    pass
-
-class traktAuthException(traktException):
-    pass
-
-class traktServerBusy(traktException):
-    pass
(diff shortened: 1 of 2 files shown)


abeloin:
You shouldn't do that this way. Catch the raise error instead.

fernandog:
How?

abeloin:
From were the trakt function is call in the code

abeloin:
Example:
try:
    trakt = TraktAPI()
except traktException as e:
    logger.log('I've catched an error and won't show a traceback', log.INFO)

abeloin:
If it's related to https://github.com/SiCKRAGETV/sickrage-issues/issues/669 then change the exception in https://github.com/SiCKRAGETV/SickRage/blob/develop/sickbeard/traktChecker.py#L56 to:
```
except Exception as e:
    logger.log('Trakt: Error Syncing library: {0}'.format(str(e)), logger.DEBUG)
```

PR #1369, opened 2015-02-20
Try fix special char in filename

fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/122


fernandog:
@abeloin will it work? is everything ok?

PR #1370, opened 2015-02-20
Remove trademark from filename

fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/122


fernandog:
@abeloin thanks!

PR #1371, opened 2015-02-20
Possible fix for kodi notify

fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/658


fernandog:
@abeloin are you sure this is the problem? seems so. was that easy? =)

abeloin:
@fernandog My bad, change str(response) to str(response.read()) because we want the content not the object. Sorry

fernandog:
@abeloin done

PR #1372, opened 2015-02-20
Added Default Info Language in general settings

gborri:
give the ability to choose default language in ortder to use it when adding show from api, trending show, recommended show etc.


abeloin:
@gborri Thank You for this PR. Here some points while looking at your code:

You shouldn't change function default language to None. E.g.:
from
```
def addRecommendedShow(self, whichSeries=None, indexerLang="en", rootDir=None, defaultStatus=None,
```
to
```
def addRecommendedShow(self, whichSeries=None, indexerLang=None, rootDir=None, defaultStatus=None,
```
It will break as the function was intended to have a default value for indexerLang, you should replace with the indexer variable:
```
def addRecommendedShow(self, whichSeries=None, indexerLang=sickbeard.DEFAULT_LANGUAGE, rootDir=None, defaultStatus=None,
```
Also the variable shouldn't be called DEFAULT_LANGUAGE but rather INDEXER_LANGUAGE. This ways it leave DEFAULT_LANGUAGE free if Sickrage is one day multi-languages.

abeloin:
I forgot to mention there's no need to make an external call to get the language, use this in the template:

```
$sickbeard.indexerApi().config['valid_languages']
```

gborri:
@abeloin for the call to get the language i've copied from editShow.tmpl.
probabily need to chenage also that.

gborri:
@abeloin for the default value in call procedure what you have suggested was my first try but it gave me an error: Traceback (most recent call last):
  File "SickBeard.py", line 58, in <module>
    import sickbeard
  File "/home/mediacenter/src/sickbeard/sickrage/sickrage_choose_default_language/sickbeard/__init__.py", line 42, in <module>
    from sickbeard import helpers, db, exceptions, show_queue, search_queue, scheduler, show_name_helpers
  File "/home/mediacenter/src/sickbeard/sickrage/sickrage_choose_default_language/sickbeard/show_queue.py", line 34, in <module>
    class ShowQueue(generic_queue.GenericQueue):
  File "/home/mediacenter/src/sickbeard/sickrage/sickrage_choose_default_language/sickbeard/show_queue.py", line 135, in ShowQueue
    lang=sickbeard.INDEXER_LANGUAGE, subtitles=None, anime=None, scene=None, paused=None):
AttributeError: 'module' object has no attribute 'INDEXER_LANGUAGE'

next i've set to None and managed the default value with this peace of code in the procedure:

        if not indexerLang or indexerLang == 'null':
            indexerLang = sickbeard.INDEXER_LANGUAGE

gborri:
@abeloin i got the way to remove external call to get list of languages.

gborri:
@abeloin could you merge it with your pull request #1393?

fernandog:
@gborri what happens if there's no enought information on that language when adding show? will it fallback to ENGLISH? otherwise we will have lot's of issues about not adding the show. That's why the default is english.

abeloin:
@gborri Forget what I said about declaring in the function, it wont w orkbecause python will substitute the variable with the value at compile time.

Question: Why do you check for the word null in indexerLang? Is there something that send the word null to these function? If so we should find them and correct them. Or did you meant to if it's Null? If so you can replace the whole if by ```if indexerLang is None:```

abeloin:
@fernandog The proposed changes here don't touch that at all it will stay the same.

abeloin:
@gborri Will try to merge it.

gborri:
@abeloin i will modify the check as you suggested: if indexerLang is None:

gborri:
@abeloin do you think is possible to merge also this PR for the next release?

abeloin:
@gborri Yes, it will be in 4.0.14

abeloin:
@gborri Squashed your PR into one commit: https://github.com/SiCKRAGETV/SickRage/pull/1633

PR #1381, opened 2015-02-21
Updated network logos.

neoatomic:
Added missing History Channel 2 logo.
Replaced the last white logos with colored logos.
Redone the Comedy Central logos.
Resized tokyotoshokn.png provider logo as is was to large.


abeloin:
@neoatomic A small question, why replace the new comedy central logo for the old one?

Ref: http://en.wikipedia.org/wiki/Comedy_Central

neoatomic:
The old logo looks better in the webui in my opinion.
But if you prefer the "new" Comedy central logo i will restore it. 
Let me know want you think.

neoatomic:
@abeloin I restored the old comedy central logos. I just saw that they where 65x32 instead of 64x32. 

@dsoms I cant make the "3" in black as it will not show properly on the dark Sickrage theme. (transparant background) Thats why i chose to make it red.

dsoms:
@neoatomic totally agree, thanks ;-)

PR #1384, opened 2015-02-21
Subtitles 3/3: Suppressing subliminal logs on windows

abeloin:
Due to issue converting some character to Windows cp1252 failing, suppress log line from being created on Windows.


abeloin:
Do not merge

PR #1386, opened 2015-02-22
Added a specific regex for horriblesubs

abeloin:
See https://github.com/SiCKRAGETV/sickrage-issues/issues/482 for more info

### DIFF
--- a/sickbeard/name_parser/regexes.py
+++ b/sickbeard/name_parser/regexes.py
@@ -198,6 +198,18 @@ normal_regexes = [
 ]
 
 anime_regexes = [
+    ('anime_horriblesubs',
+     # [HorribleSubs] Maria the Virgin Witch - 01 [720p].mkv
+     '''
+     ^(?:\[(?P<release_group>HorribleSubs)\][\s\.])
+     (?:(?P<series_name>.+?)[\s\.]-[\s\.])
+     (?P<ep_ab_num>((?!(1080|720|480)[pi]))\d{1,3})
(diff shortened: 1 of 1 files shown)


dmarcelino:
Given I'm a big consumer of horriblesubs translated anime, thanks! :)

djack42:
Hi, this regex is not 100% ok
It should support the following format with season :
[HorribleSubs] Kindaichi Case Files R S2 - 26 [1080p].mkv
Where S2 is the season and 26 is the episode number (not absolute in this case)

Here is some others :
[HorribleSubs] Fairy Tail S2 - 78 [1080p].mkv
[HorribleSubs] Haikyuu!! S2 - 01 [720p].mkv
[HorribleSubs] Soukyuu no Fafner Dead Aggressor - Exodus S2 - 14 [1080p].mkv

Also, maybe a good feature would be scene exception by season, cause horriblesubs often use a different title for a season without adding the S2 :
[HorribleSubs] K - Return of Kings - 01 [1080p].mkv
For the show K, where Return of Kings is for the second season.
This case is not handled at all by sickrage, best case, the episode is identified in the first season.

@miigotu seems to be really active lately, can you do something about it ? That would really help.

miigotu:
@djack42 Your request would require a second regex, and honestly anime is like the lowest priority on my list atm. However, you could put in FR on [Feathub](http://feathub.com/SiCKRAGETV/SickRage)

fernandog:
This is part of guessit FR isnt?

miigotu:
Actually I stand corrected. This should be part of the guessit FR, as all of those except the last one without season works with guessit.

PR #1390, opened 2015-02-22
Don't remove logs folder when git reset is enabled in UI

fernandog:
git reset should only reset git files not user files
@abeloin what do you think about set "get reset" by default in config.ini?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ cache/
 cache.db*
 config.ini*
 Logs/*
+Logs/
 sickbeard.db*
 failed.db*
 autoProcessTV/autoProcessTV.cfg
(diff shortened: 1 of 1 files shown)


abeloin:
@fernandog Small question: why do we need this addition? Line 7 "Logs/*" already prevent the Logs folder from being deleted.

As for the git reset, I think we should further discuss it implication.

fernandog:
/Logs keep only the folder
Logs/* contents of the folder

You can see there is this for the others

abeloin:
They mean exactly the same thing(not exactly in technical term but on the practical they are):
Logs/ : Ignore the folder and its content
Logs/* : Ignore the folder content (Imply keep the folder)

See http://git-scm.com/docs/gitignore for more information.

fernandog:
@abeloin ok. will change this commit and fix the cache/ too

fernandog:
@abeloin done

PR #1391, opened 2015-02-22
Fix submit error condition

abeloin:
* Fix submit error condition
* Disable auto-submit option in config_general.tmpl as it is disabled in
the code.

See comment https://github.com/abeloin/SickRage/commit/66a2e5c46892fb74b643199d270c549d46813a79#commitcomment-9858577 for more info

### DIFF
--- a/gui/slick/interfaces/default/config_general.tmpl
+++ b/gui/slick/interfaces/default/config_general.tmpl
@@ -634,8 +634,8 @@
 							<label for="git_autoissues">
 								<span class="component-title">Git auto-issues submit</span>
 								<span class="component-desc">
-									<input type="checkbox" name="git_autoissues" id="git_autoissues" #if True == $sickbeard.GIT_AUTOISSUES then 'checked="checked"' else ''#/>
-									<p>automatically submit bug/issue reports to our issue tracker when errors are logged</p>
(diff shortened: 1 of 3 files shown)


fernandog:
I didnt get it.
the auto submit was never intended to submit without user intervation. User must click submit. If user dont set git user/pass and didnt enable the auto submit the button wont appear
its working as is today. We should not change this. i said earlier to just rename the feature

fernandog:
As is working we have a lot of submited issues be cause users click "submit"

abeloin:
@fernandog Honestly it is not good way of working than to hijack a property that was meant for something else and re-purpose it to something else keeping the same name. 99% of all English speaking people will read the text and assume it will send the error report without user intervention.

fernandog:
So we rename it to "Enable Submit issues to GITHUB"
The auto is to auto create the issue. Not automate the process. I undestand is confusing. Lets change the setting name then

abeloin:
He was see line: https://github.com/SiCKRAGETV/SickRage/blob/master/sickbeard/logger.py#L131 . 

This is why @Thraxis made is comment about the auto-submit being disable in logger and the fact it made no sense that the auto-submit needed to be enabled to manually submit an issue via Sickrage.

fernandog:
I guess its remanescent code because auto submit without user intervation will create thousands of errosr (also repeated) . Would be insane to manage it

PR #1393, opened 2015-02-23
Replace language selection in add show.

abeloin:
It now display each language by their name.
Use bootstrap formhelpers 2.3.0 under Apache License v2.0
http://plugins.jquery.com/bfh-language/


abeloin:
Imgur

fernandog:
Awesome!

gborri:
@abeloin could be useful to modify also editShow.tmpl.
i have seen that also in gui//slick/js/newShow.js there is a call to external to retrieve Language, but i'm not sure if it is used in the code.

abeloin:
Ok I will check

abeloin:
@gborri https://github.com/SiCKRAGETV/SickRage/pull/1416

PR #1395, opened 2015-02-23
Handles multi-page results and improved login

flightlevel:
If search results are more than 1 page, it now iterates through each page. Also improved login.
Relates to issue https://github.com/SiCKRAGETV/sickrage-issues/issues/632

### DIFF
--- a/sickbeard/providers/freshontv.py
+++ b/sickbeard/providers/freshontv.py
@@ -19,6 +19,7 @@
 import re
 import traceback
 import datetime
+import time
 import urlparse
 import sickbeard
 import generic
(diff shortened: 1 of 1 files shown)


fernandog:
@flightlevel can you add multi-page to rarbg??

flightlevel:
@fernandog I'll try if I get some free time on the weekend

@abeloin @fernandog Any tips for debugging? For this commit I used PyCharm to write the code, but then to debug I restarted SickRage, made an episode wanted to see what happened, which was bit slow, any faster debugging tips?

abeloin:
There's not much faster than that for live debugging(Via the IDE or import pdb).

The fastest way to check and debug your code would involve creating test cases which you would submit to the modules and compare the results with the predetermined results. But this take time to setup up, have a look in the tests folder.

PR #1397, opened 2015-02-23
Remove torrent file after download

gborri:


fernandog:
Does it need to change DB version? We cant commit this now.

fernandog:
Also can you check that tranmission error I send you?

gborri:
@fernandog sent where? i can't find it

fernandog:
@gborri 
https:// github .com /SiCKRAGETV/sickrage-issues/issues/679

gborri:
@abeloin do you think is it possible to test and merge also this PR before merge to master? It has a db change.

fernandog:
first develop>master
merge new features to develop
create experimental branch to test db changes.

abeloin:
@gborri I won't be possible as we are now in release mode(ie: accepting bugfix and small feature only) to stabilize before releasing to master.

fernandog:
@abeloin should we create an experimental branch? The problem is that users can switch to it. The db check must be in master before create the experimental branch

abeloin:
@fernandog See your PM

fernandog:
@gborri can you fully explain how this works?  we are discussing about it here: 
https://github.com/SiCKRAGETV/sickrage-issues/issues/911

gborri:
@fernandog @abeloin dou you think ti PR could be merged?

fernandog:
Not now. needs DB upgrade

gborri:
@miigotu I don't want to change the InitialSchema and db versione. The philosophy (as i have understood) is to start from the actual InitialSchema and add following change by migration.
For your suggestion on code note: is not needed to check if is trasmission because if not the torrent_hash field will be empty.

miigotu:
@gborri It is needed to update the InitialSchema when you change the database. The other check you already added is only supposed to be used when upgrading an existing database. Initial schema is on a fresh install, and should need no upgrades, so you have to modify the initial also.

miigotu:
@gborri Also, you're right about not needing to check if its transmission, as you have a default 
def remove_torrent_downloaded(self,hash) in generic.py which just returns true, and that is properly done.

gborri:
@miigotu ok, no problem for initial schema change. what i've said before is due to the fact that for a long period the initial schema was not changed. Now i've cheched and see that have update. 
for versione i can set 43 but when merged you need to check the correct one to set.

gborri:
@miigotu Done.

miigotu:
Ok, when I get around to making it possible to restore db to a lower version  backup when downgrading SR we will look into merging this.

miigotu:
@gborri I like this idea, but I see two problems with it. It only handles transmission, and it should only try to remove from the client if the process method is 'move'. I would like to have this ready and on the list for the next db change.

gborri:
@miigotu i handle only transmission because is the only client that I know. We can start from it and if there is someone with more knowledge about other client we can integrate it. for the "move" part i have not understood. the remove is in the postprocess procedure and i haven't seen anything related to the operation. If you give me any highlights i will check.

miigotu:
If someone has the pp method set to copy or hard link or w/e they dont want it removed from their torrent client (they are seeding it)

If it is set to move, they cannot seed it regardless, so it should be removed without question, as the torrent client just gives an error about missing data.

This does not need to be configurable if made this way, we already know if a user wants to keep seeding based on the post processing procedure.

gborri:
ok i git it. the point is the function of removing torrent need to be available only if process metod is set to move. In other cases the user have to remove it manually, correct?

miigotu:
And it doesn't need to have any user configuration or options, just do it automatically. There shouldnt even be a gui checkbox.

sickbeard/postProcessor.py
```diff
@@ -957,6 +958,15 @@ def process(self):
                 else:
                     cur_ep.release_name = ""
 
+                if cur_ep.torrent_hash != '':
+                    client = clients.getClientIstance(sickbeard.TORRENT_METHOD)()
```
miigotu: why trailing () ?

sickbeard/webserve.py
```diff
@@ -1688,6 +1688,10 @@ def setStatus(self, show=None, eps=None, status=None, direct=False):
         if eps is not None:
 
             sql_l = []
+
+            if int(status) == SKIPPED:
+                client = clients.getClientIstance(sickbeard.TORRENT_METHOD)()
```
miigotu: trailing () ?

sickbeard/webserve.py
```diff
@@ -1738,6 +1742,10 @@ def setStatus(self, show=None, eps=None, status=None, direct=False):
 
                     trakt_data.append((epObj.season, epObj.episode))
 
+                    if int(status) == SKIPPED and epObj.torrent_hash != '':
```
miigotu: something is not right here.

sickbeard/webserve.py
```diff
@@ -1688,6 +1688,10 @@ def setStatus(self, show=None, eps=None, status=None, direct=False):
         if eps is not None:
 
             sql_l = []
+
+            if int(status) == SKIPPED:
```
miigotu: and here, something is not right.

sickbeard/postProcessor.py
gborri: because if for any reason the value cur_ep.torrent_hash i don't want to call transmission

gborri: i don't remember. probably i have get this peace of code from other part of sickrage....

sickbeard/webserve.py
gborri: same answear

sickbeard/webserve.py
gborri: the point here is: if you change status of episode to default, you have to remove torrent from transmission and clean torrent_hash

sickbeard/webserve.py
gborri: Why do you think is not right?

miigotu:
Default is not always skipped.

It should only be removed on successful PP or marked as failed.

matigonkas:
Just leaving my opinion ,i would prefer if it where configurable

gborri:
@miigotu the line you have commented is on webserver.py. Here i will check if you want to change status to SKIPPED and in this case i will remove torrent from client and clean torrent hash.

gborri:
@matigonkas ok

fernandog:
I remove mine after download finished. Using script in transmission

miigotu:
@matigonkas If your PP method is "Move", and the file is moved, why leave it in the client? You just get errors in your torrent client.

matigonkas:
@miigotu its to keep track of what specific torrents i have downloaded,and the total nr

gborri:
@matigonkas this information is on SickRage..... or if you have trakt enable also on it.

gborri:
@miigotu is there a reason why have you closed this PR?

miigotu:
16 commits, out of date, db version change, unmergable pr cleaning.

it is possible that this may be re-opened in the future, after we do the next db upgrade.  The next db upgrade will be planned with several db changes in one version so that hopefully the need for a db upgrade will not be needed for a very long time. Hash will be in the next version bump after the chnges have been decided and finalized.

PR #1401, opened 2015-02-23
Added apikey to censoredformatter

abeloin:
This is needed because Newznab apikey isn't stored as key=value in a
section in config.ini.


fernandog:
@abeloin does it censor the other api keys names (notifiers and etc)?

_apikey
_api
api_key

abeloin:
@fernandog If it's stored as key=value in a section in config.ini you can check if it's hidden by look at is key starting at https://github.com/SiCKRAGETV/SickRage/blob/master/sickbeard/__init__.py#L500
If you see a ```censor_log=True``` in the ```check_setting_str``` then it mean it will be censored.

abeloin:
@fernandog I also have to check if some torrent providers are pushing sensitive data in the log.

PR #1403, opened 2015-02-23
Urllib quote keyerrors issue #708

joshjowen:


fernandog:
@abeloin we can merge it

PR #1404, opened 2015-02-24
Fix PR #1403 extra )

abeloin:


fernandog:
@abeloin travis error. is that a problem?

abeloin:
In this case no, the error happened because I committed the change before Travis had time to finish checking the build.

PR #1407, opened 2015-02-24
Added iTunes parser

Deddiekoel:
For show folders that are also managed by iTunes (re)scanning existing episodes can fail. Added a simple regex to match the format used by iTunes: [Show Name]\\[Season folder]\\[Ep#] [Epsiode name]

This regex was put first as otherwise other regexes could get in the way.

### DIFF
--- a/sickbeard/name_parser/regexes.py
+++ b/sickbeard/name_parser/regexes.py
@@ -19,6 +19,12 @@
 # all regexes are case insensitive
 
 normal_regexes = [
+    ('itunes',
+     # 01 Ep Name
+     '''
+     ^(?P<ep_num>\d{1,3})[ ](?P<extra_info>.*)$      # 02 Episode Name
+     '''),
+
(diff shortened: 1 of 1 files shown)


abeloin:
This PR will not be merge:
A broad regex like that shouldn't be first, it should be the last one.
E.g.: 450 chemin du golf.S01E01.L'arrivée en banlieue.mkv will match with this regex.

fernandog:
@Deddiekoel please reply

cvandenthillart:
Ok, moving it down prevented it from working with me as the default season_only regex would pick it up.
Leave is pull-request for now. I'll try to get it to work and create a new pull request.

Deddiekoel:
Ok, moving it down prevented it from working with me as the default season_only regex would pick it up.
Leave is pull-request for now. I'll try to get it to work and create a new pull request.

PR #1408, opened 2015-02-24
Overhaul the network logos system

abeloin:
The current system as limitation with the system code page.
This pull request remove this limitation by using uuid to identify network url.
The current removal of accent è and é and tolower() are still in place but should be removed in the future.


neoatomic:
Isn't it possible to let SR only switch to uuid if a network logo is not found in the old system.?

abeloin:
@neoatomic What do you have in mind?

abeloin:
By the way Tokyo Broadcasting System has the tbs logo.

neoatomic:
@abeloin 
I would like to keep the old system and see if it can be "extended". So in case no logo is found it falls-back to the UUID format and tray's if it can find a logo that way. Would that be technically possible.?

If we implement UUID for all logos it will also have some cons.
(Its more complex, less user friendly, not compatible with other forks, might discourage users from creating there own logos. etc. )

When i take a quick look at the most used networks names on TVDB its seems that there are almost no network names that would course problems.
Currently we have only 1 logo that isn't working, so I'm wondering if its worth it.?

abeloin:
@fernandog Any thought on this?

fernandog:
the "not compatible with other forks" its a good point. Don't know how many forks are. It's just one, they will adapt to uuid.
what's t he PROs of UUID?

abeloin:
@neoatomic It would be technically possible, but should we? That is the big question.

We have more then one logo that as trouble:é,è and russian as of now. The first 2 were "patched" but those are stop gap measure and the third can't be fixed. Also we have a lot of duplicate icons.
The current system it's unable to cope with name collision e.g: tbs(Japon) and tbs(USA) as we add more indexer in the future. So if we can get ahead of this before it become a major problem it would be great. Ref: [network name list](https://raw.githubusercontent.com/Prinz23/sb_network_timezones/master/network_timezones.txt)

Maybe the UUID filename isn't the best choice but we need something that can separate the network name from the filename and that is why I created this PR so that we can discuss the issue.

Ultimately, I would like the data be stored in a sqlite db(instead of being hardcoded) so that we can do changes via the webUI(add, remove, update) but that would require an update mechanism.

In the end, you don't have to worry as this PR wont be applied soon. ;)

fernandog:
@echel0n what do you think? you ask me to tell you about major changes. I think this PR its a good ideia.

abeloin:
> what's t he PROs of UUID?
http://en.wikipedia.org/wiki/Universally_unique_identifier

Basically we need to keep the filename in the ascii table(128 chars) and must be unique.

That's is why I choose UUID which is easy to generate, but it can be any char in the ascii table as long it is unique.

neoatomic:
I defiantly see the pro's of this system, and also like your idea of using a sql database.
But i agree with fernadog that we need echel0n's input fist before we do any changes.

Also i have a list of about 60+ logos that i still need to create/add. Will try to PR them as soon as i can.

mbxplex:
@neoatomic Found a TV Show with a missing network logo.

A Football Life
Network: NFL Network

miigotu:
It's been working ok for now, so we will just close this. If The need for it pops up in the future the info/history is still here.

PR #1415, opened 2015-02-24
Small change: add label for deleteFile

theemstra:
When deleting a file, text should be clickable.

### DIFF
--- a/gui/slick/js/confirmations.js
+++ b/gui/slick/js/confirmations.js
@@ -47,7 +47,7 @@ $(document).ready(function () {
 		var showname = document.getElementById("showtitle").getAttribute('data-showname');
 		$.confirm({
 			'title'		: 'Remove Show',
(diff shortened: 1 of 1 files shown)


fernandog:
Did you checked if the size of the path+filename wont make the dialog box larger or unreadable? Can you send a screebshot showing this?
thanks

theemstra:
This will not change anything about the size of the path or any filenames.
This is just a usability fix. Screen:

theemstra:
After clicking the red text:
image

fernandog:
I see. sorry! thanks!

abeloin:
@fernandog I'm wondering if it's a good idea or not because now the user can click anywhere on the red text to activate the deleting of all files. Any thought?

fernandog:
@abeloin good point. let's not merge it. this way user must be really sure and click only in the little box.

theemstra:
I do still think that in the usability is pretty important. The user has to click the "Delete" button, then click the red line and the button to actually delete anything.

PR #1417, opened 2015-02-25
PP Syncfiles, List_associated_files fix, and Delete RAR contents fix

TagForce:
Fixes the following:
- List_associated_files also looks in subfolders to allow metadata in subfolders to be renamed properly. This causes associated_files for Post Processing (not renaming) to also include subfolders, which in turn can lead to files being deleted that are still in the queue to be processed.
- Delete RAR contents uses delete_files, which checks if result is true. If it isn't then it won't delete rar contents. It will now Delete RAR contents regardless if the process_method is not 'move'.

Adds the following:
- Config option under Post Processing to add your own syncfile extensions in a comma seperated value list.
- Changed is_syncfile function to accept a list, rather than hardcoded values.

### DIFF
--- a/gui/slick/interfaces/default/config_postProcessing.tmpl
+++ b/gui/slick/interfaces/default/config_postProcessing.tmpl
@@ -123,12 +123,21 @@
                                 <span class="component-desc">Move srr/srt/sfv/etc files with the episode when processed?</span>
                             </label>
                         </div>
-						
+						<div class="field-pair">
+                            <label class="nocheck">
+                                <span class="component-title">Sync File Extensions</span>
+                                <input type="text" name="sync_files" id="sync_files" value="$sickbeard.SYNC_FILES" class="form-control input-sm input350" />
+                            </label>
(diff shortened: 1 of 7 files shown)


fernandog:
This will PP only 1.mkv and not 2.mkv 2.!sync ?
thanks!

TagForce:
@fernandog Nope, not yet. This just adds an option to input your own syncfiles, which is easier than us maintaining the list of syncfiles in the code.

The problem with checking the syncfiles for associated files is that they are associated files, and we're already well into the process run before we get to those.
I'm still working on fixing that part, but this was already done so I commited it.

fernandog:
@TagForce shouldn't sickbeard.SYNC_FILES have default values? the actual ones?

TagForce:
@fernandog https://github.com/TagForce/SickRage/blob/68ede901053685e9f71edf5ea6f651c74694d384/sickbeard/__init__.py#L492
They are set.

PR #1418, opened 2015-02-25
Use WayBack Machine to search dead EZTV RSS provider

matigonkas:
Please do not merge this Pull Request as its more of a proof of concept .It is more as an ideia that anything else (Also It wont work for shows older than 2012, and most wont even work in the first place...)


fernandog:
Eztv is really gone?

matigonkas:
No it isn't you can find it at eztv.ch  but the thing is their RSS is still offline...

PR #1419, opened 2015-02-26
Possible fix for Fix TypeError: object of type 'NoneType' has no len()

abeloin:
Possible fix for Fix TypeError: object of type 'NoneType' has no len()
See https://github.com/SiCKRAGETV/sickrage-issues/issues/231 and https://github.com/SiCKRAGETV/sickrage-issues/issues/764

### DIFF
--- a/sickbeard/webserve.py
+++ b/sickbeard/webserve.py
@@ -186,10 +186,29 @@ class BaseHandler(RequestHandler):
                                </html>""" % (error, error, trace_info, request_info, sickbeard.WEB_ROOT))
 
     def redirect(self, url, permanent=False, status=None):
+        """Sends a redirect to the given (optionally relative) URL.
+
+        ----->>>>> NOTE: Removed self.finish <<<<<-----
+
+        If the ``status`` argument is specified, that value is used as the
+        HTTP status code; otherwise either 301 (permanent) or 302
+        (temporary) is chosen based on the ``permanent`` argument.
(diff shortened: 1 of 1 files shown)


abeloin:
@fernandog FYI

abeloin:
Ref #1247

OmgImAlexis:
Pulled it and that seems to have fixed all the errors while using nginx.

abeloin:
@OmgImAlexis Thank you for the testing and feedback.

OmgImAlexis:
I'm not sure if you were trying to just fix the nginx issue but this also fixed the ``object of type 'NoneType' has no len()`` issue I was getting from running manual post processing.

abeloin:
Nginx only exacerbated the thread safety issue with the missing ioloop.callback. It should fixes all related error like this one:
```
File "/home/media/sickrage/tornado/iostream.py", line 1383, in _merge_prefix
    if len(deque) == 1 and len(deque[0]) <= size:
TypeError: object of type 'NoneType' has no len()
```

OmgImAlexis:
Okay good, I'll let it run tonight(it's 11:40AM here) as that's usually when I get the 100s of errors and let you know if it's fixed then.

abeloin:
@OmgImAlexis Any issues so far?

OmgImAlexis:
Sorry I got carried away with trying other pull requests and such. This worked fine.

PR #1430, opened 2015-02-27
Fix Exception generated: Cannot find attribute 'contentrating'

TagForce:
- Fixes POSTPROCESSER :: Exception generated in thread POSTPROCESSER:
Cannot find attribute 'contentrating' errors in Mede8ter metadata
creation.


abeloin:
Ref https://github.com/SiCKRAGETV/sickrage-issues/issues/782

PR #1431, opened 2015-02-27
Fix Show Lookups on Indexer

TagForce:
- Now actually returns the correct show instead of nothing for post
processing.

Fixes:
Shownames not being found with "Show not in DB, you need to add it..."
messages during Post Processing. (most notably Forever (US) or Forever (2014))

### DIFF
--- a/sickbeard/helpers.py
+++ b/sickbeard/helpers.py
@@ -288,21 +288,22 @@ def searchIndexerForShowID(regShowName, indexer=None, indexer_id=None, ui=None):
                 continue
 
             try:
-                seriesname = search.seriesname
+                seriesname = search[0]['seriesname']
             except:
                 seriesname = None
 
             try:
-                series_id = search.id
+                series_id = search[0]['id']
             except:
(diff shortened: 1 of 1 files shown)


fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/25

OmgImAlexis:
Tested and I can now post-process bob's burgers and forever. SiCKRAGETV/sickrage-issues#780 seems fixed.

dels78:
Not trying to wake up the dead here, but I am having the same issue on a show in French (Canada) called Unité 9. Sickrage is searching for Unité 9 (2012) and won't download any episode and to post-process manually I need to add .2012 in the file name else it won't parse.

OmgImAlexis:
Do you have a scene exception for it added?

dels78:
Oh my god... I had not seen this option there (just switched from SB last week), my bad, I just added one and it worked right away. My apologies, and thanks for the help!

PR #1432, opened 2015-02-27
Add syncthing's ".syncthing.*" files to isSyncFile().

blasphemy:
Syncthing is an open source alternative to programs like BT Sync. It would be helpful for these files to be ignored by default.


fernandog:
@blasphemy check this merge PR

extension is no more hardcoded. there's a new option in UI
https://github.com/SiCKRAGETV/SickRage/pull/1417/

PR #1434, opened 2015-03-01
Add "Use failed downloads" to search settings

fernandog:


abeloin:
@fernandog That won't work, as it isn't saved in searchsaved

fernandog:
@abeloin I need to do something or it's not possible at all?

abeloin:
You need to add those 2 settings in searchSaved so they be saved.

PR #1437, opened 2015-03-01
Restart with exec

dgholz:
A feature branch to fix SiCKRAGETV/sickrage-issues#794


dgholz:
Bogus, should have used develop as PR target branch...

PR #1446, opened 2015-03-02
Change Init Scripts to Reflect Sick Rage Naming

fernandog:
TO-DO:
- Change SickBeard.py to SickRage.py
- Change sickbeard.db to sickrage.db

https://github.com/SiCKRAGETV/sickrage-issues/issues/45

### DIFF
--- a/init.debian
+++ b/init.debian
@@ -1,15 +1,15 @@
 #!/bin/sh
 #
 ### BEGIN INIT INFO
-# Provides:          sickbeard
+# Provides:          sickrage
 # Required-Start:    $local_fs $network $remote_fs
 # Required-Stop:     $local_fs $network $remote_fs
 # Should-Start:      $NetworkManager
 # Should-Stop:       $NetworkManager
 # Default-Start:     2 3 4 5
 # Default-Stop:      0 1 6
-# Short-Description: starts instance of SickBeard
-# Description:       starts instance of SickBeard using start-stop-daemon
+# Short-Description: starts instance of SickRage
+# Description:       starts instance of SickRage using start-stop-daemon
 ### END INIT INFO
 
 # Load the VERBOSE setting and other rcS variables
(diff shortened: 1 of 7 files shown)


abeloin:
@fernandog it won't work you renamed SickBeard.py to sickrage.py

fernandog:
Yes. the TO DO is to change that file too. Dont know the impact in SR code.

TagForce:
SR can't change the init scripts in the /etc/init.d folder unless it's run as root, which is something that is really not recommended.
Auto updating this will not work, as after a reboot the installed init scripts will still look for sickbeard.py.

Can this be done? Yes, but not using auto-update. It would require user interaction or a script that is run as root that loops through all the init files for all the platforms and changes every instance of sickbeard to sickrage. And it's not just the files that are supplied by sickrage either. There's an /etc/defaults/sickbeard file that has values for system variables used by SR on Linux that should be changed, It can be added to the xsession.(L, X, K)ubuntu startup apps etc.

Changing the filenames will not affect the inner workings of SickRage if you change all the occurances in the code, but it WILL break system configurations, and there's no real way to find out which files we need to change (we can add a lot of them, but we can't figure out all of them. There's a million and one options to concider).

All in all, I think it's too much of a hassle, as it would require users to basically re-install and re-configure their systems.

Carlosgrr:
Since I am being referenced on this, I'd like to comment the reasons why we can't go ahead:

>SR can't change the init scripts in the /etc/init.d folder unless it's run as root, which is something that is really not recommended.

Indeed, we are not going to change the already running SickRage installations, only those newer, that's how I see it, so only initscripts are changed.

>Auto updating this will not work, as after a reboot the installed init scripts will still look for sickbeard.py.

Also Indeed, don't change the filename yet, only after a full merge of sickbeard, or until someone recode the whole thing outside of it... (I don't think this will ever happen...)
My PR actually address this problem mantaining the Sickbeard.py and Sickbeard.db intact.

>Can this be done? Yes, but not using auto-update. It would require user interaction or a script that is run as root that loops through all the init files for all the platforms and changes every instance of sickbeard to sickrage. And it's not just the files that are supplied by sickrage either. There's an /etc/defaults/sickbeard file that has values for system variables used by SR on Linux that should be changed, It can be added to the xsession.(L, X, K)ubuntu startup apps etc.

Well, please, DO IT, only new installs are actually affected by this PR, so no worries, and in case anyone try to update it's code, nothing is changed in it's initscript.

>Changing the filenames will not affect the inner workings of SickRage if you change all the occurances in the code, but it WILL break system configurations, and there's no real way to find out which files we need to change (we can add a lot of them, but we can't figure out all of them. There's a million and one options to concider).

Well, I believe a first step would be to change the initscript's so that there is one less problem in sight, what do you think?

>All in all, I think it's too much of a hassle, as it would require users to basically re-install and re-configure their systems.

Nope, only those interested in a more customized messaging on their boot-up, stating that SickRage has been started.

PR #1448, opened 2015-03-02
Fetch all log files when viewlogs or searchLog

fernandog:
No more excuses "there is nothing in the log" =)

### DIFF
--- a/gui/slick/interfaces/default/viewlogs.tmpl
+++ b/gui/slick/interfaces/default/viewlogs.tmpl
@@ -83,7 +83,7 @@ Filter log by: <select name="logFilter" id="logFilter" class="form-control form-
 #end for
 </select>
 Search log by:
-<input type="text" name="logSearch" id="logSearch" value="#if $logSearch then $logSearch else ""#" class="form-control form-control-inline input-sm" />
+<input type="text" name="logSearch" placeholder="Clear to reset" id="logSearch" value="#if $logSearch then $logSearch else ""#" class="form-control form-control-inline input-sm" />
 </div>
 <br />
 <div class="align-left"><pre>
--- a/sickbeard/webserve.py
+++ b/sickbeard/webserve.py
@@ -4827,10 +4827,11 @@ class ErrorLogs(WebRoot):
                 data = f.readlines()
 
         while len(data) <= maxLines :
-            for i in range (1 , int(sickbeard.LOG_NR)):
+            for i in range ( 1 , int(sickbeard.LOG_NR) ):
                 if os.path.isfile(logger.logFile + "." + str(i)):
                     with ek.ek(codecs.open, *[logger.logFile + "." + str(i), 'r', 'utf-8']) as f:
                         data += f.readlines()
+            break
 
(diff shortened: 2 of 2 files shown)


TagForce:
@fernandog 
Change the loop to check for len(data) < MaxLines last instead of first.
If len(data) < MaxLines in your code, it will open and add each and every logfile to data.

```
      for i in range (1 , int(sickbeard.LOG_NR)):
            if os.path.isfile(logger.logFile + "." + str(i)) and len(data) <= maxLines:
                       with ek.ek(codecs.open, *[logger.logFile + "." + str(i), 'r', 'utf-8']) as f:
                            data += f.readlines()
```

This way it will loop through every logfile number, but will only open them when the file actually exists, AND if we need to read lines from it.

fernandog:
@tagforce, @abeloin and I decided to change how log works. Its better to Read & Search than Read all lines & Search.  The logSearch/logfilter feature will fetch all log files available

TagForce:
@fernandog Keep in mind that reading ALL the lines from a file and then checking the results in memory is often faster than reading a file line by line, and checking each line as you go along.
So you might want to keep reading the entire logfile in one go, but have the file contents go into a placeholder function called extractlogs or something that runs through the entire filecontents and returns only the lines you want to show.

something like data += extractlogs(f.readlines())
with extractlogs returning the subset of lines you want.
This way, you won't be loading your memory with a lot of data (but only the data you're interested in), and the contents of the logfile in memory will be removed from memory after that line.

It will use more memory than doing it line by line (filecontents will still be loaded into memory entirely) but it will be much faster. By not actually storing the filecontents in a variable, but just passing it to a function should make it be released after the function has run, which means that it won't keep using all that memory.

fernandog:
@TagForce so the actual code is better then read one line & match regex & search?

anyway the logger must be re-written

Or we commit this to "fix" the log issue and in near future re-write it or re-write now.
The issue exists (open viewlog and show only a few lines (sickrage.log has only few lines)
@abeloin 

what should we do?

TagForce:
@fernandog https://github.com/TagForce/SickRage/commit/5ce313446258270e4d12eccf26e361b1202154e1
I think that does exactly what you want?
It reads the logfiles and immediately filters only the lines you want. It then counts only the lines that actually get added to the output as counting towards the maxLines.

I tested this on my system, and if I set the logs to something small (like 10kb) and have 15 logs, it will correctly continue with the next log if it's not reached maxLines yet. It needs some testing though.

fernandog:
@TagForce thanks! will test it

TagForce:
@fernandog It's the same code. Just placed at a different moment in execution. It will now pass through each logfile seperately and only returns lines that match the if statements for logFilter and logSearch.
It will also only count lines that are sent to the output, instead of each line checked.
This means that filtering on SEARCHLOG will filter all logs until it has 500 lines of SEARCHLOG entries, instead of all the SEARCHLOG entries in the first 500 lines of log. So you'll probably see older entries when filtering than without a filter.

fernandog:
@TagForce what if the search result has less than 500 lines? it has to stop until 500 lines or when there are no more log files

abeloin:
@fernandog If there is less than 500 lines, the for loop will stop and return finalData will be run.

fernandog:
@TagForce you can PR to SR! thanks!

PR #1452, opened 2015-03-03
freshontv: Limit number of pages for RSS search

flightlevel:
Search was iterating through 15 pages of results for the regular RSS cache search. Limited this to 3 pages.


fernandog:
@flightlevel is there any reason to reduce it?

flightlevel:
No need to perform 15 http requests every time the cache updates (there aren't that many new torrents to have to go pages and pages back). The multi page search was really added for backlog searches.

PR #1456, opened 2015-03-03
Remove traceback from generic.py

fernandog:
Show only error message. Not traceback

https://github.com/SiCKRAGETV/sickrage-issues/issues/830
https://github.com/SiCKRAGETV/sickrage-issues/issues/596
https://github.com/SiCKRAGETV/sickrage-issues/issues/251


Hellowlol:
Am i the only one that likes traceback directly on the page? :D

fernandog:
@Hellowlol I like to see it too. but with auto issue, user are opening issues that don't need to be open.

PR #1458, opened 2015-03-03
Compare remote DB versions with local before update

fernandog:
Added log messages to DEBUG comparisson

TODO:
- Compare DB when auto-updating
- Compare DB when manual-updating

@abeloin


fernandog:
@TagForce can you test this too? Just to be sure!

PR #1459, opened 2015-03-03
Added backup when updating

abeloin:
Keep the latest backup only.
Saved in DATA_DIR/backup


abeloin:
@fernandog If you can test it

PR #1460, opened 2015-03-03
Fix coloring episodes as good when "archive first match" selected

grad-michal:
Currently when "archive first match" is selected and custom quality is used downloaded episodes with quality in archive list but lower than best are colored as low quality (despite they wont be redownloaded). 

It also applies to qualities that are better than minimal archive quality but are not in archive list (for example when downloaded manually or quality preference was changed).


abeloin:
Thanks for the PR

PR #1465, opened 2015-03-04
Fix shutil import error

TagForce:
- Fixes NameError: name 'shutil' is not defined caused by monkeypatching
shutil.copyfile

### DIFF
--- a/sickbeard/__init__.py
+++ b/sickbeard/__init__.py
@@ -24,6 +24,7 @@ import socket
 import os
 import re
 import os.path
+import shutil
 import lib.shutil_custom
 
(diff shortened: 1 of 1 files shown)


TagForce:
@fernandog @abeloin Please merge this PR immediately, or fix the monkeypatch for shutil.copyfile another way, but be quick about it. In its current state it breaks SR completely without manually editing __init__.py to include above line.

abeloin:
@TagForce Thank You

fernandog:
what we do about python 2.6? Fallback to previous code?

abeloin:
Make a PR that fix the issue: https://github.com/SiCKRAGETV/SickRage/pull/1469

PR #1474, opened 2015-03-04
Fix list branches

fernandog:
@abeloin @TagForce


abeloin:
You need to rebase this commit

PR #1475, opened 2015-03-04
Change viewLog to use rolled over logfiles

TagForce:
- Changed the viewlog to read the log from all the logfiles in order if
the number of lines is < maxLines.
- Viewlog will now only count log entries that are significant for your
search entry or filter setting.

### DIFF
--- a/sickbeard/webserve.py
+++ b/sickbeard/webserve.py
@@ -4903,7 +4903,50 @@ class ErrorLogs(WebRoot):
         return self.redirect("/errorlogs/")
 
     def viewlog(self, minLevel=logger.INFO, logFilter="<NONE>",logSearch=None, maxLines=500):
+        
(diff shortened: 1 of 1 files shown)


fernandog:
@abeloin we can merge it

fernandog:
@TagForce @abeloin don't know if you didn't notice, but some [App submitted] issues don't have LOGs uploaded

here:
https://github.com/SiCKRAGETV/SickRage/blob/develop/sickbeard/logger.py#L165
and 
https://github.com/SiCKRAGETV/SickRage/blob/develop/sickbeard/logger.py#L177

if fetches 500/50 lines of the log.
but if log rotated and the ERROR is in the next log, SR won't be able to fetch it, correct?
so that's why some issues don't have uploaded log.

can you confirm?

PR #1476, opened 2015-03-04
Change list_associated_files

TagForce:
- Changed setting base_name to use the old way when subfolders is not
true.


abeloin:
Ref https://github.com/SiCKRAGETV/sickrage-issues/issues/842

PR #1485, opened 2015-03-05
Redo metadata assignments for mede8er

TagForce:
- Redid most of the metadata assignments using getattr() to keep it from
throwing exceptions when info is not available for a certain show or
episode.


TagForce:
@fernandog Actually is a bugfix, not a feature. The current mede8er.py generates exceptions on almost any info tvrage lacks. This fixes that.

PR #1489, opened 2015-03-05
Ability to delete media files with status change

Eiber:
Feature: 
Ability to let SickGear delete media files automatically when the status of an episode is changed to the desired state (instantly or with a delay time in minutes to prevent mistakes). The desired state could be archived, ignored or skipped and maybe in the feature a new status: watched. This feature may be useful when you don't have fysical access to your media server, and still would like to delete the watched episodes using the web page or API.

How it works:
1. activate/deactivate ability to delete episodes with a desired state on the postprocess config page
2. choose delay time (in minutes) so any mistake in status change can be corrected
3. choose to either delete files permanently or send to trash first so files can be recoverd.
4. after a mistake the planned deletion can be canceled on the displayshow page

There is one thing not working yet, but I don't know you to fix:

The file ajaxEpCancelDelete.js needs to be created in gui/slick/js to show a loading picture and prevent the page from following the link. I've tried but JS I don't know anything about JS. Maybe someone can help?

### DIFF
--- a/gui/slick/css/style.css
+++ b/gui/slick/css/style.css
@@ -1113,6 +1113,10 @@ ul.tags li a{
 	vertical-align: -2px;
 }
 
+.delete {
+  background-color: #B2B2B2 !important;
+  text-decoration: line-through;
+}
 .unaired {
   background-color: #f5f1e4;
 }
@@ -2535,6 +2539,11 @@ input sizing (for config pages)
 
 .btn-inline {
 	margin-top: -3px;
+}
+
+.input50 {
+	width: 50px;
+	margin-top: -4px;
 }
 
 .input75 {
--- a/gui/slick/interfaces/default/config_postProcessing.tmpl
+++ b/gui/slick/interfaces/default/config_postProcessing.tmpl
@@ -34,7 +34,8 @@
                 	<li><a href="#core-component-group3">Metadata</a></li>
                 </ul>
                 
-                <div id="core-component-group1" class="component-group">
+                <div id="core-component-group1">
+				<div class="component-group">
 
                     <div class="component-group-desc">
                         <h3>Post-Processing</h3>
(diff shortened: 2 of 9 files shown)


lib/subliminal/services/addic7ed.py
```diff
@@ -39,7 +39,7 @@ class Addic7ed(ServiceBase):
     api_based = False
     #TODO: Complete this
     languages = language_set(['ar', 'ca', 'de', 'el', 'en', 'es', 'eu', 'fr', 'ga', 'gl', 'he', 'hr', 'hu',
-                              'it', 'pl', 'pt', 'ro', 'ru', 'se', 'pb'])
+                              'it', 'nl', 'pl', 'pt', 'ro', 'ru', 'se', 'pb'])
```
abeloin: Not sure why this is here?

Eiber:
That's unrelated to the feature. Should have been in a different PR, but it doesn't really matter actually. It only adds support for an additional language for Addic7ed

fernandog:
@Eiber skip_removed_files must be enabled. Otherwise removed episodes will be set to IGNORED in the next SHOWUPDATE (re-scan)

https://github.com/SiCKRAGETV/SickRage/blob/develop/sickbeard/tv.py#L1085

Eiber:
@fernandog That's not necessary. Not only the media episode files are deleted, but also the reference in the database. Sickrage won't know the files were there, and thus won't know they are missing.

fernandog:
@Eiber you do the same thing that code does? all database fields?

Eiber:
Well, I've just checked. It doest the same thing, but not in all database fields. The following are missing:
- curEp.subtitles_searchcount
- curEp.subtitles_lastsearch 
- curEp.hasnfo = False
- curEp.hastbn = False

Those are not really important, but I will add them..

Eiber:
All necessary database fields are now included. I also removed the additional language for addic7ed as it is unrelated to this PR

PR #1493, opened 2015-03-06
Added scraper for MoreThan.TV.

Sparhawk76:
Created a scraper for morethan.tv, this time changes are against develop


fernandog:
@Sparhawk76 any issues with this since you PR? Have you tested this in the past 11 days?

Sparhawk76:
I've been using it for the past couple weeks ... reliably returns new shows as they release on MoreThan.tv.  There is one minor issue that I am trying to arrange with the MoreThan.tv devs to fix, in that their site does not return any results if you search for Show.Name S01 ... when there are many individual episodes for that Show/Season.  So searching for a whole season at once doesn't work, but searching for individual episodes of a season does.

abeloin:
@Sparhawk76 Have a look in https://github.com/SiCKRAGETV/SickRage/blob/master/sickbeard/providers/iptorrents.py as it also doesn't support season search.

You need to copy/modify/implement your own findSearchResults which will only search for episode.

Sparhawk76:
Ok ... I may still need to make one for this case. The devs of morethan.tv got back to me and if you add a * to the search it will still work, eg search for "Show Name S01*" will return all episodes in s01

PR #1494, opened 2015-03-06
Use Scene Exceptions in Post Processing

TagForce:
- Changed NameParser to include trySceneExceptions boolean
- Changed calls to NameParser from Post Processing function to include
trySceneException=True when tryIndexers is also True.
- Changed get_show function to try looking up a parsed showname in the
scene exceptions table (using get_scene_exception_by_name) if both cache
and Indexer lookups fail and trySceneException is True.

As found here:
https://github.com/SiCKRAGETV/sickrage-issues/issues/25
https://github.com/SiCKRAGETV/sickrage-issues/issues/769
https://github.com/SiCKRAGETV/sickrage-issues/issues/582
https://github.com/SiCKRAGETV/sickrage-issues/issues/356
https://github.com/SiCKRAGETV/sickrage-issues/issues/864

And others.

### DIFF
--- a/sickbeard/helpers.py
+++ b/sickbeard/helpers.py
@@ -917,7 +917,7 @@ def _check_against_names(nameInQuestion, show, season=-1):
     return False
 
 
-def get_show(name, tryIndexers=False):
+def get_show(name, tryIndexers=False, trySceneExceptions=False):
     if not sickbeard.showList:
         return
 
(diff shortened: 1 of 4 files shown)


fernandog:
Nice! PP seems very stable and with this I think we close all issues about it.

TagForce:
I made a boo-boo... Can you merge the next one as well?
It throws an exception when no show is found from the scene_exception list. Latest PR fixes that.

PR #1495, opened 2015-03-06
Fixed exception on failing to find show in Scene Exceptions

TagForce:
Whoops... I failed to account for shows actually not being in the scene exceptions. Since it then passes a None to int(), that will throw an exception.
- Assigns ShowID instead of direct calling it. If not None, then it will pass to findCertainShow as an int.


TagForce:
https://github.com/SiCKRAGETV/SickRage/pull/1494

TagForce:
@fernandog thanks... stupid me.

PR #1496, opened 2015-03-06
Fix for PP always returning "Problem(s) during Processing"

TagForce:
- result.result was falsely assigned the value from the process function
return, which doesn't return anything.
- Will now return the correct strings (Problem when there was a problem,
or succesfully Processed if not)

This problem was exaggerated by changing the function and copying the
erroneous line to all methods.


fernandog:
@Tagforce Can I merge it?

TagForce:
@fernandog Yes, it just removes the result.result assignments that caused the result.result to always be None if the file was in the root PP folder.

PR #1500, opened 2015-03-08
Possible fix to always show collapsed menu menu on mobile

fernandog:
**Pro**: collapsed menu will be always in the screen. Don't need to scroll
**Con**: user can't zoom-in / zoom-out in the phone (or when menu is collapsed. (I still can zoom-in/zoom-on in Chrome for desktop)

http://stackoverflow.com/questions/27319200/bootstrap-navbar-covers-content-on-zoom

Don't know if we can to this too:
http://stackoverflow.com/a/257972

***Before this change:***

***After this change:***

@joshjowen can you check this ? What do you think?


joshjowen:
I think that taking away the ability to zoom will cause to many us ability problems but it did fix the menu for me

PR #1503, opened 2015-03-08
rarbg: switch to request library

abeloin:
Clean up code


rarbg:
https://rarbg.com/pubapi/apidocs.txt
I think this might be useful , instead of doing all that parsing.

abeloin:
@rarbg Thank You for the information

PR #1504, opened 2015-03-09
Dont backlog when paused

TagForce:
- Will skip adding the episode to the backlog search when changing
status to Wanted and the show is paused.
- Adds log message "some episodes were set to Wanted but show is paused.
Not adding to backlog until show is unpaused"


fernandog:
@TagForce shouldn't we add paused check here too?

https://github.com/SiCKRAGETV/SickRage/blob/develop/sickbeard/tv.py#L1267

I coudn't add. Information is not there.

PR #1508, opened 2015-03-09
Fix mede8er rating info

TagForce:
- Changed mede8er myEp['rating'] assignment.

I forgot that one... Please merge into develop and master.


fernandog:
@abeloin another hotfix

abeloin:
@TagForce When it's a hotfix(meaning it is to be applied to both branch) commit to master not develop.

PR #1509, opened 2015-03-09
Remove traceback from showupdater

fernandog:
and add more info to log message
https://github.com/SiCKRAGETV/sickrage-issues/issues/370
@TagForce

### DIFF
--- a/sickbeard/show_queue.py
+++ b/sickbeard/show_queue.py
@@ -78,16 +78,13 @@ class ShowQueue(generic_queue.GenericQueue):
     def updateShow(self, show, force=False):
 
         if self.isBeingAdded(show):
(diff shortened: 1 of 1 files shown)


fernandog:
@TagForce so this is wrong? so why SR is trying to update twice the show?

fernandog:
@TagForce  but why SR is trying to update twice the show?

TagForce:
@fernandog I have no idea

fernandog:
@TagForce are you sure that when PP finishes it did'nt start a show-refresh to updated the status and quality of the new processed episode?

TagForce:
@fernandog I do think that it should log the message and go to the next show, though. So your PR is fine except the string

fernandog:
updated

PR #1517, opened 2015-03-10
history-television

flightlevel:
Added icon for History Television


fernandog:
@neoatomic size is ok?

neoatomic:
@fernandog Yes, its fine.

PR #1518, opened 2015-03-10
Send user locale when submitting issues

fernandog:
@abeloin should we consider NAS/FreeNAS/etc too ?

http://stackoverflow.com/questions/842059/is-there-a-portable-way-to-get-the-current-username-in-python

### DIFF
--- a/sickbeard/logger.py
+++ b/sickbeard/logger.py
@@ -24,6 +24,7 @@ import logging
 import logging.handlers
 import threading
 import platform
+import locale
 
 import sickbeard
 from sickbeard import classes, encodingKludge as ek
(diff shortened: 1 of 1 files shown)


fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/910

fernandog:
@abeloin 

Just for fun I executed this script https://gist.github.com/fernandog/5c3a650afc3f3a84cf9b to list all OS from all issues with have:
```
Darwin
FreeBSD
FreeNAS
Linux
OSX
Ubuntu
Windows
```

only Windows don't have locale?

abeloin:
@fernandog You shouldn't include the username in the the data sent.

As far as I know, yes windows is the only one without locale

fernandog:
@abeloin ok, will remove it

PR #1523, opened 2015-03-10
Fix submit issue not reading all log files

fernandog:


fernandog:
@abeloin will squash commits.

PR #1528, opened 2015-03-11
Fix deluge move completed default value

Toilal:
Use deluge configuration when `move_completed_path` is not defined in sickrage.

With actual code, if deluge is configured to move downloaded files to another directory, it will be overriden with a blank path making deluge keep files in initial download location.

Tested OK !

### DIFF
--- a/sickbeard/clients/deluge.py
+++ b/sickbeard/clients/deluge.py
@@ -97,8 +97,7 @@ class DelugeAPI(GenericClient):
     def _add_torrent_uri(self, result):
 
         post_data = json.dumps({"method": "core.add_torrent_magnet",
-                                "params": [result.url, {"move_completed": "true",
-                                                        "move_completed_path": sickbeard.TV_DOWNLOAD_DIR}],
+                                "params": [result.url],
                                 "id": 2
         })
         self._request(method='post', data=post_data)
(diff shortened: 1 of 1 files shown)


fernandog:
@Toilal thanks! and welcome to SR (never saw you here before). feel free to help code it

Toilal:
Thanks @fernandog, i'm new yes !

I'm including SickRage in a all in one docker image, maybe you will be interested : https://github.com/Toilal/stealthbox.

fernandog:
Is it fully tested?

abeloin:
@fernandog Not sure we should merge this PR as the code assume deluge is run on the same machine as sickrage. No other clients those that. I think we should do a PR removing setting move_complete on sickbeard.TV_DOWNLOAD_DIR and let the user use sickbeard.TORRENT_PATH instead.

Toilal:
:+1:

I Didn't know the content of TV_DOWNLOAD_DIR variable, I've just change to make it work when the the variable is undefined.

Do you want me to replace with TORRENT_PATH variable ?

abeloin:
@Toilal No, as I think we should remove the code your PR removed and the code you added. This way deluge default move on complete is applied(label or not) and if the user want to override the default, it can be set in the webui via `Downloaded files location` in the torrent client config.

Deluge (_set_torrent_path)[https://github.com/Toilal/SickRage/blob/fix/deluge-move-completed/sickbeard/clients/deluge.py#L198]

Référence:
TV_DOWNLOAD_DIR = Post-process TV Download Dir
TORRENT_PATH  = Downloaded files location
Deluge (_set_torrent_path)[https://github.com/Toilal/SickRage/blob/fix/deluge-move-completed/sickbeard/clients/deluge.py#L198]

Toilal:
Ok i'll update abeloin, I understand (after reading the whole file, it's clear).

Toilal:
It must be OK now.

PR #1529, opened 2015-03-11
Set amActive to False when return

fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/821


fernandog:
@abeloin hotfix or not ?

abeloin:
@fernandog No need it's not yet in master

PR #1531, opened 2015-03-11
Use consistent behavior in word filtering (global and show settings)

Toilal:
Following discution from https://github.com/SiCKRAGETV/sickrage-issues/issues/919

This implements consistent word filtering for global words filtering and show words filtering. In both case, regexp are escaped and only full words are catched.

This has been refactored using a single function `show_name_helpers.containsAtLeastOneWord`. Also labels in UI have been updated accordingly.

### DIFF
--- a/sickbeard/show_name_helpers.py
+++ b/sickbeard/show_name_helpers.py
@@ -37,6 +37,25 @@ resultFilters = ["sub(bed|ed|pack|s)", "(dk|fin|heb|kor|nor|nordic|pl|swe)sub(be
                  "(dir|sample|sub|nfo)fix", "sample", "(dvd)?extras",
                  "dub(bed)?"]
 
+
+def containsAtLeastOneWord(name, words):
+    """
+    Filters out results based on filter_words
+
+    name: name to check
+    words : string of words separated by a ',' or list of words
+
+    Returns: False if the name doesn't contain any word of words list, or the found word from the list.
+    """
+    if isinstance(words, basestring):
+        words = words.split(',')
+    items = [(re.compile('(^|[\W_])%s($|[\W_])' % re.escape(word.strip()), re.I), word.strip()) for word in words]
+    for regexp, word in items:
+        if regexp.search(name):
+            return word
+    return False
+
+
 def filterBadReleases(name, parse=True):
     """
     Filters out non-english and just all-around stupid releases by comparing them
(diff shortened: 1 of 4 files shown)


sickbeard/show_name_helpers.py
```diff
@@ -37,6 +37,25 @@
...
+    Filters out results based on filter_words
+
+    name: name to check
+    words : string of words separated by a ',' or list of words
+
+    Returns: False if the name doesn't contain any word of words list, or the found word from the list.
+    """
+    if isinstance(words, basestring):
+        words = words.split(',')
+    items = [(re.compile('(^|[\W_])%s($|[\W_])' % re.escape(word.strip()), re.I), word.strip()) for word in words]
```
Toilal: What about replacing `(^|[\W_])` with `\b` which means "word boundary" ?

Toilal: Forget about this, after more tests this doesn't seems to work as well.

fernandog:
Can you edir/add a label explaining its OR ?

Toilal:
Currently it's "Results without one of these words in the title will be filtered out" for required words, and "Results with any of these words in the title will be filtered out" for ignored words.

It seems clear to me, but maybe you have an idea to make it more clear ?

fernandog:
@Toilal if is clear, don't bother

Toilal:
Maybe it's better :

ignored words => "Results with one or more word from this list will be ignored."
required words => "Results with no word from this list will be ignored."

I'll update

fernandog:
Is it fully tested?

sickbeard/show_name_helpers.py
```diff
@@ -59,24 +78,21 @@ def filterBadReleases(name, parse=True):
     #    return False
 
     # if any of the bad strings are in the name then say no
+    ignore_words = list(resultFilters)
     if sickbeard.IGNORE_WORDS:
-        resultFilters.extend(sickbeard.IGNORE_WORDS.split(','))
```
Toilal: I think I've fixed a memory leak here, because content of global variable `resultFilters` is growing at each run of `filterBadReleases`. Fix is to create a new list from `resultFilters`.

fernandog:
@Toilal everything working? this possible memory leak was fixed too?

Toilal:
Yes, I've tested and it's working.

fernandog:
@Toilal can you update the "ignored/required words" text in editshow page too?

PR #1525, opened 2015-03-10
Travis-CI has not actually been running the tests.

anthonyryan1:
When `./all_tests.py` is run from outside of the tests directory it fails to run the
majority of the test suite. This is either something that needs to be fixed in
all_tests.py or we need to cd into the tests dir before running them.

This pull request does the later, but if someone would rather solve this problem in all_tests.py this can serve as a place to discuss that.

PR #1533, opened 2015-03-11
Fix feedparser import error

fernandog:
@anthonyryan1 did you get this error?

https://github.com/SiCKRAGETV/SickRage/pull/1525#issuecomment-78362472

I'm also getting a lot of
```
 File "/home/pi/SickRage/sickbeard/db.py", line 170, in action
    sqlResult = self.execute(query, args, fetchall=fetchall, fetchone=fetchone)
  File "/home/pi/SickRage/sickbeard/db.py", line 92, in execute
    raise e
OperationalError: disk I/O error

```

even with SR shutdown

### DIFF
--- a/lib/feedcache/cache.py
+++ b/lib/feedcache/cache.py
@@ -32,7 +32,7 @@ __module_id__ = "$Id$"
 #
 # Import system modules
 #
-from feedparser import feedparser
+from lib.feedparser import feedparser
 
 import logging
 import time


anthonyryan1:
Hello @fernandog,

I did not encounter the feedparser error myself, but I did get the disk I/O errors.

fernandog:
@anthonyryan1 

>>> feedparser.__version__
'5.1.2'
>>>

can you check yours?

anthonyryan1:
I don't have feedparser installed globally on my system, I'm using the version in [lib/feedparser](https://github.com/SiCKRAGETV/SickRage/tree/master/lib/feedparser).

So from the root of the sickrage tree with python2 I get:
```
>>> from lib.feedparser import feedparser
>>> feedparser.__version__
'5.1.3'
```

It sounds like this particular failure you're running into comes from having feedparser installed on the system in addition to the version found in the library. Python is selecting the system version over the version that is bundled in the lib directory.

**Update:** When I install a copy of feedparser globally on my system I then run into the same error.

fernandog:
@anthonyryan1 changed to from lib.feedparser import feedparser

PR #1548, opened 2015-03-13
Changed failure handling in PostProcessing

TagForce:
- Will now continue if an error is encountered with a single file
- Will skip only the current folder when syncfiles are encountered
- Will keep track of failed items using an aggresult variable in
processResult()
- Will show a summary at the end of Post Processing listing all failed
items (files and folders) with the reason for failing.
- Will only set aggresult to False when actual processing fails for a
file, not if no processable items are found or files are already
processed (will show those in the summary)

### DIFF
--- a/sickbeard/processTV.py
+++ b/sickbeard/processTV.py
@@ -43,8 +43,10 @@ shutil.copyfile = lib.shutil_custom.copyfile_custom
 
 class ProcessResult:
     def __init__(self):
-       self.result = True
-       self.output = ''
+        self.result = True
+        self.output = ''
+        self.missedfiles = []
+        self.aggresult = True
 
 def delete_folder(folder, check_empty=True):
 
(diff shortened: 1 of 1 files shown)


TagForce:
This needs more testing, I couldn't test all possible events on my test rig, right now.

It should, however work as intended, and most common issues have been captured and tested.

This can be merged into develop, but wouldn't yet merge into master.

TagForce:
As far as I'm concerned this can be merged into develop. If you want me to clean it up and squash the commits into one, let me know. 

If anyone has encountered issues with this, also let me know.

I've been running this for a few days with multiple downloads and PP runs, and I've found no errors.

fernandog:
@TagForce sorry, didn't test this because I don't use PP

fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/709

PR #1560, opened 2015-03-14
[FIX] Download URLs upper hash and URL change

nicolasmartinelli:
- torcache and zoink need a hash in uppercase
- zoink is now zoink.ch
- torrage is down, so moved at the bottom of the list


nicolasmartinelli:
Ooops, I should have done my pull request on develop... Sorry about that.

PR #1568, opened 2015-03-15
Remove extra raise

fernandog:


abeloin:
Don't do that we need the raise to inform of the error

PR #1570, opened 2015-03-16
Support for Foreign Releases based on Show Language

cytec:
see https://github.com/SiCKRAGETV/sickrage-issues/issues/979

### DIFF
--- a/sickbeard/__init__.py
+++ b/sickbeard/__init__.py
@@ -495,7 +495,7 @@ DELETE_FAILED = False
 
 EXTRA_SCRIPTS = []
 
-IGNORE_WORDS = "german,french,core2hd,dutch,swedish,reenc,MrLss"
+IGNORE_WORDS = "core2hd,dutch,swedish,reenc,MrLss"
 REQUIRE_WORDS = ""
 SYNC_FILES = "!sync,lftp-pget-status,part,bts"
 
@@ -1469,7 +1469,7 @@ def halt():
                     ADBA_CONNECTION.join(10)
                 except:
                     pass
-            
+
             __INITIALIZED__ = False
             started = False
 
--- a/sickbeard/common.py
+++ b/sickbeard/common.py
@@ -94,7 +94,7 @@ class Quality:
     RAWHDTV = 1 << 3  # 8  -- 720p/1080i mpeg2 (trollhd releases)
     FULLHDTV = 1 << 4  # 16 -- 1080p HDTV (QCF releases)
     HDWEBDL = 1 << 5  # 32
-    FULLHDWEBDL = 1 << 6  # 64 -- 1080p web-dl                        
+    FULLHDWEBDL = 1 << 6  # 64 -- 1080p web-dl
     HDBLURAY = 1 << 7  # 128
     FULLHDBLURAY = 1 << 8  # 256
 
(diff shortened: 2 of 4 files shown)


fernandog:
@cytec why remove german and french but not dutch and swedish?

cytec:
@fernandog well i only knew search keywords for german and french, so i only added those... i could have guessed spanish, dutch, swedish etc but i think its better if someone reports and it gets changed then :)

supporting additional languages is now as easy as just adding the ISO Code and the REQUIRED words list to `languageMap` dict in `show_name_helpers`

fernandog:
@cytec shouldn't we add a NOTE in ignore words about show language feature?

cytec:
@fernandog maybe also change this line: https://github.com/SiCKRAGETV/SickRage/blob/master/gui/slick/interfaces/default/home_newShow.tmpl#L72 so it is more clear that changing the Language in fact will search for releases in that language at least for german/french/english with this changes

adding a NOTE somewhere that there are dynamically added ignore/required words when selecting a language also would be nice, but i don't know if and where to put that note to :/

My fork was basically targeting german users which all knew that changing the quality will result in require german in the release name. Maybe for a much larger and multi language userbase that will not be clear without explicit stating it somewhere.

Im open to suggestions on how to fix/change that. My suggestions:
Change the line mentioned above to reflect the recent changes to something like this:
`This will affect the language of the retrieved metadata and <b>WILL ALLOW</b> SickRage to download non-english TV episodes!`

and here: https://github.com/SiCKRAGETV/SickRage/blob/master/gui/slick/interfaces/default/editShow.tmpl#L79 
something like:

`Show Language: (this will dynamically add required/ignored words for the selected language)`

fernandog:
@abeloin thoughts?

abeloin:
@fernandog Not really

cytec:
@fernandog @abeloin i don't get it... first: `don't merge` tag, then it gets merged, then reverted... and not a single word about any of those decisions :/

Just curious... is this something you don't want or just not as it is right now?

fernandog:
@cytec not right now. We need to make sure our master is stable first. We have a lot of PR for new features that have the "don't merge". Then we need to understand all these new features and make Wikis explaining how to config them and etc. Otherwise we won't be able to support/reply issues from users. I hope you understand!

cytec:
@fernandog yup everything clear now, thanks for the response ;)

fernandog:
@cytec can you still make a wiki full explaining how this works?

will insert here:
https://github.com/SiCKRAGETV/SickRage/wiki

cytec:
I'll do

cytec:
@fernandog since i was unable to edit SickRage Wiki, i've made a page here:
https://github.com/cytec/SickRage/wiki/Multilanguage-Settings

fernandog:
@cytec nice! thanks!

cytec:
I've just added `spanish` and `italian` to the list of supported languages.
Also i'm currently working on language change in `massEdit`

fernandog:
@cytec we are working on a preferred words feature. Isn't better with that feature?

cytec:
@fernandog prefered words would be nice too... so one can prefer a release of a specific group :)

fernandog:
@cytec Or language. I prefer english but I accept italian for example.

LexMih:
Hello,

Would it be possible to add the same show twice without the error: "xxxxxx is already in your show list", so I can have for exemple the same show in English and French in the same library ?
Is there a way around ?

Thank you !

PR #1572, opened 2015-03-16
Fix HDTorrents

fernandog:
TODO: disable partial search string / season mode

@abeloin to disable season mode, should I only remove the entire "def _get_season_search_strings(self, ep_obj)" ?


abeloin:
@fernandog No you need to implement findsearchresult in HDTorrent and remove the part were if there more than 1 episode do a season search.

fernandog:
I did that because in IPT you removed that and added the warning message

abeloin:
Also added findSearchResult

PR #1573, opened 2015-03-16
Fix providers urllib quote and fixing title in log message

fernandog:


abeloin:
@fernandog All the providers modified in this this commit use requests, so there's no need to urllib.quote them.

PR #1576, opened 2015-03-17
Add option to delete files/folders when using manual Post Processing

TagForce:
- Adds a checkbox to delete files/folders just like auto Processing does
- Defaults to off


fernandog:
@TagForce is this a valid request?

https://sickrage.tv/forums/forum/main-category/feature-requests/20881-delete-files-folders-when-using-manual-post-processing

TagForce:
@fernandog That's going to be problematic... If you select your TV Download folder, it WILL delete everything in it. So, rather just delete folders that are empty, just to be sure.

fernandog:
@TagForce ok! Just want to make sure to reply to user that we won't do that

PR #1580, opened 2015-03-18
Add Provider Animenzb - A Fanzub Clone

Thraxis:
Had to re-create the PR...


fernandog:
@Thraxis should this commit be included? 
"Remove WebAPI logging from personal testing"

Thraxis:
Yes it should... the logging code was from my testing the sqlite3.row issue.  Was only for my own use to see if I could catch when the wrong row info was returned.  It's also not even the correct version of that logging as ep['show_name'] doesn't exist yet so that logging returns an error and doesn't actually log the info it should have.

PR #1582, opened 2015-03-18
Safety feature to convert sqlite3.row results to dict 

Thraxis:
Phantom changes to the DB connection results in the return of
sqlite3.row results when the connections was set to return dict.row
results before queries are called.  Possible cause is multi-threading
with a global DB connection.


fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/846

PR #1583, opened 2015-03-18
Add note to reminder user to add quality pattern in PP renaming

fernandog:
@TagForce until we change DB to fix this

UPDATE: the name patterns that SR have, only one have "quality" in it. Won't this cause issues if user select them?

### DIFF
--- a/gui/slick/interfaces/default/config_postProcessing.tmpl
+++ b/gui/slick/interfaces/default/config_postProcessing.tmpl
@@ -283,6 +283,10 @@
                                         <img src="$sbRoot/images/legend16.png" width="16" height="16" alt="[Toggle Key]" id="show_naming_key" title="Toggle Naming Legend" class="legend" class="legend" />
                                     </span>
                                 </label>
+                                <label class="nocheck">
+                                    <span class="component-title">&nbsp;</span>
+                                    <span class="component-desc"><b>NOTE:</b> Dont' forget to add quality pattern. Otherwise after post-procesing the episode will have UNKNOWN quality</span>
+                                 </label>
                             </div>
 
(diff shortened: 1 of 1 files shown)


TagForce:
@fernandog Yeah, well... Quality is a bit of a strange entity in SR at the moment. It gets the quality from the release_name, and it stores it in the db as a bitwise addition to the status int. It's a bit weird. I'll have to check to see if it uses the release_name afterwards, though.

fernandog:
@TagForce about the PR itself, do you agree?

TagForce:
@fernandog Not sure... It does not magically lose the quality info. It gets lost sometimes but I'm not convinced it's because there's no quality in the filename. It should use the release_name for that as that is the original name... It's 'under investigation' as I rewrite the quality tracking, along with implementing a db change and mediainfo scanning. So it'll be a while before it's done.

Ah, stuff it. Go ahead and merge this. It's just text. Won't hurt and is easily changed back later.

fernandog:
@TagForce @abeloin should we revert that?

TagForce:
No, not just yet.

abeloin:
@fernandog I would have said yes but ...
@TagForce Something special planned? New feature??

TagForce:
@abeloin Not really a feature... It still loses quality info sometimes. It can be reverted if you want, though. It shouldn't happen, but apparently does.

I'm working on re-thinking the whole quality setting in SR... Should make it more reliable, and also gives the user more control over what is wanted, and what is not. For instance, it seperates quality from status. It will also change the way it gets the quality (and the filename after post processing is not one of them), and will allow you to use mediainfo.

In short... when all of it is finished, there will be no need to use the quality info in the filename. But it will be some time before it is.

PR #1594, opened 2015-03-19
Fix default new special status

TagForce:
- When status for a new episode is unknown and it is a special, it
should now change its status to SKIPPED rather than default status set
in Show settings.

https://github.com/SiCKRAGETV/sickrage-issues/issues/21

### DIFF
--- a/sickbeard/tv.py
+++ b/sickbeard/tv.py
@@ -1770,7 +1770,10 @@ class TVEpisode(object):
 
                 # if we somehow are still UNKNOWN then just use the shows defined default status or SKIPPED
                 elif self.status == UNKNOWN:
-                    self.status = self.show.default_ep_status
+                    if self.season > 0: #If it's not a special
+                        self.status = self.show.default_ep_status
+                    else:
+                        self.status = SKIPPED
 
                 else:
                     logger.log(


TagForce:
Please test this, as this is seriously untested.

fernandog:
@TagForce shouldn't we set to IGNORED and add an option to UI? 

```
                 elif self.status == UNKNOWN:
                    if self.season = 0: #If it's not a special
                        if $sickbeard.IGNORE_SPECIALS:
                            self.status = IGNORED
                        else:
                            self.status = self.show.default_ep_status
                    else:
                        self.status = self.show.default_ep_status
```

fernandog:
@TagForce shoudn't we add this too?

https://github.com/TagForce/SickRage/blob/default_special_status_to_SKIPPED/sickbeard/tv.py#L1767

TagForce:
@fernandog Default special status is SKIPPED, not IGNORED. So it should set it to SKIPPED.

No, this should not be an option in the config, as this is default behaviour that is broken (it sets the specials to skipped when adding a show, regardless of what you set the default ep_status to, and also does the same when it finds a special that is unaired, so it should behave the same way for any new specials that are released with an unknown status).

https://github.com/TagForce/SickRage/blob/default_special_status_to_SKIPPED/sickbeard/tv.py#L1767
sets the status to skipped for specials that have a status of UNAIRED. I think the idea behind this was that new episodes released on the indexers would be gotten by SR before it airs... But sometimes they are added later, especially for specials since they are not always clearly added to indexers well before they air. It then fails, and sets the status to WANTED.

fernandog:
So if I want to not download specials what should I do?
and what to do If I want to download it?

TagForce:
@fernandog If you DON'T want to download specials... Merge this PR ;-)
If you DO want to download specials, go to your show and set them to WANTED. That is functionality that has always been the same.

fernandog:
@TagForce ok! will copy your reply to post in the issue =)

TagForce:
@fernandog More elaboration...
Most specials that are added are DVD featurettes. So they will always have an airdate of never. Those will now be set to SKIPPED, instead of WANTED (as they don't match any of the if statements and would be UNKNOWN).

PR #1595, opened 2015-03-19
Add Log directory to config page

fernandog:
@neoatomic can you test this in SYNOLOGY?

check if it will appear:
/share/MD0_DATA/.qpkg/SickBeard-TVRage/config/Logs

or something like that


neoatomic:
@fernandog 
 
naamloos

fernandog:
@neoatomic is that correct or not?

neoatomic:
@fernandog Yes, i store all logs on a usb stick so the disks can go into standby. Path is correct. :)

PR #1597, opened 2015-03-19
Add "Previous Ep" airdate in /Home

fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/218

Should we add a option to it?
If we do add an option, I don't know how to handle the sorting because the number of columns will change
Needs more testing sorting columns

Logic is:

```
SELECT airdate FROM tv_episodes WHERE showid=tv_eps.showid AND airdate <> 1 AND status <> ' + str($UNAIRED) + ' ORDER BY airdate DESC LIMIT 1) AS ep_airs_prev '
```

@TagForce please test it. Don't need to shutdown SR, Just replace the file and full refresh

### DIFF
--- a/gui/slick/interfaces/default/home.tmpl
+++ b/gui/slick/interfaces/default/home.tmpl
@@ -29,7 +29,9 @@
 #set $sql_statement += ' AND ((airdate <= ' + $today + ' AND (status = ' + str($SKIPPED) + ' OR status = ' + str($WANTED) + ' OR status = ' + str($FAILED) + ')) '
 #set $sql_statement += ' OR (status IN ' + status_quality + ') OR (status IN ' + status_download + '))) AS ep_total, '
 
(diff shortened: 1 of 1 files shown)


fernandog:
known issue: does not sort empty values. I think it's ok! What matter is to show dates in the firsts lines

fernandog:
@TagForce @abeloin here is how it looks like


home

abeloin:
@fernandog You need to edit the code for tablesorter

gui/slick/interfaces/default/home.tmpl
```diff
@@ -89,10 +91,11 @@
         sortList: [[5,1],[1,0]],
```
abeloin: This need to be updated

abeloin:
@fernandog See https://github.com/SiCKRAGETV/SickRage/pull/1600

PR #1592, opened 2015-03-19
Fix "mass edit" subtitle enable action

fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/836

@Hellowlol can you try? use this file please

PR #1598, opened 2015-03-19
Revert "Fix "mass edit" subtitle enable action"

fernandog:
Reverts SiCKRAGETV/SickRage#1592


abeloin:
@fernandog Don't forget to delete the branch after reverting.

PR #1599, opened 2015-03-19
Add workaround for improperly cleaned scgi over unix domain socket URI

anthonyryan1:
The newly added test should demonstrate the currently broken behavior.

Corresponding Sick-Beard pull request can be found here: https://github.com/midgetspy/Sick-Beard/pull/932

### DIFF
--- a/sickbeard/config.py
+++ b/sickbeard/config.py
@@ -28,6 +28,11 @@ from sickbeard import logger
 from sickbeard import naming
 from sickbeard import db
 
+# Address poor support for scgi over unix domain sockets
+# this is not nicely handled by python currently
+# http://bugs.python.org/issue23636
+urlparse.uses_netloc.append('scgi')
+
 naming_ep_type = ("%(seasonnumber)dx%(episodenumber)02d",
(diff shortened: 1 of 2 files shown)


fernandog:
@anthonyryan1 about tests, we are not running it.

anthonyryan1:
Even if the tests aren't being run consistently. I see no reason not to document behavior so we can try and avoid regressions. I ran only the config_tests.py group for verifying this change didn't break anything else in `clean_url`.

fernandog:
@abeloin

abeloin:
For scgi, we use urllib.splittype, so we aren't affected by this. But the (python docs)[https://docs.python.org/2/library/urllib.html] recommend to use urlparse.urlparse:
```
Note

urllib also exposes certain utility functions like splittype, splithost and others parsing url into various components. But it is recommended to use urlparse for parsing urls than using these functions directly. Python 3 does not expose these helper functions from urllib.parse module.
```

The benefit maybe none for now, but in the future if we correct the urllib.splittype usage it will be useful.

PR #1600, opened 2015-03-20
Added the ability to choose displaying columns in home.tmpl.

abeloin:
Added new column "previous episodes", disable by default.

### DIFF
--- a/gui/slick/css/dark.css
+++ b/gui/slick/css/dark.css
@@ -786,6 +786,26 @@ td.tvShow a:hover {
     color: #09A2FF;
 }
 
+#popover-target label {
+  margin: 0 5px;
+  display: block;
+}
+#popover-target input {
+  margin-left: 5px;
+}
+.popover {
+  margin-left: -50px;
+  background-color: #333;
+}
+
+.popover-content {
+  background-color: #333;
+}
+
+.popover.bottom .arrow:after {
+  border-bottom-color: #333;
+}
+
 /* =======================================================================
 home_addShows.tmpl
 ========================================================================== */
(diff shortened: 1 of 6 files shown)


fernandog:
@abeloin wow! thanks!

fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/218

TagForce:
@abeloin 
Now this is awesome!
Works flawlessly. Kudos!

fernandog:
@TagForce @abeloin I was thinking add this column widget to "fix" this:

https://sickrage.tv/forums/forum/main-category/feature-requests/866-filenames-on-show-pages-taking-up-a-lot-of-real-estate

so we can show/hide filesize/filename 

not now, in near future

TagForce:
@fernandog I was thinking the same thing... Just add all columns to the page, and use the column chooser to hide/display them. That way we can lose the "Show file size" option in settings, which people apparently cannot find (as evidenced by the forumpost you linked to just now, because I told them it was there and that is no longer an issue).

fernandog:
@abeloin 

JS / tablesorter does not work while adding shows:

abeloin:
@fernandog let me check

abeloin:
This is what I get:
Imgur
Imgur

abeloin:
@fernandog It's another bug, here without the PR:
! [Imgur](http://i.imgur.com/lJNZevr.png)

fernandog:
@abeloin yes. the problem is the "loading" in the first column. It's no a date. So JS fail.

PR #1606, opened 2015-03-20
Show log message when force propers when they are disabled

fernandog:


abeloin:
@fernandog This message seem to be not needed, as the queue is not started if false(see sickbeard/__init__.py#L1353)

fernandog:
@abeloin even force search? so why there's that check if is enabled?

abeloin:
@fernandog Yes because the force search, set the variable to force the search but since the scheduler is not active it never happend.

See https://github.com/SiCKRAGETV/SickRage/blob/develop/sickbeard/scheduler.py#L47

abeloin:
@fernandog What you could do is disable the force search proper in Manage Searches when it's not enable

PR #1610, opened 2015-03-20
home.tmpl: Fix loading show js error.

abeloin:
Also make sure they are always on top.

### DIFF
--- a/gui/slick/interfaces/default/home.tmpl
+++ b/gui/slick/interfaces/default/home.tmpl
@@ -465,8 +465,9 @@ $myShowList.sort(lambda x, y: cmp(x.name, y.name))
         </tr>
     </tfoot>
     
-	<tbody>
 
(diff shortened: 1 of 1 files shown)


fernandog:
@abeloin what about the first colum "(loading)" ?

abeloin:
@fernandog What about it?

fernandog:
Is a string. Not a date. wasnt that the problem?

abeloin:
No, the javascript error was on line 482. [Tablesorter extraction](https://github.com/abeloin/SickRage/blob/patch-home_loading_shows/gui/slick/interfaces/default/home.tmpl#L92) is looking for span title in that column which was missing.

Also depending on how and/or if you changed the sort order the line was sorted with everything else, so I've added a <tbody class="tablesorter-infoOnly"> which prevent sorting.

PR #1612, opened 2015-03-21
Remove hardlink and symlink from actions if Win32

fernandog:
@TagForce

### DIFF
--- a/gui/slick/interfaces/default/config_postProcessing.tmpl
+++ b/gui/slick/interfaces/default/config_postProcessing.tmpl
@@ -1,5 +1,6 @@
 #import os.path
 #import sickbeard
+#import sys
 #from sickbeard.common import *
 #from sickbeard import config
 #from sickbeard import metadata
@@ -71,7 +72,12 @@
                                 <span class="component-desc">
                                     <select name="process_method" id="process_method" class="form-control input-sm">
                                         #set $process_method_text = {'copy': "Copy", 'move': "Move", 'hardlink': "Hard Link", 'symlink' : "Symbolic Link"}
-                                        #for $curAction in ('copy', 'move', 'hardlink', 'symlink'):
+                                        #if sys.platform == 'win32'
+                                            #set $process_action = ('copy', 'move')
+                                        #else
+                                            #set $process_action = ('copy', 'move', 'hardlink', 'symlink')                                        
+                                        #end if
+                                        #for $curAction in $process_action:
                                           #if $sickbeard.PROCESS_METHOD == $curAction:
                                             #set $process_method = "selected=\"selected\""
                                           #else


gui/slick/interfaces/default/config_postProcessing.tmpl
```diff
@@ -71,7 +72,12 @@
                                 <span class="component-desc">
                                     <select name="process_method" id="process_method" class="form-control input-sm">
                                         #set $process_method_text = {'copy': "Copy", 'move': "Move", 'hardlink': "Hard Link", 'symlink' : "Symbolic Link"}
-                                        #for $curAction in ('copy', 'move', 'hardlink', 'symlink'):
+                                        #set $process_action = ('copy', 'move', 'hardlink', 'symlink')
+                                        #if sys.platform == 'win32'
```
abeloin: @fernandog Missing : at the end

gui/slick/interfaces/default/config_postProcessing.tmpl
```diff
@@ -71,7 +72,12 @@
...
+                                            #set $process_method_text = {'copy': "Copy", 'move': "Move"}
```
abeloin: @fernandog This is unneeded

gui/slick/interfaces/default/config_postProcessing.tmpl
```diff
@@ -71,7 +72,12 @@
                                 <span class="component-desc">
                                     <select name="process_method" id="process_method" class="form-control input-sm">
                                         #set $process_method_text = {'copy': "Copy", 'move': "Move", 'hardlink': "Hard Link", 'symlink' : "Symbolic Link"}
-                                        #for $curAction in ('copy', 'move', 'hardlink', 'symlink'):
+                                        #set $process_action = ('copy', 'move', 'hardlink', 'symlink')
```
abeloin: @fernandog You should do a if ... else instead of set and if.
```
#if sys.platform == 'win32':
    #set $process_method = ('copy', 'move')
#else
    #set $process_method = ('copy', 'move', 'hardlink', 'symlink')
#end if
#for $curAction in $process_method:

**Comment by fernandog** [comment]
@abeloin done!

File: gui/slick/interfaces/default/config_postProcessing.tmpl [hunk 4]
```diff
@@ -70,8 +71,14 @@
                                 <span class="component-title">Process Episode Method:</span>
                                 <span class="component-desc">
                                     <select name="process_method" id="process_method" class="form-control input-sm">
-                                        #set $process_method_text = {'copy': "Copy", 'move': "Move", 'hardlink': "Hard Link", 'symlink' : "Symbolic Link"}
```
abeloin: You don't need to remove this line as it is use to translate the action into a text label, so you don't need L75 and L78 which can be removed.

fernandog:
@abeloin but this way is working. I don't get it what is wrong

abeloin:
@fernandog Nothing wrong with the code you posted, but its easier to maintain by having only one list to query for the text label which is used on https://github.com/fernandog/SickRage/blob/pp_action_win/gui/slick/interfaces/default/config_postProcessing.tmpl#L87.

Eg: Hypothetically, let say we have a new action "move_WithoutPermission" then in your code I have to add it into 4 places:
* win32:
  * $process_action
  * $process_method_text
* not win32:
  * $process_action
  * $process_method_text

Using to other method you need 3 places:
* $process_method_text
* win32:
  * $process_action
* not win32:
  * $process_action

fernandog:
@abeloin done

rjrudman:
Out of curiosity, why was this removed? Hardlinks are extremely powerful with torrents and media centers. I added it back in locally, and it worked fine. What issues was it causing?

fernandog:
None

rjrudman:
@fernandog I mean, what was the reasoning of this pull request? What issues were hardlinks and symlinks causing on NT systems?

fernandog:
No issues

abeloin:
@rjrudman There was no issue other than peoples complaining because it didn't work the way they wanted, creating sym/hard link across partition result in a failure and SR ends up copying or/and move the date:
* symlink: move the file first then create a symlink(when fail try to copy it back)
* hardlink: link(when fail copy the files)

The code need to be changed for the windows part to detect those situation.

rjrudman:
@abeloin Thanks for the explanation; makes sense. Is there any open issue for this? I'd be happy to contribute

abeloin:
@rjrudman There was none in sickrage-issue so I created one: https://github.com/SiCKRAGETV/sickrage-issues/issues/1179

rjrudman:
@abeloin Cheers!

PR #1617, opened 2015-03-21
Ability to ignore embedded subtitles and layout fixes in config_subtitles.tmpl

Eiber:
Embedded subtitles are not always of good quality. This adds the ability to ignore embedded subtitles so new subtitles are downloaded. Also added NL in the addic7ed provider.

Note: it would be better to make this function a show setting, but this would need database changes..

### DIFF
--- a/sickbeard/webserve.py
+++ b/sickbeard/webserve.py
@@ -4794,7 +4794,7 @@ class ConfigSubtitles(Config):
 
     def saveSubtitles(self, use_subtitles=None, subtitles_plugins=None, subtitles_languages=None, subtitles_dir=None,
                       service_order=None, subtitles_history=None, subtitles_finder_frequency=None,
-                      subtitles_multi=None):
+                      subtitles_multi=None, embedded_subtitles_all=None):
 
         results = []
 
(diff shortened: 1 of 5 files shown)


fernandog:
@Eiber subliminal module don't download subs when there's embedded subtitles?

sickbeard/webserve.py
```diff
@@ -4778,14 +4778,14 @@ def index(self):
...
                       service_order=None, subtitles_history=None, subtitles_finder_frequency=None,
-                      subtitles_multi=None):
+                      subtitles_multi=None, embedded_subtitles_all=None):
 
         results = []
 
         if subtitles_finder_frequency == '' or subtitles_finder_frequency is None:
             subtitles_finder_frequency = 1
 
-        if use_subtitles == "on" and not sickbeard.subtitlesFinderScheduler.isAlive():
```
abeloin: @Eiber Do not change this value it should be "on"

Eiber:
@fernandog subliminal indeed doens not download new substitles when the preferred language is already embedded in the video file

@abeloin value changed back to "on". Not sure why this is necessary because the value of use_subtitles is never "on".

PR updated and rebased..

TagForce:
@Eiber use_subtitles is returned by the checkbox... When the box is checked, it's value is 'on':

from a debug:
subtitles_dir: str:
subtitles_finder_frequency: str: 1
subtitles_languages: str: nl
use_subtitles: str: on

Eiber:
Thanks for explaining!

PR #1619, opened 2015-03-22
GlypeProxy: Fix identation

abeloin:


fernandog:
@abeloin strange thing build has not yet finished after 10min

PR #1625, opened 2015-03-22
Catch read only database error and remove "raise"

fernandog:


abeloin:
@fernandog Do you understand the impact these changes do?

It'll be a cold day in Hell before this PR is merge into develop!!!

PR #1629, opened 2015-03-23
Added Free Space check in PP before deleting files

TagForce:
- Checks free diskspace on destination before deleting old episodes and
copying/moving the new one over.


TagForce:
@fernandog @abeloin Final version

PR #1510, opened 2015-03-09
Check available episode

gborri:
This function check for episode on torrent/nzb and set status on available if found.
Tested only with torrent.
For color code of new episode no preference, if you want to change it its ok.
- [x] if not curShow.paused chenge check
- [x] move log inside markAvailable procedure

PR #1632, opened 2015-03-23
Warn if backlog search days is less than oldest episode

fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/1018

It does not count paused shows with wanted episodes

### DIFF
--- a/gui/slick/interfaces/default/config_search.tmpl
+++ b/gui/slick/interfaces/default/config_search.tmpl
@@ -81,10 +81,26 @@
 							<label>
 								<span class="component-title">Backlog search day(s)</span>
 								<span class="component-desc">
+#from sickbeard import db
+#import datetime
+#set $myDB = $db.DBConnection()
+#set sql_statement = ' SELECT min(tv_episodes.airdate) AS airdate FROM tv_episodes'
+#set sql_statement += ' JOIN tv_shows ON tv_episodes.showid = tv_shows.indexer_id'
+#set sql_statement += ' WHERE tv_episodes.airdate <>1 AND tv_episodes.status = 3 '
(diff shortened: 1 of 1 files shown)


gborri:
Do you think is possible to set backlog search day to what needed in order to get all needed episode?

abeloin:
@gborri It's already been done via the dailySearcher. The back is now these day is mostly used for forcing a show backlog(set episode to wanted, add a new show with wanted as default)

TagForce:
@abeloin Maybe we should add a backlog search when changing providers in the search settings as well? As that's (as far as I can tell) the only scenario in which you might miss episodes that are old and wanted, at the moment. All the others are already backlogged before (indeed when adding a show, or manually changing to wanted).

abeloin:
@TagForce The function [searchForNeededEpisodes](https://github.com/SiCKRAGETV/SickRage/blob/develop/sickbeard/search.py#L369)(which is run by DailySearchQueueItem) already does that.

It call [wantedEpisodes](https://github.com/SiCKRAGETV/SickRage/blob/develop/sickbeard/search.py#L332) with a fromDate=1, which it then check for quality status and WANTED.

I'm currently writing a doc on this because it's not easy to find and understand.

In the end, from my understanding of the way dailySearcher now work(found that out yesterday's while discussing with @fernandog about his backlog PR and looking at the code), backlog should only be called manually.

fernandog:
@abeloin or we can "fix" the concept of dailysearch changing wantedEpisodes with a fromDate=1 to today()

gborri:
@abeloin @TagForce @fernandog take in mind also that with the #1510 backlog (and also daily) will search also for SKIPPED status.

PR #1635, opened 2015-03-24
Hounddawgs: Various fixes.

abeloin:
Added a check for "Dine cookies er ikke aktiveret." in login.
Added a session.get before the post

See https://github.com/SiCKRAGETV/sickrage-issues/issues/1019 for more info


abeloin:
@fernandog FYI

PR #1640, opened 2015-03-25
Fix active status on home accordingly to manage rolling download enable/disable

gborri:


gui/slick/interfaces/default/home.tmpl
```diff
@@ -680,7 +680,11 @@ $myShowList.sort(lambda x, y: cmp(x.name, y.name))
 		</td>
 	
         <td align="center">
+#if sickbeard.TRAKT_USE_ROLLING_DOWNLOAD
```
abeloin: Add a check if trakt is enable: `if TRAKT_USE_ROLLING_DOWNLOAD and USE_TRAKT:`

gborri:
@abeloin sorry, you are right. Done

abeloin:
@gborri Thanks

PR #1642, opened 2015-03-25
Only allow proxy for indexers if proxy host is enabled

fernandog:
@abeloin is there a better solution?

Also we don't check if proxy_host has http/https/port/etc... shouldn't we ?


fernandog:
TODO: fix using JS

PR #1643, opened 2015-03-25
Change removed episodes status from IGNORED to ARCHIVED

fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/1032

### DIFF
--- a/sickbeard/tv.py
+++ b/sickbeard/tv.py
@@ -1082,12 +1082,12 @@ class TVShow(object):
                 # check if downloaded files still exist, update our data if this has changed
                 if not sickbeard.SKIP_REMOVED_FILES:
                     with curEp.lock:
-                        # if it used to have a file associated with it and it doesn't anymore then set it to IGNORED
+                        # if it used to have a file associated with it and it doesn't anymore then set it to ARCHIVED
                         if curEp.location and curEp.status in Quality.DOWNLOADED:
                             logger.log(str(self.indexerid) + u": Location for " + str(season) + "x" + str(
-                                episode) + " doesn't exist, removing it and changing our status to IGNORED",
+                                episode) + " doesn't exist, removing it and changing our status to ARCHIVED",
                                        logger.DEBUG)
-                            curEp.status = IGNORED
+                            curEp.status = ARCHIVED
                             curEp.subtitles = list()
                             curEp.subtitles_searchcount = 0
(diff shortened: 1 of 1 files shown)


sickbeard/tv.py
```diff
@@ -1087,7 +1087,7 @@ def refreshDir(self):
                             logger.log(str(self.indexerid) + u": Location for " + str(season) + "x" + str(
                                 episode) + " doesn't exist, removing it and changing our status to IGNORED",
```
abeloin: @fernandog Should change the text here

fernandog:
@abeloin done

fernandog:
@TagForce can't we just remove line `curEp.status = ARCHIVED` so even if we remove the file, it keeps the original status? "Downloaded HDTV" instead or "ARCHIVED", Must we set a status?! I don't thin so...

TagForce:
Yes, we should have a status to show that the file is no longer there... If we keep it on downloaded we might try some other stuff on it (like rename) which will then fail because there's no such file.

fernandog:
@TagForce 

do we need status even when we set?
```
                        curEp.location = ''
                        curEp.release_name = ''
```

TagForce:
@fernandog Yep, even if it's just for management purposes...
Still, without location or releasename some things would go nuts if the status is still downloaded.
I'd just keep it on Archived, and in the future maybe even add an option in SR to really 'archive' to somewhere else.

fernandog:
@TagForce I'm saying this because I only have episodes on my HDD that I haven't watched. All others are deleted and I keep the "downloaded (quality)" status. 
I run an external script that cleans size and location. all because I want to keep the downloaded + quality information. I don't use PP

TagForce:
@fernandog You set Skip Removed Files? That keeps the last status an episode had, doesn't it?
So, there's already an option to do that... No need to change anything for those that don't skip removed files.

The change of status only occurs when:
```
if not sickbeard.SKIP_REMOVED_FILES:
```

Right?

gborri:
@abeloin @fernandog @TagForce i will make a PR in order to decide which status will be set for media file that are deleted (default choice not hard coded). I like the ignore status because on home page the number of episode is related only on what i have as media or not yet downloaded.
The archived status are always been count.

gborri:
@abeloin @fernandog @TagForce done. SiCKRAGETV/SickRage#1650

PR #1644, opened 2015-03-25
Add Dutch language support for Addic7ed

whizzoo:
This will allow SickRage to search Addic7ed for Dutch subtitles. Tested and works great.


abeloin:
@whizzoo Already in develop. Thanks for the PR

PR #1650, opened 2015-03-26
Make removed episodes status configurable

gborri:
- [x] enabled SKIP_REMOVED check


fernandog:
Show this option only if user didnt enabled SKIP_REMOVED

gborri:
Ok.

fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/1032

gborri:
@fernandog the check should be done against SKIP_REMOVED and also against TRAKT_USE_ROLLING_DOWNLOAD. 
EP_DEFAULT_DELETED_STATUS is also used for changing status of watched episode.

fernandog:
@gborri no because the original code dont check that. Of check skip first and then rolling. Dont check both in the same if

gborri:
@fernandog why not checking on the same if?

fernandog:
@gborri I have skip_removed ON and USE_ROLLING_DOWNLOAD OFF

with a single if , the option will appear to me, but it shouldn't

gborri:
you need to use the "or": skip_removed ON or USE_ROLLING_DOWNLOAD 
this way it works.

abeloin:
@gborri squash your commits into one with a rebase, and I'll mere the PR

gborri:
@abeloin I'm feeling crazy..... everytime i rebase or merge i get inside the LFCTRL problem.

abeloin:
@gborri New PR should be ok. PR merged

PR #1652, opened 2015-03-26
Add show: fix improperly decoded utf-8 characters

abeloin:
See https://github.com/SiCKRAGETV/sickrage-issues/issues/937 for more info


fernandog:
works in linux

PR #1655, opened 2015-03-26
Remove jwplayer feature

fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/1042

@abeloin is it ok?


abeloin:
@fernandog Everything looks ok

PR #1657, opened 2015-03-27
Fix: cannot find 'PLAY_VIDEOS'

abeloin:
See https://github.com/SiCKRAGETV/sickrage-issues/issues/1046 for more info


fernandog:
Thanks! Sorry about that

PR #1661, opened 2015-03-27
Prevent seeing debug and db log in viewlog when debug=0 in config.ini

abeloin:


abeloin:
@fernandog Is this ok with you?

fernandog:
this will prevent users to submit errors if debug=0 
or will submit log with debug missing

abeloin:
Nope, it doesn't display debug/db line in vewlog page if debug=0:
Imgur

fernandog:
@abeion! ok then!
do we have an option in UI to enabled debug or only editing config.ini?

abeloin:
config.ini only

fernandog:
ok! later we add it to the UI. not urgent

maxcore:
@abeloin @fernandog  where we can actived the debug log ? in the ui

PR #1663, opened 2015-03-28
PostProcess: Fix lower quality, bigger sized file overwriting current episode.

abeloin:
See https://github.com/SiCKRAGETV/sickrage-issues/issues/1023 for more info


fernandog:
Nice one!

abeloin:
@TagForce FYI

fernandog:
@abeloin 

what about this part of the code. Is it still checking size instead of quality:
https://github.com/SiCKRAGETV/SickRage/blob/develop/sickbeard/postProcessor.py#L137

abeloin:
@fernandog Look at the definition for the function:
Checks if a file exists already and if it does whether it's bigger or smaller than the file we are post processing

PR #1670, opened 2015-03-29
Trakt API 2.0:  Authentication - OAuth

Thraxis:
Implements the Trakt API 2.0 Authentication - OAuth process

To use you need to go to https://trakt.tv/oauth/applications and create an application to get your Client ID and Client Secret Keys  you will need to use "urn:ietf:wg:oauth:2.0:oob" as the Redirect URI.

Once you have the application created when you run SickRage and goto notifications for the trakt page you put in the Client ID, Client Secret and Username for your account.  Click the [Get Authorization] button to authorize SickRage.  Copy the key from the window that opens then click the button again (should now say [Input Authorization] ) paste the authorization key into the dialog box that pops open and click OK.

Once Authorized your sickrage will get two tokens that are stored in the config.  One is an Access Token, the Other is a refresh Token.  Periodically the system will check to see if your access token is close to expiring.  if it is it will automatically request a new access token.   By default it takes 3 months for an access token to expire.

All trakt functions should be working with this code.  Watchlist Sync, Trending Shows, etc.


fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/1059

Thraxis:
Just a note that this could also be changed to use a hard coded client id / client secret vs. having each sickrage user create their own.  They would still need to Authenticate the application access.

Which would bring up the issue of using "Authentication - Media Centers" as the method,  which the current code used by SickRage supports. But the Client ID has to be approved as a Media Center Application by Trakt to use the trakt-user-login, trakt-user-token so you get the 401 Client not authorized errors

Thraxis:
Looks like this code may not be needed.  Retesting today with current develop branch and I'm not seeing any problems with the API working other then timeout on the https://api.trakt.tv URL changing to https://api-v2launch.trakt.tv looks to work as well.  

@abeloin @fernandog  Flag this as a don't commit for now

Thraxis:
I'm closing this PR as it doesn't need to be used. I'll keep the branch on my Repo if it ever does become required the code will be available for reference.

PR #1675, opened 2015-03-30
Some tweaks for Trakt

Thraxis:
https://github.com/SiCKRAGETV/sickrage-issues/issues/1059

Changes the API URL to the V2 API URL as it has more resources then the V1 API URL
Fixes an error that can occur in Trackt Checker for shows that do not have a tvrage id on Trakt


Thraxis:
Wrong PR compare

PR #1681, opened 2015-03-30
KAT : enclosure season search with " when searching for season number

fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/1055


abeloin:
@fernandog Why not use the season and episode tag instead?
`https://kickass.to/usearch/The%20Vampire%20Diaries%20category%3Atv%20season%3A1/`

fernandog:
@abeloin user reported that the fix worked. the season tag it's not good for searching season packs

PR #1686, opened 2015-03-30
Add RSS to EZTV provider and save magnet link as fallback

nicolasmartinelli:
EZTV:
- use of https://eztv.ch as URL for RSS feed, since http://eztvapi.re does not provide RSS.

Generic:
- remove torrage, does not exists anymore
- fallback: save magnet link if torrent could not be downloaded. Some softwares can use it,
  in the worst case the user will have to add it manually.

The fallback is not a perfect solution since not all softwares will handle a ".magnet" file. However, the user does not face a dead end, and still has a possibility to get the torrent. This could be useful if, in the future, services such as zoink or torcache are closed down.

### DIFF
--- a/sickbeard/providers/eztv.py
+++ b/sickbeard/providers/eztv.py
@@ -16,13 +16,13 @@
 #
 # You should have received a copy of the GNU General Public License
 # along with Sick Beard.  If not, see <http://www.gnu.org/licenses/>.
-
 import traceback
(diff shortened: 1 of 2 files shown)


abeloin:
@nicolasmartinelli Please do not put multiple fix/feature in one PR, create a PR for each one of them.

PR #1688, opened 2015-03-31
Patch tntvillage season pack manage

gborri:
This PR will get season pack as in TNTVillage (XXX S01E01-24) and change it accordingly to what sickrage expect (xxx S01).


abeloin:
@gborri Thanks

PR #1689, opened 2015-04-01
Add RSS to EZTV provider

nicolasmartinelli:
Use of https://eztv.ch as URL for RSS feed, since http://eztvapi.re does not provide RSS.


abeloin:
@nicolasmartinelli Thanks

PR #1690, opened 2015-04-01
Save magnet link as fallback

nicolasmartinelli:
- Save magnet link if torrent could not be downloaded. Some softwares can use it,
in the worst case the user will have to add it manually.
- Remove torrage, does not exists anymore

The fallback is not a perfect solution since not all softwares will handle a ".magnet" file. However, the user does not face a dead end, and still has a possibility to get the torrent. This could be useful if, in the future, services such as zoink or torcache are closed down.

- [ ] Send notification - if user has any enabled - about the location of the magnet link direcory

### DIFF
--- a/sickbeard/providers/generic.py
+++ b/sickbeard/providers/generic.py
@@ -185,6 +185,21 @@ class GenericProvider:
                 if self._verify_download(filename):
                     return True
 
+        # Fallback: if the torrent link could not be downloaded, we save the magnet link instead
+        # Softwares such as Deluge with plugin AutoAdd can automatically add it, otherwise the user
+        # can add it manually.
+        if self.providerType == GenericProvider.TORRENT:
+            filename_fallback = re.sub(r'torrent$', r'magnet', filename)
+            if filename_fallback != filename and re.match('^magnet:', result.url):
+                try:
+                    f = open(filename_fallback, 'w')
+                    f.write(result.url)
+                    f.close()
+                    logger.log(u"Saved magnet link to " + filename_fallback, logger.INFO)
+                    return True
+                except:
+                    pass
+
         logger.log(u"Failed to download result", logger.WARNING)
         return False
 


abeloin:
Pushed back to next week

nicolasmartinelli:
ok, thanks for letting me know

fernandog:
@nicolasmartinelli no errors/issues with this PR? working ok?

nicolasmartinelli:
Still ok as of today.

fernandog:
@nicolasmartinelli 

"otherwise the user can add it manually." - how would that be?

nicolasmartinelli:
@fernandog 

The user gets a text file containing the magnet link. He can do simple copy-paste of the link in his torrent client.

Once again, this is not a perfect solution, but I came across the situation where the torrent is neither available in torcache nor zoink a couple of times. Hence, I was facing a 'dead end' with Sickrage: Sickrage was not able to get the torrent while a magnet link exists. So I had to go on the website of the provider, search for the torrent, and click on the magnet link to download it. At least with this solution, all the magnet links are gathered in the 'black hole', and I simply have to do a copy paste of every link. It is not great, but still more convenient than having to search for every single file manually.

fernandog:
@nicolasmartinelli but AFAIK torrents clients supports magnet links, doesn't they? Can't we send the ML to them ?

nicolasmartinelli:
@fernandog yes, torrents clients support magnet links. And so far, they are sent to the clients, as long as you have configured one of them (deluge, µtorrent, ruTorrent...). My patch only applies to the 'black hole' configuration, where the torrent (or the magnet link, in this case) is saved in a folder on the drive.

fernandog:
@nicolasmartinelli I see. thanks!

fernandog:
@nicolasmartinelli what do you think sending a notification (if use has any enabled) that says the magnet link was saved in ....  and also add a note to UI if user has blackhole set?

nicolasmartinelli:
@fernandog : That would be better, indeed.

For a note in the UI, I guess something like this should do the trick, right?
`ui.notifications.message('Torrent could not be downloaded, so magnet link was saved instead', result.name)`

According to the notification, should I create a new method in emailnotify.py?

fernandog:
@nicolasmartinelli that will only will work for user who user manual search. Otherwise dailysearch/backlog are automatic and most users don't open SR for days/weeks

nicolasmartinelli:
@fernandog : yes I understand. I will have a look to send a notification as well.

fernandog:
@nicolasmartinelli any news on the notification? working as expected?

nicolasmartinelli:
@fernandog : I am a bit busy at the moment with other projects, unfortunately I couldn't work on it. I think it could be set "on hold", and I will notify when I have something done.

Thanks!

miigotu:
@nicolasmartinelli would this become unnecessary if the hash were in the db, and could be seen by maybe a column in history page or a js popup when clicking the action in the history page?

PR #1693, opened 2015-04-01
Check is showupdate is active and alive before changing hour

fernandog:


abeloin:
See https://github.com/SiCKRAGETV/SickRage/pull/1695

PR #1694, opened 2015-04-01
Fix default delete episode in GUI

abeloin:
See https://github.com/SiCKRAGETV/sickrage-issues/issues/1085 for more info


gborri:
@abeloin thanks.

PR #1699, opened 2015-04-02
Fix unicode error in verify_freespace

TagForce:
- Fixes unicode error by using EncodingKludge for isfile check instead
of assuming it is unicode.


abeloin:
Ref https://github.com/SiCKRAGETV/sickrage-issues/issues/1088

PR #1705, opened 2015-04-03
use gir-notify instead of pynotify

3flex:
libnotify support was broken by ef4470b which introduced gi.repository for the first time.
This is incompatible with pynotify. Switching to gir-notify resolves the issue.

Also fixes the file path for the notification icon.

Fixes SiCKRAGETV/sickrage-issues#733

Tested, works on Ubuntu 15.04 which uses gir1.2-notify-0.7 (which is same package used in all Debian/Ubuntu versions except Debian Squeeze and Ubuntu Precise).

### DIFF
--- a/sickbeard/notifiers/libnotify.py
+++ b/sickbeard/notifiers/libnotify.py
@@ -29,10 +29,11 @@ def diagnose():
     user-readable message indicating possible issues.
     '''
     try:
-        import pynotify  #@UnusedImport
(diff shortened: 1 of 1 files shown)


abeloin:
I fail to see how send2trash have broken pnotify?

Ref: https://github.com/SiCKRAGETV/SickRage/commit/ef4470bd789da13878d1a8db328b08374f501f63

abeloin:
Pushed back for next week

PR #1711, opened 2015-04-04
Fix freuqency in start queue

gborri:


abeloin:
@gborri Merged

abeloin:
@gborri Does it really need to be run every 15 minutes? I would think running it 2 or 3 times a days should be enough.

fernandog:
@gborri low powered devices may have problems with this high frequency

gborri:
@abeloin @fernandog if you let me know how to set default value I will change it.

PR #1718, opened 2015-04-06
Add daily search footer highlight updater

DungFu:
Added javascript code to update the daily search timeLeft footer highlight every second when the tab is active. The Date.now() logic is required so that the time properly updates when the tab is inactive. The timeLeft will also reset using the cycle time when the timeLeft reaches or goes below 0.

### DIFF
--- a/gui/slick/interfaces/default/inc_bottom.tmpl
+++ b/gui/slick/interfaces/default/inc_bottom.tmpl
@@ -77,9 +77,18 @@
         // with the cycle time.
 
         function getTimeNumber(timeArr) {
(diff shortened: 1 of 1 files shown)


DungFu:
updated to handle days correctly.

fernandog:
@DungFu please squash the commits into one

abeloin:
@fernandog Don't think this should be merged as when the status page will auto refresh it will have to query the server for that kind of info which then this code could use it to display the information.

fernandog:
@abeloin ok!

@DungFu we now have a status page with all threads (develop). We won't merge this. Feel free to do other PRs. thanks!

PR #1724, opened 2015-04-08
Adds option to add a filter row on main show page:

Thraxis:
Implements Feature Request:
https://github.com/SiCKRAGETV/sickrage-issues/issues/1112
Added as an option on the General Config -> Interface Page.  enabling it
adds a filter row that hides when not in use.


sickbeard/__init__.py
```diff
@@ -1126,6 +1127,7 @@ def path_leaf(path):
         TIMEZONE_DISPLAY = check_setting_str(CFG, 'GUI', 'timezone_display', 'network')
         POSTER_SORTBY = check_setting_str(CFG, 'GUI', 'poster_sortby', 'name')
         POSTER_SORTDIR = check_setting_int(CFG, 'GUI', 'poster_sortdir', 1)
+        FILTER_ROW =  bool(check_setting_int(CFG, 'GUI', 'display_filesize', 0))
```
abeloin: @Thraxis Shouldn't that be filter_row instead of display_filesize?

Thraxis: Yes it should... It's what I get for coding late at night and using copy
paste

fernandog:
@Thraxis please squash commits into one. thanks!

Thraxis:
@fernandog One single Commit now

abeloin:
@Thraxis PR merged

fernandog:
@Thraxis how do I filter the column "Active" ?
We should add a note in the setting about use "|" for multiple filters

that kind of filter I posted in the issue, is it possible? It list of values from that column

Thraxis:
Yes we can add in the options to do a drop down for select columns, or exclude them as well as add a "Try (xxx|ss*) to show up in the filter boxes for filter examples.

fernandog:
@Thraxis how to filter the active? As it's an image I couldn't filter it. Like show only Active

Thraxis:
Have to make a modification to allow it to be filtered.  it needs to have a title="Active" tag ... working up a pt2 to add this.

abeloin:
@Thraxis No need I already made a PR

PR #1726, opened 2015-04-08
Check response before return

fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/1134


fernandog:
@abeloin can't understand why the problem with the PR

PR #1729, opened 2015-04-08
Show wanted in displayShow

abeloin:
Revert commit https://github.com/SiCKRAGETV/SickRage/commit/728b79f5db6fd6ead56b874c2a030dafc794e320 which cause issue https://github.com/SiCKRAGETV/sickrage-issues/issues/1138#issuecomment-90972225

### DIFF
--- a/sickbeard/tv.py
+++ b/sickbeard/tv.py
@@ -1265,7 +1265,7 @@ class TVShow(object):
 
     def getOverview(self, epStatus):
 
-        if epStatus == WANTED and not self.paused:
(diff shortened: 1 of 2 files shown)


fernandog:
@abeloin why are you showing PAUSED in backlog overview? SR isnt looking/searching them

abeloin:
@fernandog Don't forget getOverview is used by displayShow and backlogOverview, so I move the check for paused show from getOverview to backlogOverview. In backlogOverview, it return only show with paused = 0 skipping the display of paused show

PR #1732, opened 2015-04-10
home filter: change to allow to use parsed data for active(yes/no)

abeloin:
See  https://github.com/SiCKRAGETV/SickRage/pull/1724

### DIFF
--- a/gui/slick/interfaces/default/home.tmpl
+++ b/gui/slick/interfaces/default/home.tmpl
@@ -102,7 +102,10 @@
 			1: { columnSelector: false },
 			2: { sorter: 'loadingNames' },
 			4: { sorter: 'quality' },
-			5: { sorter: 'eps' }
+			5: { sorter: 'eps' },
+			#if $sickbeard.FILTER_ROW:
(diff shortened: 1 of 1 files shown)


abeloin:
@fernandog

Thraxis:
May want to add Network in there as well, it's not working for filtering either

abeloin:
@Thraxis Network works fine here. The one not working is Downloads, but I don't know how we should handle filtering for that column

Thraxis:
@abeloin Odd.. wasn't working when I tested a minute ago.

The Downloads progress bar filtering is actually done by percentage.  ">20"  "<50" "100" all should work for filtering it

abeloin:
@Thraxis Thanks, I forgot about < > sign, no wonder it wasn't working.

Thraxis:
Ahh... clarification on the Network Filtering...  when you are in Simple Mode the filtering works, When you are in banner mode it does not.

fernandog:
@abeloin thanks!

Downloads: 100 does not work. But >99 do.
Am I doing something wrong?

PR #1745, opened 2015-04-12
Fix for xml_declaration unexpected keyword in python 2.6.

miigotu:
SiCKRAGETV/sickrage-issues#1166
Must be 'UTF-8' instead of 'utf-8' to get automatically written as xml and portable across python versions, as xml_declaration does not exist in 2.6.
Should work regardless of version now. Needs Tested.

Note:If this does not work as expected, method="xml" instead of xml_declaration=True should be tried.

### DIFF
--- a/sickbeard/metadata/mede8er.py
+++ b/sickbeard/metadata/mede8er.py
@@ -390,7 +390,7 @@ class Mede8erMetadata(mediabrowser.MediaBrowserMetadata):
 
             nfo_file = ek.ek(open, nfo_file_path, 'w')
 
-            data.write(nfo_file, encoding="utf-8", xml_declaration=True)
+            data.write(nfo_file, encoding="UTF-8")
             nfo_file.close()
             helpers.chmodAsParent(nfo_file_path)
         except IOError, e:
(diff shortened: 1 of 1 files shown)


shadowjustice:
@miigotu  I inserted the fix you proposed. With the fix he proposed metadata xml now looks like this:

http://pastebin.com/2FvqD3s8

I am however getting next error:

http://pastebin.com/RDs5BN52

I think however it is totally unrelated to this issue. And this is very strange since notifications mail is not checked.

image


I tested with:

SR Version:	BRANCH: (develop) / COMMIT: (e917dc613ac17943dbe795331325c228c3d77f05) 
Python Version:	2.6.6 (r266:84292, Dec 27 2010, 21:57:32) [GCC 4.4.5 20100902 (prerelease)]

miigotu:
Pretty sure that second error is unrelated. Did you also make the change at line 438 in mede8er.py?

shadowjustice:
@miigotu I changed both lines :)

miigotu:
OK. But the prowl notification works now correct?

shadowjustice:
Yes for me it is OK.

Verstuurd vanaf mijn iPhone

PR #1750, opened 2015-04-12
Build tests fix, replaces #1525

miigotu:
Fixed travis build tests, all are working and being ran. None of the tests actually got done before.  Fix for #1525 without breaking the build.

Is it ok that I set journal_mode = OFF? Build does not work without this setting.

### DIFF
--- a/sickbeard/db.py
+++ b/sickbeard/db.py
@@ -59,7 +59,8 @@ class DBConnection(object):
                 self.connection = sqlite3.connect(dbFilename(self.filename, self.suffix), 20, check_same_thread=False)
                 self.connection.text_factory = self._unicode_text_factory
                 self.connection.isolation_level = None
-
+                self.connection.cursor().execute('''PRAGMA journal_mode = OFF''')
+                self.connection.commit()
                 db_cons[self.filename] = self.connection
             else:
                 self.connection = db_cons[self.filename]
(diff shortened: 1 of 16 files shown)


fernandog:
@miigotu Is issue submitter test runned ok?
this one shouldnt run everytime otherwise will create a lot of issues.

miigotu:
I'll either add the blacklisting from #1525 to avoid the issue_submitter_test, or make a change in the issue_submitter_test code now and push

miigotu:
Ok, blacklisted, but the test does run. I dont have my github info in the ini (and travis doesnt either) so it doesnt actually post the issue though even when run manually :)

fernandog:
@miigotu thanks and welcome! I sent you a invite to be moderator too

miigotu:
Yeah, thanks. Already accepted/joined.

abeloin:
@miigotu What is the error when PRAGMA is on?

miigotu:
@abeloin If you mean without disabling journal_mode the errors are here: https://gist.github.com/miigotu/391b5577a822e5f6508e

The databases are not created, but the journals are created but are 0b and are not removed after the test. Strange bug but it is safer this way anyways. I even set the parent folder and all files 777 as well to make sure it wasnt a permissions problem.

abeloin:
@miigotu Ok, I put this for v4.0.18. This leave us time to figure the issue

PR #1753, opened 2015-04-12
ShowUpdater: referenced before assignment curQueueItem

MaximusCub:
A small change to fix https://github.com/SiCKRAGETV/sickrage-issues/issues/1193 : 

UnboundLocalError exception can be thrown at piList.append(curQueueItem) in cases where the CantUpdateException is thrown in the earlier try block around line 81. In this case, curQueueItem is never assigned a value and the list appending fails.

This change just skips the need for a temp holding variable "curQueueItem" and appends directly to the list the return from the updateShow/refreshShow methods.

Now if updateShow does throw a CantUpdateException, the piList.append in the try block is never attempted and so the UnboundLocalError is avoided.

Note: This is my fist github contribution, so apologies if I mess anything up.

### DIFF
--- a/sickbeard/showUpdater.py
+++ b/sickbeard/showUpdater.py
@@ -78,16 +78,14 @@ class ShowUpdater():
                 # if should_update returns True (not 'Ended') or show is selected stale 'Ended' then update, otherwise just refresh
                 if curShow.should_update(update_date=update_date) or curShow.indexerid in stale_should_update:
                     try:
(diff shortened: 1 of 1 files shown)


miigotu:
Looks proper, however in the future your patch branch name could be more descriptive of your change, for both your benefit and the committers here.

fernandog:
@MaximusCub no issues with this PR? safe to merge?

PR #1757, opened 2015-04-13
Auto determine indexer when indexer tag not present in nfo

miigotu:
Fixes adding existing shows automation.

Removed "episodeguideurl" tag from being created in the nfo as it is not valid and kodi doesn't even read it.

TODO: Find why indexer and id are set twice each in nfo files created by SR

Fixes https://github.com/SiCKRAGETV/sickrage-issues/issues/1205

### DIFF
--- a/sickbeard/metadata/generic.py
+++ b/sickbeard/metadata/generic.py
@@ -951,11 +951,6 @@ class GenericMetadata():
 
             name = showXML.findtext('title')
 
-            try:
-                indexer = int(showXML.findtext('indexer'))
-            except:
-                indexer = None
-
             if showXML.findtext('tvdbid') != None:
                 indexer_id = int(showXML.findtext('tvdbid'))
             elif showXML.findtext('id') != None:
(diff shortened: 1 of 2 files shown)


abeloin:
> TODO: Find why indexer and id are set twice each in nfo files created by SR

@miigotu For clarification, the doubling of tag was there before this patch?

miigotu:
@abeloin yes, the double `<id>` and `<indexer>` tags are present in develop before this patch.
I made the TODO note because I just noticed the issue while I was working on this patch. I had to let SR create an NFO for TVRage so I could test that my patch worked for TVRage, since all of my NFO are using theTVDB.

The NFO SR created before my patch: https://gist.github.com/miigotu/f011870739cb438c7a7b

With my patch I also removed creation of the `<episodeguideurl>` tag as it is invalid.

miigotu:
@abeloin this is safe to merge, the double/triple id and indexer bug precedes this patch, and I dont change how the nfo is written at all besides taking out the invalid `<episodeguideurl>`.
I have tested this extensively for both tvrage and thetvdb, and my change only effects when mass adding existing shows with existing nfo.

abeloin:
@miigotu Ok, thanks

fernandog:
@miigotu still no issues with this PR? safe to merge?

fernandog:
@TagForce any comment?

miigotu:
@fernandog Very safe. Best to merge this now and then phase out `<indexer>` within next 2 milestones.

PR #1758, opened 2015-04-13
iptorrent: Add missing import

abeloin:
See https://github.com/SiCKRAGETV/sickrage-issues/issues/1201 for more info


gborri:
@abeloin probably USER_AGENT is not needed

abeloin:
@gborri You're right I'll modify the PR. Thanks

PR #1762, opened 2015-04-13
Ignore OSError when obtaining the size of a path.

ElmerLastdrager:
Fix bug where a broken symlink will break viewing a show.
Reference: SiCKRAGETV/sickrage-issues#1224

### DIFF
--- a/sickbeard/helpers.py
+++ b/sickbeard/helpers.py
@@ -1411,9 +1411,9 @@ def get_size(start_path='.'):
             fp = ek.ek(os.path.join, dirpath, f)
             try:
                 total_size += ek.ek(os.path.getsize, fp)
-            except OSError, e:
-                logger.log(u"Unable to get size for file " + fp + ": " + repr(e) + " / " + str(e),
-                                           logger.DEBUG)
+            except OSError as e:
+                logger.log('Unable to get size for file {filePath}. Error msg is: {errorMsg}'.format(filePath=fp, errorMsg=str(e)), logger.ERROR)
+                logger.log(traceback.format_exc(), logger.DEBUG)
     return total_size
 
 def generateApiKey():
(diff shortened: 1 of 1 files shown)


sickbeard/helpers.py
```diff
@@ -1409,7 +1409,11 @@ def get_size(start_path='.'):
     for dirpath, dirnames, filenames in ek.ek(os.walk, start_path):
         for f in filenames:
             fp = ek.ek(os.path.join, dirpath, f)
-            total_size += ek.ek(os.path.getsize, fp)
+            try:
+                total_size += ek.ek(os.path.getsize, fp)
+            except OSError, e:
```
abeloin: @ElmerLastdrager Try to use `except OSError as e:` instead, we are trying to write the new code to be as much possible compatible with python3 to make it easier to port one day.

sickbeard/helpers.py
```diff
@@ -1409,7 +1409,11 @@ def get_size(start_path='.'):
...
+                logger.log(u"Unable to get size for file " + fp + ": " + repr(e) + " / " + str(e),
```
abeloin: Here I would recommend using 2 log line:
1. Error log line to inform the user of the error:
  `logger.log('Unable to get size for file {filePath}. Error msg is: {errorMsg}'.format(filePath=fp, errorMsg=str(e)), logger.ERROR)`
2. A debug line for the traceback(for us to be able to debug where is was thrown):
   `logger.log(traceback.format_exc(), logger.DEBUG)`

Don't forget to import traceback if it isn't

ElmerLastdrager: @abeloin thanks for your comments. I have updated the patch accordingly.

abeloin:
@ElmerLastdrager Thank you

PR #1763, opened 2015-04-13
SceneTime.com torrent provider

bjornhamra:
Basically a plain copy of the torrentbytes provider, with a few minor compat tweaks.


fernandog:
@bjornhamra working without any issues?

bjornhamra:
@fernandog Has been for me for quite a while, although I don't think anyone else has tried it yet. I could provide an invite if you would like to test it yourself. 

As previously noted, the commit is pretty much a plain copy of the torrentbytes provider (Credits go to whoever wrote that). A quick `diff -u sickbeard/providers/{torrentbytes,scenetime}.py` will shows that the only actual changes besides naming/urls is the markup/bs4 parsing, where I had to make a few minor changes because ST's markup is horribly messy.

PR #1766, opened 2015-04-14
Fix Episode Thumbnail url when show is in DVD Order

TagForce:
- Uses DVD order on TVDB and TVRage queries to get the correct episode

https://github.com/SiCKRAGETV/sickrage-issues/issues/1132


fernandog:
@TagForce thanks!

PR #1771, opened 2015-04-14
HTML changes of https://eztv.ch

nicolasmartinelli:
Slight changes in the HTML occurred on 2015-04-14


nicolasmartinelli:
According to my log files, it was working on April 13th but not anymore on April 14th...

PR #1773, opened 2015-04-15
Fix RegEx Issues

Thraxis:
* Fixed Regex patterns to use both lower and upper case character
matching in cases where it was searching only lower case.


abeloin:
@Thraxis Regex are searched with

Thraxis:
Didn't notice that as all the original Regex were built for case.

Doesn't explain why things aren't matching in the parser then as I know I've seen it fail show names that I know should have passed

abeloin:
It depend if it's searching for season pack, if the show exists, ...

In the end, without the log, it almmost impossible to know

PR #1774, opened 2015-04-15
Fix Case Sensitive associated files check

TagForce:
Changes the Associated files check when running during PP (subfolders =
False) to do the following:
- Lists all files in current folder
- Loops through each file and checks if they are the same as the
videofile in a case-insensitive way

This is needed because both glob.glob and fnmatch act in a
case-sensitive matter on case-sensitive file systems.

Should fix https://github.com/SiCKRAGETV/sickrage-issues/issues/1117 for
the most part. Differently named files (not only case differences) will
still be missed

### DIFF
--- a/sickbeard/postProcessor.py
+++ b/sickbeard/postProcessor.py
@@ -183,10 +183,19 @@ class PostProcessor(object):
         # don't confuse glob with chars we didn't mean to use
         base_name = re.sub(r'[\[\]\*\?]', r'[\g<0>]', base_name)
         
-        if subfolders:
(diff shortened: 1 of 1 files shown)


fernandog:
@TagForce thanks!

fernandog:
@TagForce what if we add a log message telling user about the missed ones? a simple explanation. what do you think?

TagForce:
@fernandog There's no need for that, as the files that are missed are also in the 'unwanted files' list that is already in the log.

PR #1776, opened 2015-04-15
Check if rtorrent timed out and show a warning

fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/1105


fernandog:
@abeloin shouldn't this timeout be handled here?

https://github.com/SiCKRAGETV/SickRage/pull/1760/files

why generic is not catching the timeout error?

fernandog:
@abeloin what about this one?

PR #1780, opened 2015-04-16
FindSearchResult speed up

gborri:
with this PR if all episode are on cache the search on providers will not be performed.

### DIFF
--- a/sickbeard/providers/generic.py
+++ b/sickbeard/providers/generic.py
@@ -276,25 +276,27 @@ class GenericProvider:
                 # found result, search next episode
                 continue
 
-            # skip if season already searched
-            if len(episodes) > 1 and searched_scene_season == epObj.scene_season:
-                continue
-
-            # mark season searched for season pack searches so we can skip later on
-            searched_scene_season = epObj.scene_season
-
-            if len(episodes) > 1:
-                # get season search results
-                for curString in self._get_season_search_strings(epObj):
-                    itemList += self._doSearch(curString, search_mode, len(episodes), epObj=epObj)
-            else:
-                # get single episode search results
-                for curString in self._get_episode_search_strings(epObj):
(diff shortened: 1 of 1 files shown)


fernandog:
@gborri all episodes and all qualities or only Season+Episode?

gborri:
@fernandog I havent undestood your question. Before this PR the code will check for episode on cache all episode has beed found on cache it will search for it on provider then exit. With this PR if all episode is found on cache it will not search on providers

fernandog:
@gborri what if in the provider there's a PROPER?

fernandog:
@gborri or a higher quality one?

gborri:
@fernandog is not a matter of quality or proper. the old code look into the cache, then look onto the provider if in the cache find all episode return result and doesn't manage what have been retrieved by provider. The PR skipp search on provider if all episode is found in cache

fernandog:
@gborri imagine I want HDTV for initial and 1080p WEB for archive. If in the cache there's only HDTV, it will download it without looking the provider for 1080p web (available already)

gborri:
It happen the same with the actual code. I haven't changed the way it works.

Il giorno ven 17 apr 2015 00:10 Fernando <notifications@github.com> ha
scritto:

fernandog:
@TagForce thougts?

TagForce:
@gborri @fernandog What if not all episodes are on cache?
As far as I can tell, if the number of results is lower than the number of the episodes, it will iterate over all the episodes, regardless of if they were found on the cache or not...
So, let's say there's 2 episodes, one of which is on the cache, then in the new else statement it will go over the found one as well.

I don't think this speeds anything up. If they are on the cache they will simply not execute the code you moved (there's a continue at the end of the if). If there are any that are not on cache, ALL of them will be searched, which will undoubtedly take longer.

Right?

gborri:
You are right. i haven't realized that there was the continue code.

PR #1782, opened 2015-04-16
Change Opensubtitles 503 error to warning

fernandog:
@abeloin is this ok?

### DIFF
--- a/sickbeard/tv.py
+++ b/sickbeard/tv.py
@@ -1439,7 +1439,10 @@ class TVEpisode(object):
                         helpers.chmodAsParent(subtitle.path)
 
         except Exception as e:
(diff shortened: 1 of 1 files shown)


sickbeard/tv.py
```diff
@@ -1439,7 +1439,10 @@ def downloadSubtitles(self, force=False):
                         helpers.chmodAsParent(subtitle.path)
```
abeloin: except ServiceError as e:
            logger.log("Service is unavailable: {0}".format(str(e)), logger.INFO)
        except Exception:
            logger.log("Error occurred when downloading subtitles: " + traceback.format_exc(), logger.DEBUG)

PR #1786, opened 2015-04-17
FR# 774: Stay on Show Page after "Update Kodie"

Thraxis:
* changed KODI Update button to return to show page when clicked


Thraxis:
https://github.com/SiCKRAGETV/sickrage-issues/issues/774

PR #1788, opened 2015-04-17
Check if url error code is mapped

fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/1194

@Tagforce is it wrong?


TagForce:
@fernandog Move the def to inside the def for getURL...

So:
```
def getURL(blahblah):
    def codeDescription(codenumber):
        somecodetoexecute_in_codeDescription
    
    codetoexecute_in_getURL
```

That way you can leave the self. off of the call to codeDescription (which will now fail since codeDescription is not part of self.

fernandog:
@TagForce why I created defs there and it worked. Why I can't do it here?

https:// github.com /SiCKRAGETV/SickRage/pull/1532/files

TagForce:
@fernandog You can create it there...  It's just neater to define a function that is only used inside another function within that function.

fernandog:
@TagForce the problem is that we need def download_file too.

TagForce:
@fernandog My bad... I missed the definition for download_file...
Then leave it as is... Does it work with self? If so leave it, otherwise if it works without self. leave it out (it shouldn't need it as it's defined in the same file and not part of a class reference)

PR #1790, opened 2015-04-17
Fix NameError: global name 'ShowObj' is not defined

gborri:
as per subject


fernandog:
@TagForce

TagForce:
@fernandog Ah... Whoopsie... I tested it in a wrong branch, then switched to the correct one, adding that line by hand... I PRed that one without actually checking it. Made a typo :)

It's good that there are more people on this now, stuff like this gets found quicker :)

PR #1796, opened 2015-04-18
Show subtitle service unavailable as info

fernandog:


fernandog:
@abeloin this PR is causing this https://github.com/SiCKRAGETV/sickrage-issues/issues/1337

We need hotfix it.

abeloin:
@fernandog Simply import the serviceError from https://github.com/SiCKRAGETV/SickRage/blob/master/lib/subliminal/exceptions.py into sickbeard/tv.py

I can do it later today when I'll be back

PR #1804, opened 2015-04-20
Remove indexer from tvshow.nfo and series.xml

miigotu:
I think there is still some work to be done to make this process more proper.  I don't think that any other indexer_id should be stored in `<id>` other than tvdb, as kodi has no idea that it isn't thetvdb, and an art refresh from kodi could return the wrong show.

One problem of removing `<indexer>` (which needs to be done), is that Series.xml has no epg, or any urls at all, to check for theTVDB or TVRage and decide what indexer the `<id>` is referencing.

There should be a method to query each indexer with the provided `<id>` and compare it to the `<title>` we have, and set the indexer in SR db accordingly.  Once it is in our db, it never needs to be read again.

Finally, all of these problems exist prior to these changes, and this is a step in the right direction.

### DIFF
--- a/sickbeard/metadata/generic.py
+++ b/sickbeard/metadata/generic.py
@@ -286,14 +286,9 @@ class GenericMetadata():
                 with ek.ek(open, nfo_file_path, 'r') as xmlFileObj:
                     showXML = etree.ElementTree(file=xmlFileObj)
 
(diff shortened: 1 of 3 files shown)


fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/1205

PR #1801, opened 2015-04-20
Hotfix: Remove Success codes from http_error_code list

TagForce:
Fixes erroneous error messages in download clients, causing nzb/torrent
sending to not finish succesfully and causing extra settings to not be
sent (like labels).

Effectively removes http status codes in the 100 - 299 range from the
list (those are success messages and should not be handled as errors)

https://github.com/SiCKRAGETV/sickrage-issues/issues/1331

PR #1805, opened 2015-04-21
Merge pull request #1801 from TagForce/master

lee234:
T411: Catch invalid data, log, and continue


abeloin:
@lee234 What do you want to do exactly?

lee234:
I was trying to undo my branch so I could update to the latest master.
Sorry if I caused any trouble. I finally just deleted the branched file and
it worked.

Is there an easy option to overwrite any local changes with a remote master?

Thanks Bazl

abeloin:
git reset --hard

PR #1807, opened 2015-04-21
Unit test: change journal mode when in unit test

abeloin:
Use journal_mode MEMORY when in unit test


miigotu:
See https://github.com/miigotu/SickRage/commit/5521490d1f5314de62ff65d3299811db936b54fb for what I propose. Should create a PR with that?

Tests all work, SR seems to run faster. Noticed no issues.

fernandog:
@miigotu merged

PR #1808, opened 2015-04-21
Fix Plex Authentication Token

Arcanemagus:
Fixes https://github.com/SiCKRAGETV/sickrage-issues/issues/1348 and allows the use of just an authentication token for Plex Media Servers again.

Further work is needed to get the notifications working properly again, currently this reports a connection error, but still talks to the server properly.

### DIFF
--- a/sickbeard/notifiers/plex.py
+++ b/sickbeard/notifiers/plex.py
@@ -184,12 +184,12 @@ class PLEXNotifier:
                 password = sickbeard.PLEX_PASSWORD
                 
             if not plex_server_token:
-                token = sickbeard.PLEX_SERVER_TOKEN
+                plex_server_token = sickbeard.PLEX_SERVER_TOKEN
             
             # if username and password were provided, fetch the auth token from plex.tv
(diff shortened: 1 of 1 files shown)


Arcanemagus:
No, the code does _not_ work as it is. If you only specify an authentication token it attempts to connect to the server without specifying the token in the request.

Before the fix in this commit is implemented the following behavior is observed:
Auth Token only:

    http://plex.localnet.local:32400/library/sections

Credentials:

    http://plex.m4gusnet.local:32400/library/sections?X-Plex-Token=<plex.tv token value>

After the above fix is put in place the _structure_ of the request URL's are the same, however it seems the token being returned from plex.tv for the authentication credentials is different from the one obtained from a library item's XML. (Thus providing a test to ensure that it is properly utilizing the credentials...)

Thraxis:
@abeloin @fernandog This is a good fix.   Plex Auth Tokens for non PlexPass PMS  mean nothing as it will accept anything at all as valid for the token.

PR #1809, opened 2015-04-21
Fix Plex notification status check

Arcanemagus:
The return value of update_library() was changed from True/False to None/Error string in https://github.com/SiCKRAGETV/SickRage/commit/4c9d9185b0c25d435a6bbb20efa60b1f4337a016 but the check for the return value was never updated. This corrects that so the notification sent out correctly states that it sent a notification to the server instead of defaulting to stating it was unable to connect.

This fixes the rest of the issues described in https://github.com/SiCKRAGETV/sickrage-issues/issues/1348.

### DIFF
--- a/sickbeard/webserve.py
+++ b/sickbeard/webserve.py
@@ -1708,7 +1708,7 @@ class Home(WebRoot):
             return self.redirect('/home/')
 
     def updatePLEX(self):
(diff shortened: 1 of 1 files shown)


Arcanemagus:
Sent to the wrong branch, sorry.

PR #1810, opened 2015-04-21
Fix Plex notification status check

Arcanemagus:
The return value of update_library() was changed from True/False to None/Error string in https://github.com/SiCKRAGETV/SickRage/commit/4c9d9185b0c25d435a6bbb20efa60b1f4337a016 but the check for the return value was never updated. This corrects that so the notification sent out correctly states that it sent a notification to the server instead of defaulting to stating it was unable to connect.

This fixes the rest of the issues described in SiCKRAGETV/sickrage-issues#1348.

### DIFF
--- a/sickbeard/webserve.py
+++ b/sickbeard/webserve.py
@@ -1708,7 +1708,7 @@ class Home(WebRoot):
             return self.redirect('/home/')
 
     def updatePLEX(self):
-        if notifiers.plex_notifier.update_library():
+        if None is notifiers.plex_notifier.update_library():
             ui.notifications.message(
                 "Library update command sent to Plex Media Server host: " + sickbeard.PLEX_SERVER_HOST)
         else:


Thraxis:
That function is tied only to the force update button not the call made to update Plex during post processing.  The messages generated are UI notifications... little boxes that pop up on the screen.   I will agree that yes it needs to be fixed.  But your change only partially fixes the issue as it doesn't actually use the return to display the hosts that did have an error with the update.

Arcanemagus:
This doesn't seem to be working properly for me, so I'll leave fixing that to the people that actually use Python and know how to debug it:

    def updatePLEX(self):
        res = notifiers.plex_notifier.update_library()
        if None is res:
            ui.notifications.message(
                "Library update command sent to Plex Media Server host(s): " + sickbeard.PLEX_SERVER_HOST)
        else:
            ui.notifications.error("Unable to contact Plex Media Server host(s): " + res)
        return self.redirect('/home/')

----
When setting an invalid IP address as another server it properly times out and says that it has failed when pressing the Test button from the settings, but the "Update PLEX" button on the main page immediately shows the success message for the real server and never notifies that it failed to send the command to the fake server.

If there is only one (invalid) server the above code works as intended though, but then it would be identical to the current code in behavior anyway.

PR #1814, opened 2015-04-22
HotFix for: local variable 'subtitles' referenced before assignment

TagForce:
https://github.com/SiCKRAGETV/SickRage/pull/1796 adds a check for
Exception ServiceError, but doesn't return on the exception. For every
other exception the function is returned from.

Fixes https://github.com/SiCKRAGETV/sickrage-issues/issues/1359


TagForce:
This is a hotfix... Committed against the current master.

PR #1818, opened 2015-04-22
BTN - Show warning instead of error for api call exceeded

fernandog:
@SiCKRAGETV/moderators any of you have BTN to test this?


fernandog:
@gangrenejellybean can you test this please? it's a simple change in btn file

miigotu:
@fernandog that looks safe from here.

PR #1819, opened 2015-04-22
Fix scenetime provider for some people

bjornhamra:
Apparently some people don't have the 'Download' and 'Bookmark' columns in their scenetime torrent listing. The previous commit assumed the seeders/leechers fields would always be columnd 7 & 8 respectively, but for some users this is apparently not true. This commit fixes that by referencing the header columns so that the column id is derived from which header column contains the respective field, instead of hardcoding the cell IDs.


fernandog:
@bjornhamra please squash commits into one

bjornhamra:
My git-fu is severely lacking. Let's try that again.

fernandog:
@bjornhamra user reported it worked. Waiting new PR

PR #1820, opened 2015-04-23
Fixup regex for issue submitter, Dont allow dupe app submitted issues. More

miigotu:
Duplicate submitted issues are now comments to the original instead of
duplicating. ***DO NOT CHANGE APP SUBMITTED ISSUE TITLES FOR NOW**
Maybe later we can move to parsing issue comments to match errors and
make it ok to change APP SUBMITTED issue titles, but currently I have it using
the issue title to search for originals.

Fix some unit tests not running some or all of their tests when ran standalone.

See how it just adds a comment instead of a  new issue in SiCKRAGETV/sickrage-issues/issues/1392

Also, the gists it pasts are just the error it finds, not the entire log, which was the original writer's intent.

### DIFF
--- a/sickbeard/logger.py
+++ b/sickbeard/logger.py
@@ -151,6 +151,7 @@ class Logger(object):
 
     def submit_errors(self):
         if not (sickbeard.GIT_USERNAME and sickbeard.GIT_PASSWORD and len(classes.ErrorViewer.errors) > 0):
+            self.log('Please set your GitHub username and password in the config, unable to submit issue ticket to GitHub!')
             return
 
         gh_org = sickbeard.GIT_ORG or 'SiCKRAGETV'
@@ -184,7 +185,7 @@ class Logger(object):
                     title_Error = u"Unable to extract title from error"
 
                 gist = None
-                regex = "^(%s)\s*([A-Z]+)\s*(.+?)\s*\:\:\s*(.*)$" % curError.time
+                regex = "^(%s)\s+([A-Z]+)\s+([0-9A-Z\-]+)\s*(.*)$" % curError.time
                 for i, x in enumerate(log_data):
                     x = ek.ss(x)
                     match = re.match(regex, x)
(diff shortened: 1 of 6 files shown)


sickbeard/logger.py
```diff
@@ -151,6 +151,7 @@ def log_error_and_exit(self, error_msg, *args, **kwargs):
 
     def submit_errors(self):
         if not (sickbeard.GIT_USERNAME and sickbeard.GIT_PASSWORD and len(classes.ErrorViewer.errors) > 0):
+            print u'Please set your GitHub username and password in the config, unable to submit issue ticket to GitHub!'
```
abeloin: @miigotu This should be a logger.log instead of a print

miigotu:
@fernandog what do you think about this?

miigotu:
oops, fixing.

miigotu:
@abeloin @fernandog Should be all gravy now.

fernandog:
@miigotu thanks! Log file fixed and None title in some cases too? @tagforce find out that in some cases the title is None or Empty.

fernandog:
@miigotu the problem is log rotation. That's why we read all log files. If the error is in sickrage.log.1 after a log rotation , the submit wouldn't work when user press "submit errors"

fernandog:
@miigotu do we need to update this regex also?

https://github.com/SiCKRAGETV/SickRage/blob/develop/sickbeard/webserve.py#L5039

miigotu:
@fernandog That regex is working for every line in my log, however it could be improved for performance by changing it to this: https://regex101.com/r/mE8wS5/1
Notice the "X Match - Y Steps" at the top, now look at the original:
https://regex101.com/r/hK2tX4/1

This regex is used for each line in the log, so a 17 step savings in this particular regex is a savings of approximately 30% processing power. Less instructions = More performance.

It could be improved further by not capturing the date and time as well, which saves two steps.
Either way, if changing the regex then the capture group numbers would need changed to correctly represent the match we are looking for in:

https://github.com/SiCKRAGETV/SickRage/blob/develop/sickbeard/webserve.py#L4981
https://github.com/SiCKRAGETV/SickRage/blob/develop/sickbeard/webserve.py#L4982

It works as it is, but performance can be improved. Shall we leave it alone or who shall PR?

fernandog:
@abeloin what do you think?

fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/1282

PR #1821, opened 2015-04-23
Check file lock

TagForce:
- When copying or hardlinking checks if the source file is locked for
reading
- When moving or moveAndSymlinking checks if the source file is locked
for reading and writing

### DIFF
--- a/sickbeard/postProcessor.py
+++ b/sickbeard/postProcessor.py
@@ -1017,15 +1017,21 @@ class PostProcessor(object):
         try:
             # move the episode and associated files to the show dir
             if self.process_method == "copy":
+                if isFileLocked(self.file_path, False):
+                    raise exceptions.PostProcessingFailed("File is locked for reading")
                 self._copy(self.file_path, dest_path, new_base_name, sickbeard.MOVE_ASSOCIATED_FILES,
                            sickbeard.USE_SUBTITLES and ep_obj.show.subtitles)
             elif self.process_method == "move":
+                if isFileLocked(self.file_path, True):
+                    raise exceptions.PostProcessingFailed("File is locked for reading/writing")
                 self._move(self.file_path, dest_path, new_base_name, sickbeard.MOVE_ASSOCIATED_FILES,
                            sickbeard.USE_SUBTITLES and ep_obj.show.subtitles)
             elif self.process_method == "hardlink":
                 self._hardlink(self.file_path, dest_path, new_base_name, sickbeard.MOVE_ASSOCIATED_FILES,
                                sickbeard.USE_SUBTITLES and ep_obj.show.subtitles)
             elif self.process_method == "symlink":
+                if isFileLocked(self.file_path, True):
+                    raise exceptions.PostProcessingFailed("File is locked for reading/writing")
                 self._moveAndSymlink(self.file_path, dest_path, new_base_name, sickbeard.MOVE_ASSOCIATED_FILES,
                                      sickbeard.USE_SUBTITLES and ep_obj.show.subtitles)
             else:
--- a/sickbeard/postProcessor.py
+++ b/sickbeard/postProcessor.py
@@ -910,9 +910,13 @@ class PostProcessor(object):
                 logger.DEBUG)
         
         # try to find out if we have enough space to perform the copy or move action.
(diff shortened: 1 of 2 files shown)


TagForce:
@fernandog @abeloin @miigotu @Thraxis 
Needs testing by as many of you guys as possible, the feature is windows only at the moment.

sickbeard/postProcessor.py
```diff
@@ -1017,15 +1021,21 @@ def process(self):
...
+                    raise exceptions.PostProcessingFailed("File is locked for reading")
                 self._copy(self.file_path, dest_path, new_base_name, sickbeard.MOVE_ASSOCIATED_FILES,
                            sickbeard.USE_SUBTITLES and ep_obj.show.subtitles)
             elif self.process_method == "move":
+                if helpers.isFileLocked(self.file_path, True):
+                    raise exceptions.PostProcessingFailed("File is locked for reading/writing")
                 self._move(self.file_path, dest_path, new_base_name, sickbeard.MOVE_ASSOCIATED_FILES,
                            sickbeard.USE_SUBTITLES and ep_obj.show.subtitles)
             elif self.process_method == "hardlink":
                 self._hardlink(self.file_path, dest_path, new_base_name, sickbeard.MOVE_ASSOCIATED_FILES,
```
miigotu: Shouldn't we check if the dest_path is locked? For move and copy too....

sickbeard/postProcessor.py
```diff
@@ -1017,15 +1021,21 @@ def process(self):
...
+                if helpers.isFileLocked(self.file_path, False):
+                    raise exceptions.PostProcessingFailed("File is locked for reading")
                 self._copy(self.file_path, dest_path, new_base_name, sickbeard.MOVE_ASSOCIATED_FILES,
                            sickbeard.USE_SUBTITLES and ep_obj.show.subtitles)
             elif self.process_method == "move":
+                if helpers.isFileLocked(self.file_path, True):
+                    raise exceptions.PostProcessingFailed("File is locked for reading/writing")
                 self._move(self.file_path, dest_path, new_base_name, sickbeard.MOVE_ASSOCIATED_FILES,
                            sickbeard.USE_SUBTITLES and ep_obj.show.subtitles)
             elif self.process_method == "hardlink":
```
miigotu: Shouldn't we check if the dest_path is locked in hardlink? For move and copy too....

TagForce: Nope, not here anyway.
If the destination file is locked you're doing something wrong, because the file shouldn't exist (and if it did, it gets removed elsewhere in the code). All this does is make sure that the source file is accessable for whatever operation you want to perform on it.
I'll add a proper check for write permissions on the destination later on. I also want to move all the file operations to one place, to make sure nothing can go wrong between deletion of the old files and placement of the new ones.

For now, this only checks the source file, and for hardlink you actually don't need access to the file itself. You can add a hardlink to a file that is completely inaccessable. Unfortunately that hardlink will become inaccessable itself, but the creation works. That's why I skipped the test for hardlinks. At least in windows.

miigotu:
Yeah I just meant for checking write perm on the destination dir. But if you are adding that at another time then cool.

fernandog:
@TagForce let's merge into develop?

fernandog:
@TagForce should we merge into develop?

TagForce:
@fernandog I'm not sure this will work correctly on Linux... Needs to be tested, but I haven't yet had time to do that properly.

fernandog:
@TagForce @miigotu AFAIK this only happens in windows. why use WindowsError and etc?

miigotu:
Without WindowsError it won't catch the exception on windows if it cant rename the file. On linux, if there is a permissions error or the file is in use/locked it will throw either IOError or OSError, depending on circumstance.

The lock check is tested on all platforms, and it will not catch an exception if renaming the file fails on linux as it sits now.

miigotu:
@TagForce Can/Should we merge this? If we need to add some more checking another time we can.

fernandog:
@TagForce please?

miigotu:
@fernandog I think its better to merge this, then add the checks in another PR.

fernandog:
@miigotu I don't use PP so can't test this in linux

miigotu:
Just going to merge it, and we will see. I looked over the code and it seems proper, except for the extra excepts. This does not break anything, as the check wasn't there at all before and worked fine for me in linux anyways.

PR #1825, opened 2015-04-24
FR#1078 Update Specific Plex Sections

Thraxis:
* Sends Episode Object on Post process
* If the path of the PLEX Show Section is found within the Episode
Object Location,  only that section is sent an update request.
otherwise all sections are.

Note: Functionality was introduced in the Plex Notification update,
sending Episode Object on Post Process was not included at that time.


fernandog:
@Thraxis safe to merge?

PR #1826, opened 2015-04-24
FR#574 - Home Page - Show List Filtering

Thraxis:
* Return episode count weighted sorting
* Episode 'Downloads' field can be filtered by Percent value now
* supports >, >=, <=, < ,  xx to yy , xx - yy as well as exact value
with or without pre / post fix of '='

Note: >, >=, <=, < should be first, followed by a space, then the value.
ie.  '>= 50'


fernandog:
@Thraxis safe to merge?

PR #1827, opened 2015-04-24
FR#1040 - Collapse passed seasons

Thraxis:
* Adds an option in general config to collapse all but the current
season
* Buttons to show and hide seasons added to the season headers

### DIFF
--- a/gui/slick/interfaces/default/config_general.tmpl
+++ b/gui/slick/interfaces/default/config_general.tmpl
@@ -285,7 +285,15 @@
 								</span>
 							</label>
 						</div>
-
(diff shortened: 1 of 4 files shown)


fernandog:
@Thraxis can we collpase/uncollapse by demand or is it a on/off option?

Thraxis:
@Fernandog You enable the feature in general -> interface and from then on when you go into a show all but the current season is collapsed.  You can uncollapse / collapse  the other seasons on demand clicking a button.

fernandog:
@Thraxis can we align the "Show Episodes" button next to "Season x" ?

PR #1830, opened 2015-04-24
Fixes #1401 - SD Blueray not matching

miigotu:


sickbeard/common.py
```diff
@@ -212,7 +212,7 @@ def sceneQuality(name, anime=False):
             return Quality.SDTV
         elif checkName(["web.dl|webrip", "xvid|x264|h.?264"], all) and not checkName(["(720|1080)[pi]"], all):
             return Quality.SDTV
-        elif checkName(["(dvdrip|b[r|d]rip)(.ws)?.(xvid|divx|x264)"], any) and not checkName(["(720|1080)[pi]"], all):
+        elif checkName(["(dvdrip|b[rd]rip|blueray)(.ws)?.(xvid|divx|x264)"], any) and not checkName(["(720|1080)[pi]"], all):
```
fernandog: @miigotu why remove the | ?  we have BDRip and BRRip. Or it will match any of them?

miigotu: Its a character group, you don't need an |.
`[this]` would match a single character, one of `t`,`h`,`i`, OR, `s`

gangrenejellybean:
I think the use of "blueray" with an e is going to be an issue here as it's not going to match against "bluray" which is how it's usually written.

I think using "blu-?ray" or "blue?-?ray" is the best bet.

Sorry I didn't catch this before the merge. Thanks for handling this so quickly, though.

miigotu:
grr, stupid typo,  but yeah "blue?-?ray" is better since I think I have seen blu-ray before.

PR #1305, opened 2015-02-12
Update piratebay url to current

res1n:
Oldpiratebay.org is a few days behind on new show releases, claims there's nothing to download.   Once changing to original URL things seem to work just fine.  Please review. :)

PR #1833, opened 2015-04-24
Block issue submitter from running more than once at a time. Fixes #1305

miigotu:
If user presses button to submit again while its already processing/submitting, they get a notification to wait.


miigotu:
https://github.com/SiCKRAGETV/sickrage-issues/issues/1305

PR #1836, opened 2015-04-25
Enable trending shows only if Trakt is enabled

fernandog:


fernandog:
@miigotu forgot how long does it take to travis build....

PR #1842, opened 2015-04-25
FIX#1292 Masking NZB API Keys in logs

Thraxis:
* Modified Regex substitution to account for 3 possible keys that the
API can be attached to
* Modified Regex to account for API Key being at the end of the string
or followed by an & or a space
* Used capture groups to only mask out the API Key, matched prefix and
ending are kept


fernandog:
@Thraxis lots of PR! thanks!

PR #1849, opened 2015-04-26
FR#1364: Seperate Plex Media Server and Plex Media Client

Thraxis:
* Seperated PMS and PMC in the config page
* Created seperate username / password variables for PMC


fernandog:
@Thraxis thanks!

Later can you add some text label in UI explaining "Plex Media Client". All the others have.

Thraxis:
Yeah.. just have to think of what to put there

PR #1850, opened 2015-04-26
Make check_setting_int understand True/False from settings

miigotu:
To help prevent confusing users modifying the config directly.
With this, any case variation of 'true' will work the same as 1, and 
any case variation of 'false' will work the same as 0.

### DIFF
--- a/sickbeard/config.py
+++ b/sickbeard/config.py
@@ -442,7 +442,14 @@ def minimax(val, default, low, high):
 ################################################################################
 def check_setting_int(config, cfg_name, item_name, def_val, silent=True):
     try:
-        my_val = int(config[cfg_name][item_name])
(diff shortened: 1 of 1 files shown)


abeloin:
@miigotu What happen if it's not true or false(or str(0) or str(1))?

miigotu:
It would have the same behavior as before in that case. I do no checking if it is an invalid string, so the int returned would be the integer value of the string, like it did before.
str(0) would come back as int(0)

my_val = int(my_val)

I only check if its "true" or "false" before converting it to int, where before it just converted to int in the same line as assignment to my_val.

abeloin:
@miigotu Sorry too many hours in overtime this week, I'm starting to see code that doesn't exists ;)

miigotu:
lol np

PR #1856, opened 2015-04-27
if re.search returns None due to no match, .group() will cause an exception

miigotu:
Fix SiCKRAGETV/sickrage-issues/issues/1437


nicolasmartinelli:
Hi,
It prevents a Python error, but you will miss potential links with this commits. I made something more flexible: #1858

miigotu:
@nicolasmartinelli My commit will not cause SR to miss any links that it previously caught. It preserves original functionality while hotfixing an error situation.

For clarity, we separate fixes from improvements, and fixes must come first. I am not sure your code improves operation or not, but its a separate issue from simply fixing an error case.

PR #1858, opened 2015-04-27
Add flexibility when determine title from link in eztvapi.re:

nicolasmartinelli:
1) try magnet link
    2) try rarbg link
    3) try extratorrent link
    4) try '([^/]+$)' : everything after last slash character (not accurate)
    5) fallback, title is equal to link
Fixes https://github.com/SiCKRAGETV/sickrage-issues/issues/1437


nicolasmartinelli:
re.search converted into re.match, since it is more efficient.

nicolasmartinelli:
Removed unnecessary ^

PR #1864, opened 2015-04-30
Make sure we are using our included libs, and not system libs, by provid...

miigotu:
...ing the full path on import


miigotu:
This gets rid of the `InsecureRequestWarning` and `InsecurePlatformWarning`, as the call to `urllib3.disable_warnings()` was called to the system urllib3, and sometimes we used one or the other, and causing craziness overall in requests. Also fixes some other errors.

PR #1871, opened 2015-05-02
Add pyopenssl libs, and tests

miigotu:
Added the 3 libraries required to use pyopenssl with SNI.
Make tests only touch the sys.path if they are run directly.
Add a test to check ability to connect to certain SNI sites we have open issues for.
Fix feed parser test, using a random feed from showrss.info

I removed ndg.httpsclient, pyasn1, and pyopenssl from my python environment and tested this alot. Still needs the normal openssl binary installed on the system.

Should this wait until after develop->master merge tomorrow?


miigotu:
Do not merge till I figure out wtf is going on.

PR #1872, opened 2015-05-02
Add pyopenssl libs, and a inittest

miigotu:
Added the 3 libraries required to use pyopenssl with SNI.
Add a test to check ability to connect to certain SNI sites we have open issues for.

I removed ndg.httpsclient, pyasn1, and pyopenssl from my python environment and tested this alot. Still needs the normal openssl binary installed on the system, and cryptography installed through pip.
Installing cryptography requires gcc, python-dev, libssl-dev, libffi-dev, the rest are installed by pip.

Should this wait until after develop->master merge tomorrow?


miigotu:
FOOOOK. I got it working with 2.7, but not 2.6.

miigotu:
2.6 Needs cryptography, which installs like 6 other dependancies. I'll add a try/except to import cryptography on 2.6, and skip the test if it fails.

PR #1873, opened 2015-05-04
SCC doesn't support searching dates with pipe characters. Use '.'

adampetrovic:
As the commit message says, currently SickRage is attempting to search dates with the pipe character as a delimiter, which isn't supported by SCC.

E.g. 
```Jimmy.Fallon.2015|04|27``` returns no results
```Jimmy.Fallon.2015.04.27``` returns 2 results

screen shot 2015-05-04 at 12 22 36 pm
screen shot 2015-05-04 at 12 22 51 pm

### DIFF
--- a/sickbeard/providers/scc.py
+++ b/sickbeard/providers/scc.py
@@ -130,12 +130,12 @@ class SCCProvider(generic.TorrentProvider):
         if self.show.air_by_date:
             for show_name in set(show_name_helpers.allPossibleShowNames(self.show)):
                 ep_string = sanitizeSceneName(show_name) + ' ' + \
-                            str(ep_obj.airdate).replace('-', '|')
+                            str(ep_obj.airdate).replace('-', '.')
                 search_string['Episode'].append(ep_string)
         elif self.show.sports:
             for show_name in set(show_name_helpers.allPossibleShowNames(self.show)):
                 ep_string = sanitizeSceneName(show_name) + ' ' + \
-                            str(ep_obj.airdate).replace('-', '|') + '|' + \
+                            str(ep_obj.airdate).replace('-', '.') + '|' + \
                             ep_obj.airdate.strftime('%b')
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by adampetrovic** [comment]
Tested and working.

File: sickbeard/providers/scc.py [hunk 1]
```diff
@@ -130,12 +130,12 @@ def _get_episode_search_strings(self, ep_obj, add_string=''):
...
             for show_name in set(show_name_helpers.allPossibleShowNames(self.show)):
                 ep_string = sanitizeSceneName(show_name) + ' ' + \
-                            str(ep_obj.airdate).replace('-', '|')
+                            str(ep_obj.airdate).replace('-', '.')
                 search_string['Episode'].append(ep_string)
         elif self.show.sports:
             for show_name in set(show_name_helpers.allPossibleShowNames(self.show)):
                 ep_string = sanitizeSceneName(show_name) + ' ' + \
-                            str(ep_obj.airdate).replace('-', '|') + '|' + \
+                            str(ep_obj.airdate).replace('-', '.') + '|' + \
```
adampetrovic: I have a feeling this will not work either. As SCC will treat it as a literal rather than an 'OR' operation as is intended.

fernandog:
torrentday had that problem too. I fixed doing two searches instead of just one with OR

miigotu:
It isn't supposed to be 'OR' there is it? `2015 OR 03 OR 10` doesn't make sense. I think its better to replace them with a space like in kay.py, oldpiratebay.py, and thepiratebay.py, which should work and keep a little conformity.

adampetrovic:
I don't think conformity matters too much in this case. After all they are separate providers for a reason. I would argue that when you are searching by date, it should match the torrent name as much as possible, otherwise you risk erroneous results popping up.

miigotu:
Ok, I did some searching myself on SCC, I agree with you now.

PR #1877, opened 2015-05-04
Add __init__.py to autoProcessTV/libs so absolute lib paths work.

miigotu:


miigotu:
@Molleke This should fix your problem.

Molleke:
Thanks will try it again.

PR #1881, opened 2015-05-05
Temp fix for torrentday

aatje92:
Torrentday has a new layout and as a result sickrage is broken. This is a temporary fix as i am not nearly smart enough to create a full fix, but it works for now. might not work for users in UK and users who have to use tdonline.org


coolius:
I'm in the UK and can access TD via classic.torrentday.com which is a good sign

fernandog:
@coolius do we need to merge this? TD is ok here

coolius:
It was definitely broken for a bit, but seems to be working again now for me at least.

fernandog:
in the future we can reopen this if we need

PR #1883, opened 2015-05-05
Fix to issue 1460

fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/1460

### DIFF
--- a/sickbeard/helpers.py
+++ b/sickbeard/helpers.py
@@ -844,7 +844,7 @@ def restoreVersionedFile(backup_file, version):
 
 
 # try to convert to int, if it fails the default will be returned
-def tryInt(s, s_default=0):
+def tryInt(s, s_default=1):
     try:
         return int(s)
     except:


fernandog:
@miigotu do you think it's ok?

miigotu:
That fixes the value error, but masks the real problem I think @fernandog. Plus, it would display the airdate as 01/01/0001 then. And that would be not only confusing but mess up sorting and 'coming episodes' page.

fernandog:
@miigotu this issue is odd. sometimes works, sometimes not. I guess it's related to XEM connectivity issues.

miigotu:
@fernandog Maybe when air date is not provided by tvdb/tvrage, because they dont have the info?

fernandog:
@miigotu the actual code is wrong. s_default=0 will ALWAYS fail

miigotu:
@fernandog It is intended to fail, because we want to be sure that an incorrect date is not added to the db right? If we let it default to 1, then we will never get an error telling us that the date is failing to be populated somewhere else in the code.

fernandog:
@miigotu the problem is the page won't load at all and user won't know what's happening. Log only show traceback without any show info or message. any other solution?

fernandog:
@miigotu also the indexer has correct airdate. So it's not a problem there. Maybe It's a connectivity issue as I said above and SR couldn't fetch the data

miigotu:
@fernandog I think we should find where the date is not being set correctly, and provide an alternative way of getting it. iirc this was only happening when xem was down?

fernandog:
@miigotu I didn't had this issue anymore. Can't debug it anymore

miigotu:
@fernandog maybe we should just close this for now until if/when the issue resurfaces? We can always re-open and rebase this.

fernandog:
@miigotu if this default=0 wil ALWAYS raise an error why set it ?

"def tryInt(s, s_default=0)"

miigotu:
Because it is left there to catch us passing invalid values there by accident. If a -1 gets to here we have a problem somewhere else and we want to know about it instead of workaround it and allow a more complex problem to develop down the road.

miigotu:
Changing this just masks the real problem in SiCKRAGETV/sickrage-issues#1460

PR #1889, opened 2015-05-07
Fix#1453 - Fix 'NoneType' object has no attribute 'whitelist

Thraxis:
There should have always been a check if the show was anime here.  As
only anime had white and blacklists.  Added both check for is_anime and
that release_groups is not none


fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/1453

dave-tucker:
This doesn't fix the issue completely, there is a missing check on https://github.com/Thraxis/SickRage/blob/Fix%231453/sickbeard/show_name_helpers.py#L203

miigotu:
@Thraxis Status? what about the comment above by @dave-tucker ?

Thraxis:
Added in the second set of checks.. not sure why there was never a check for is_anime at the least there to begin with.

PR #1892, opened 2015-05-08
Change sr_tvrage_scene_exceptions to use SiCKRAGETV repo instead of e…

miigotu:
…chel0n's  

echel0n's repo works, but noticed the url needed changed in SiCKRAGETV/sickrage-issues#1541


fernandog:
@miigotu we can't do this because we don't update this file. 

we should use this: https://raw.githubusercontent.com/midgetspy/sb_tvdb_scene_exceptions/gh-pages/exceptions.txt

original sickbeard updates it

PR #1897, opened 2015-05-10
Fix#1558 Fix Quality Colors on History and displayShow pages

Thraxis:
Fixes Quality colors due to changes for "funky naming"

PR #1899, opened 2015-05-12
Fix broken UI labels since SiCKRAGETV/sickrage-issues#1508 (regression)

svenvanhal:
The RawHD and HDTV labels in the episode overview on ```displayShow``` would not have a background color because the class name is not replaced properly.

sickrage


miigotu:
This was fixed by @Thraxis with #1897 wasn't it? You also missed them in history.tmpl. Also, we don't merge to master from your master. See https://github.com/SiCKRAGETV/SickRage/blob/master/contributing.md#pull-requests

svenvanhal:
Ah, must have missed that. Will pay better attention next time.

miigotu:
No problem, thanks for playing xD

PR #1901, opened 2015-05-12
Network logos

matigonkas:
ESPN2 & National Geographic


neoatomic:
@matigonkas You cant use uppercase characters in the filename. All characters must be lowercase.
Did you test before submitting.?

PR #1903, opened 2015-05-12
Issue 1538 dev

gosku:
Hi,

I improved the Failed Handler to manage with the situations like the ones I indicated here:
SiCKRAGETV/sickrage-issues#1538

Mainly, I check if the provider is returning the candidates size. If not, it uses a hash coming from the url to differentiate when a candidate is different to another with the same release name.

I add a field "url" in failed and history tables (failed.db) to store the hashes.

Please, let me know if there is something you want I change, or if there are any conflict.

### DIFF
--- a/sickbeard/databases/failed_db.py
+++ b/sickbeard/databases/failed_db.py
@@ -40,11 +40,12 @@ class InitialSchema(db.SchemaUpgrade):
 
 class SizeAndProvider(InitialSchema):
     def test(self):
-        return self.hasColumn('failed', 'size') and self.hasColumn('failed', 'provider')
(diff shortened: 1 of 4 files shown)


fernandog:
@gosku 
What if the release is bad? not just the file.
torrent and nzb.
url will be diff for them

gosku:
You don't know whether the entire release is bad or only one file is bad until you check the different candidates. The "badness" depends on the search provider or tracker you use, the usenet server, missing parts, the retention... There are many reasons for a download to fail and pressuposing the reason is "because the entire release is bad" could you leave you without an episode you want (as it happened to me several times).
In the worst case, and only for the providers that don't give you the file size, the handler would try to download all the candidates with different url for a bad release.
However, **it solves the problem of it not downloading a valid file** when the provider doesn't give you the file size.

PR #1906, opened 2015-05-13
Issue #1538 - Failed Download Handler Fix

gosku:
Hi,

**I hope this time I did it right**

I improved the Failed Handler to manage with the situations like the ones I indicated here:
https://github.com/SiCKRAGETV/sickrage-issues/issues/1538

Mainly, I check if the provider is returning the candidates size. If not, it uses a hash coming from the url to differentiate when a candidate is different to another with the same release name.

I add a field "url" in failed and history tables (failed.db) to store the hashes.

Please, let me know if there is something you want I change, or if there is any conflict.

### DIFF
--- a/sickbeard/databases/failed_db.py
+++ b/sickbeard/databases/failed_db.py
@@ -40,11 +40,12 @@ class InitialSchema(db.SchemaUpgrade):
 
 class SizeAndProvider(InitialSchema):
     def test(self):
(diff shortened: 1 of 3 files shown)


miigotu:
Much better, now I can see exactly what you changed easily.

miigotu:
Superceded by #1918

PR #1910, opened 2015-05-15
EZTV Provider: save all possible links found in the homepage

nicolasmartinelli:


miigotu:
@nicolasmartinelli can you explain why we need this pls? For example, why isn't the first link for each entry good enough?

nicolasmartinelli:
@miigotu  Thinking about it, I only need to save the first link found. I added a `continue` statement at the end of the loop.

miigotu:
Better. =P

fernandog:
@nicolasmartinelli should we remove EZTV completely?

nicolasmartinelli:
@fernandog : I don't any reason why it should be removed. So far, EZTV has been providing real link, and no scam/fake has been reported. It has been a pain in the ass since they have been changing the website quite often during the past 4 weeks, but the content is still fine.

miigotu:
@nicolasmartinelli he asks this because eztv is no longer ran by eztv. There was a hostile takeover.

pauldcox:
Specifically, EZTV as we knew it no longer exists: https://torrentfreak.com/eztv-shuts-down-after-hostile-takeover-150518/

miigotu:
I think we should just remove it, for everyone's safety. There is always the chance that some government agency piloted the takeover, which would explain how easy it was for them to take all of the domains, and why the registrars were complicit in the attack. Do we really want to risk any chance that our users are being logged over the next year or two and then targeted?

nicolasmartinelli:
Indeed, Torrent Freak reported that a few weeks ago. Note that there is no
official statement of any kind, just some suspicion from torrent freak.

By the way, eztvapi, which is used for episodes, is not affiliated with
eztv... It is managed by the Popcorn Time team.

I think a vague suspicion from a website is a bit light to completely
remove a provider...

nicolasmartinelli:
OK, just saw the new report. Still, eztvapi is still not affiliated with
eztv.
For me, as long as they don't provide fakes or scam, I don't see the reason
to not use it. An illegal website has been replaced by another illegal
website... Basically, all providers are using illegal websites, and you can
never be sure of who is running it.
Le 20 mai 2015 08:17, "Nicolas Martinelli" <nicolas.martinelli@gmail.com> a
écrit :

> Indeed, Torrent Freak reported that a few weeks ago. Note that there is no
> official statement of any kind, just some suspicion from torrent freak.
>
> By the way, eztvapi, which is used for episodes, is not affiliated with
> eztv... It is managed by the Popcorn Time team.
>
> I think a vague suspicion from a website is a bit light to completely
> remove a provider...

miigotu:
@nicolasmartinelli If for example someone was buying weed from a guy (illegal) for a long time, and they trusted this guy because they knew him well, and he was a well known no bs weed dealer, and then all of a sudden, this weed dealer stops showing up when you call him, and you never speak to him again, but when you call him, some other guy you never saw before shows up, do you buy the weed from this guy you never saw before?

nicolasmartinelli:
@miigotu : I would not buy weed from the first guy either, since if I relate to a torrent website, everybody including the authorities knows he is a dealer.

My point is that the authorities have already a lot of ways to monitor the P2P traffic, by simply letting a daemon connect to a tracker and collect IP addresses. An average developer can do such a software in a few hours. P2P is intrinsically NOT safe, whatever the website/tracker you connect to.

nicolasmartinelli:
Let me be more detailed. You can fear two things with the new EZTV:
- Fakes/scam: if it happens, obviously the provider needs to be removed since the links are unusable. End of the story.
- A takeover by authorities: while I doubt this is the case, I don't see what are the additional risks you encounter. When you download a torrent, your IP address is public, so anybody can gather this information and make use of it. That is what is done in France, for example: the copyrighter provide the torrent/ed2k link to the company (Trident Media Guard) in charge of collecting the IP addresses downloading the file. The origin of the link doesn't matter. Taking over a torrent website is useful to spot the original uploaders and dismantle the release groups. But this is useless to bust the people downloading.

You can remove the EZTV provider, but that doesn't make torrent downloading any safer. You can even consider that EZTV links will be posted on other websites such as KAT, TPB or rarbg. So once again, the origin of the link doesn't matter.

pauldcox:
>You can even consider that EZTV links will be posted on other websites such as KAT, TPB or rarbg.

No:

>A Pirate Bay moderator informs TF that they have suspended the EZTV user account. Many of the older torrents are still on the site, but TPB has added a warning urging people to stay away from the compromised domain.

>Other torrent sites such as KickassTorrents, BT-chat and Rarbg have also disabled or suspended the official EZTV accounts after hearing about the takeover.

>In addition, KickassTorrents and BT-chat have added the same warning as TPB. This way they hope to keep people away from the compromised EZTV site, which is now serving various ads including pop-unders.

>Former EZTV staffers also urge people to stay away from all EZTV sites and to inform others to do the same. The real EZTV is no longer active.

Popcorntime has even moved away from it and is now using ETTV. Why you are being so stubborn and insistent to ignore all the evidence is beyond me--and beyond reason for that matter. I don't use it in Sickrage so it doesn't affect me at all but providing a service to a "provider" that has been essentially universally shunned by the P2P community is reckless and irresponsible. My two cents, I'm done.

nicolasmartinelli:
@pauldcox 
> No

Well, yes, it is already the case. First link on EZTV: magnet:?xt=urn:btih:647a2fdec9f95a1665eb3cb5b7ac9f75cc9938d4&dn=...........

The hash is 647a2fdec9f95a1665eb3cb5b7ac9f75cc9938d4. Search the hash on KAT and you'll see it is available. It is also apparently a torrent originally posted on rarbg, while I cannot confirm since I cannot search the hash on rarbg.

Now, tell me: are you safer if you get this file through eztv or through kat?

PR #1911, opened 2015-05-16
FR#1585 - Add Fanart to local image cache

Thraxis:
* Adds FanArt to the local image cache for the web interface.
* Call to show fanart in templates would be:
"$sbRoot/showPoster/?show=$show.indexerid&amp;which=fanart"


miigotu:
Excellent. Thanks for this.

Goeny:
Great!
Will use that in the theme. :)

PR #1919, opened 2015-05-19
Fix name of attribute in providers.newsznab.NewznabCache._parseItem()

gosku:
When I was debugging the cache yesterday, I realised that providers.newsznab.NewznabCache._parseItem() was looking for an attribute named "tvrageid" when the  name that Newsznab returns is "rageid".

This error provoqued TVCache._addCacheEntry() never added an entry was it was called from NewznabCache._parseItem()

### DIFF
--- a/sickbeard/providers/newznab.py
+++ b/sickbeard/providers/newznab.py
@@ -423,7 +423,7 @@ class NewznabCache(tvcache.TVCache):
 
         tvrageid = 0
         for attr in item['newznab_attr'] if isinstance(item['newznab_attr'], list) else [item['newznab_attr']]:
-            if attr['name'] == 'tvrageid':
+            if attr['name'] == 'tvrageid' or attr['name'] == 'rageid':
                 tvrageid = int(attr['value'] or 0)
                 break
 


miigotu:
Is it possible tvrageid was used in an earlier version of newznab or different implementations? I'm not sure if this is the case or if it is simply a code error.
If this is dependant on newznab version, this might be more appropriate, to remain functional with older newznab installs:
`if attr['name'] == 'tvrageid' or attr['name'] == 'rageid':`

miigotu:
@gosku btw, not trying to be a pain about every pull xD, just desperately trying to fine tooth comb everything to prevent any regressions and avoid future complications.

Added and changed code is not reviewed and tested before merging quite as much as it should be, and rarely does anyone add any unit testing for their code.

gosku:
Don't worry, actually you're right, it's better to keep the "tvrageid" just in case. I will fix it.

fernandog:
@gosku as soon you test and it's ok. let us know

gosku:
fixed, now it is going as expected through _addCacheEntry()

BTW, How can I modify a PR? I did it by "push -f origin Newsznab-parseItem-fix" after rebase 2 commits on local, but I know "push -f" is sacrilege...

PR #1922, opened 2015-05-21
Small change for additional layout.

Goeny:
Gives possibility to add coverflow layout to homepage (For a theme i am working on)

### DIFF
--- a/sickbeard/webserve.py
+++ b/sickbeard/webserve.py
@@ -401,7 +401,7 @@ class WebRoot(WebHandler):
 
     def setHomeLayout(self, layout):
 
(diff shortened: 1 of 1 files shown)


miigotu:
@Goeny shouldnt this be added along with your theme rather than separate

Goeny:
The change is neccesary to for a new layout that comes with my theme.
Since it is a file that comes seperately from the theme folders, every time i update SR it is overwritten and my theme stops working.
I figured such a small change wouldn't make an impact in the SR code and merging wouldn't be a problem.
Making it much easier to update SR and continue working on the theme without changing the same line, in the same file after every update.

PR #1923, opened 2015-05-22
Prevent TraktRolling/TraktChecker queue to crash on trakt api error

gborri:


Thraxis:
This should really be handled in the Trakt.py (API) file

gborri:
@Thraxis it seems to be handled but it crash (seen from server status: alice-->red false)

miigotu:
Agreed, the log line would only need added once in `self.trakt_api.traktRequest` and should include the requested uri, instead of a generic `could not connect`.  Where it fails/hangs I bet is some invalid data getting added to the returned list when the connection fails or trakt returns something funky, which is then ran through the for loop which follows the request.  Imho, this cannot be merged as is.

miigotu:
Fix it in `trakt_api.traktRequest` and re-submit if you want this changed. Closing this.

PR #1924, opened 2015-05-22
Changed interval for show update process to 24h

gborri:


fernandog:
???

gborri:
now it runs every hour.... sickrage is countinuous updating show. I don't think is necessary to forceupdate every hour. 
Also because if you can set the hour of the day on which run "show update", i don't think is correct to make it run every hour

gborri:
@abeloin @fernandog do you thinks is correct to perform show update every hour?

miigotu:
@gborri It is needed more frequently afaik, in order to recognize when new episodes have aired shortly after they are added. This continuous updating of shows is not a full update, just the episode guide afaik.

gborri:
But why needed? You can check update on snatch so you are sure to have it
correctly when needed. It's. It full but if you have lot of show the
process take log time and sickrage continuously make only update.

Il giorno gio 28 mag 2015 01:07 Dustyn Gibson <notifications@github.com> ha
scritto:

miigotu:
If it doesnt update, how will it know there is a new episode to even search for a snatch?

fernandog:
this is exactly as the original sickbeard. won't change it

https://github.com/midgetspy/Sick-Beard/blob/development/sickbeard/__init__.py#L730

gborri:
Could it be configurable? So everyone one can decide the better solution
for his needs?

Il giorno gio 28 mag 2015 15:52 Fernando <notifications@github.com> ha
scritto:

fernandog:
change in your branch. only you wants to change it.
the showupdater runs every 24hours at 03

```
2015-05-28 03:00:41 INFO     SHOWUPDATER :: Completed full update on all shows
2015-05-28 03:00:41 INFO     SHOWUPDATER :: Trying to clean cache folder /home/osmc/SickRage/cache
2015-05-28 03:00:41 INFO     SHOWUPDATER :: Doing full update on all shows
2015-05-27 03:00:30 INFO     SHOWUPDATER :: Completed full update on all shows
2015-05-27 03:00:30 INFO     SHOWUPDATER :: Trying to clean cache folder /home/osmc/SickRage/cache
2015-05-27 03:00:30 INFO     SHOWUPDATER :: Doing full update on all shows
```

gborri:
@fernandog are you sure that the showupdater is running every 24h? The
interval is set to 1 hour and I see it's running every hour.

Il giorno gio 28 mag 2015 16:49 Fernando <notifications@github.com> ha
scritto:

miigotu:
@gborri Have you checked how the scheduler works? When start_time is specified cycleTime just makes sure it waits at least cycleTime before running subsequent cycles. IE: If you start SR at 2:30am it will still run at 3am even though cycleTime has not passed for the first run, but then it makes sure cycleTime has passed between 3am and 3am the next day before it runs on the next start_time.

```
                    should_run = False
                    #Is self.force enable
                    if self.force:
                        should_run = True
                    # check if interval has passed
                    elif current_time - self.lastRun >= self.cycleTime:
                        # check if wanting to start around certain time taking interval into account
                        if self.start_time:
                            hour_diff = current_time.time().hour - self.start_time.hour
                            if not hour_diff < 0 and hour_diff < self.cycleTime.seconds / 3600:
                                should_run = True
                            else:
                                # set lastRun to only check start_time after another cycleTime
                                self.lastRun = current_time
                        else:
                            should_run = True
```

miigotu:
image

gborri:
is it possible that if I configure 0 (midnight), it start to update every hour?

miigotu:
@gborri yes, I think it would update every hour if you set midnight.

gborri:
@miigotu probably the parameter description need to be updated, because otherwise is understandable.....

miigotu:
@gborri Actually, `self.start_time` should initially be `None`, and then the check should be `if self.start_time is not None:`

Also, the range could be checked there. I'm not sure if it is in hours there but if it is then it could be checked that self.start_time <= 24 and self.start_time >= 0.

PR #1942, opened 2015-06-10
Fix SiCKRAGETV/sickrage-issues#1554

miigotu:
Some proxies don't support https, fallback to http on failure, check for no-json or dirty json and handle it.

### DIFF
--- a/sickbeard/providers/rarbg.py
+++ b/sickbeard/providers/rarbg.py
@@ -83,7 +83,7 @@ class RarbgProvider(generic.TorrentProvider):
                         'token': '&token={token}',
         }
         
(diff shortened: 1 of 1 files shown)


fernandog:
@miigotu check this code. Isn't handled here the SSL?

https://github.com/SiCKRAGETV/SickRage/blob/develop/sickbeard/helpers.py#L1301

miigotu:
@fernandog no, it is a different issue. The proxy service itself returns an error when trying to load the page, about it not being able to curl it.

miigotu:
Without falling back to http, the proxy loads it's homepage with the error "The requested resource could not be loaded. libcurl returned the error:
SSL connect error"

PR #1944, opened 2015-06-11
Unranked rarbg and pp logfix

miigotu:


fernandog:
@miigotu cant we add ranked as an option like "trusted" in KAT?

miigotu:
We can, but trusted and ranked are completely different things. Default definitely needs to be 0 though, imo.

PR #1948, opened 2015-06-12
Allow SD as archive quality in custom qualities, SD is better than Un…

miigotu:
…known, and we sometimes want a smaller/lower-quality file archived to save space


miigotu:
Will squash later, need to make sure pp will allow overwriting higher quality.

PR #1952, opened 2015-06-13
Anime regex

missonnier-g:
Here is a patch on anime regexp, and another one to patch a bug that the first one reveal.


fernandog:
why remove replace_map['%RG'] == 'SiCKRAGE' ?

I think PP needes release group. Also the renamer

missonnier-g:
I don't remove it. I just do it correctly :
  1 : Because the code in IF and ELSIF was the same, so it can be factorized.
  2 : Because of boolean algebra. The contrapositive of the IF condition was (hasattr(self, '_release_group') OR replace_map['%RG'] == 'SiCKRAGE'), implying that self could have had no _release_group attribute when checking the ELSIF condition.

Just image the case where self had no _release_group attribute and replace_map['%RG'] was "Hello", and you ended in a nice exception in the ELSIF condition.

I don't understand the part of the condition that checks (not replace_map['%RG'] == 'SiCKRAGE'), but I don't remove code just because I don't understand it. Nevertheless, if some other components absolutely needs _release_group to be valued, it may be an error too ^^

miigotu:
Provide example anime release names where the original regex fails, and a list of anime release names I can test your fixes with. After I can verify it doesnt break the regex I can merge this.

missonnier-g:
Examples are in the code. The regex fails to parse simple examples given in the code ^^

# [Group Name] Show Name.13-14
# [Group Name] Show Name - 13-14
# Show Name 13-14
# [Group Name] Show Name.13
# [Group Name] Show Name - 13
# Show Name 13

You have to add extra_infos for the regex to match :

[Group Name] Show Name - 13 - [1080p]


Here are the logs for a file named [DB]_Bleach_101 with and without the patch.
What I don't understand is why, without the patch, the matching regex is "bare" ("anime_bare" should have matched before it falls back to "standards regexes")


Before :
2015-06-27 10:15:45 DEBUG    Thread-28 :: formatting pattern: %SN - %Sx%0E - %EN -> Bleach - 1x01 - 001 - Le jour où je suis devenu Shinigami
2015-06-27 10:15:45 DEBUG    Thread-28 :: Episode has no release name, replacing it with a generic one: %SN - %Sx%0E - %EN
2015-06-27 10:15:45 DEBUG    Thread-28 :: Parsed /home/data/tvshows/Bleach/[DB]_Bleach_101 into Bleach - S1E1 [ANIME VER: -1] [ABD: False] [ANIME: False] [whichReg: ['bare']]


After:
2015-06-27 10:34:37 DEBUG    Thread-22 :: formatting pattern: %SN - %Sx%0E - %EN -> Bleach - 5x10 - 101 - Le Bankai de Mayuri ! Sawatari, Le clash des démons
2015-06-27 10:34:37 DEBUG    Thread-22 :: Parsed [DB]_Bleach_101 into Bleach -  [ABS: [101]] [ANIME VER: 1] [GROUP: DB] [ABD: False] [ANIME: True] [whichReg: ['anime_standard']]
2015-06-27 10:34:37 DEBUG    Thread-22 :: Parsed /home/data/tvshows/Bleach/[DB]_Bleach_101 into Bleach -  [ABS: [101]] [ANIME VER: 1] [GROUP: DB] [ABD: False] [ANIME: True] [whichReg: ['anime_standard']]

miigotu:
We cannot allow parsing a filename with no quality information.

missonnier-g:
Sorry, my patch is pointless in this case. I will keep it in my fork because it is the only for me to use sickrage.

Do you need help to rewrite all other regexes? Because extra_info is optional everywhere!

miigotu:
quality or source is always required.

missonnier-g:
4 weeks to explain a 4 lines patch...
I definitely not known sickrage as well as you do, maybe you check elsewhere that the source quality is valued, but in my world an exclamation mark means optional, and approximately all your regex match without extra_info...

miigotu:
It is not optional, it is required that a release name have some information to go on, otherwise quality and source is unknown.

It must have something like SD, HDTV, DVD, 720, 1080, 480, PDTV, x264, SOMETHING.

If the filename or release name does not have this it is not a valid scene release, period.

missonnier-g:
Did you really check the code before you answer?

I don't dream, if you put a file named s1e1.mkv, the standard regex match, here is the logs :
2015-06-28 08:31:50 DEBUG    Thread-36 :: Parsed /home/data/tvshows/Hero Corp/s1e1 into S1E1 [ANIME VER: -1] [ABD: False] [ANIME: False] [whichReg: ['standard']]

Do whatever you want with the patch, I don't care if you like so much duplicated code, exceptions, and bad regex.

miigotu:
You aren't understanding. 
`Hero Corp/s1e1` is not a valid anime release name, so NO anime regex should match it.
`[DB]_Bleach_101` SHOULD match anime_bare since it has absolute numbering but no quality or source type.

So with this information we devise that standard regex matches it (as it should) as a non anime to get as much information from it as we can. We also devise that anime_bare is borked.

So from that information you change anime_standard? The comments above anime_standard are incorrect. the `_standard` part of the regex label means it is the standardized naming convention.

Without extra_info (quality OR source type) sickrage cannot add it regardless if the regex matches or not, because we explicitly refuse any search result that parses to Quality.UNKNOWN.

All of this will be obsoleted by the replacement of these regex with guessit regardless.

fernandog:
@miigotu did you try guessit with anime?

miigotu:
Nope, but I know it works with absolute numbering and group name before the title in []

PR #1956, opened 2015-06-14
Remove eztv and ezrss due to general scene skepticism about the shady…

miigotu:
… nature of their takeover


miigotu:
@fernandog I think this needs to be reviewed by a few people before I merge it, because of the change in search.py.  How does it look to you?

PR #1957, opened 2015-06-14
Try to get the best quality from the best/archived list on first snat…

miigotu:
…ch, but accept the highest from the any/allowed list otherwise

### DIFF
--- a/gui/slick/interfaces/default/inc_qualityChooser.tmpl
+++ b/gui/slick/interfaces/default/inc_qualityChooser.tmpl
@@ -21,11 +21,11 @@
 <div id="customQualityWrapper">
 	<div id="customQuality">
 	    <div class="component-group-desc">
-	        <p>One of the <b>Initial</b> quality selections must be obtained before SickRage will attempt to search and process the <b>Archive</b> selections.</p>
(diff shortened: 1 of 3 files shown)


fernandog:
@miigotu we shouldnt change the names as users are used to it

miigotu:
@fernandog "Initial" doesn't make any sense anymore. Before SR would Initially only DL a quality from the 'Initial' list for the first snatch.

Also, in the code, it has always been "Any" not initial, and archive is "best" in code.
The previous setup only worked in one direction and didn't really make any sense.

'Preferred' is much closer to what the second list actually is/does than 'archived'

I think that people being used to something is less important than that thing being easy to understand to begin with and making sense =P.

fernandog:
But we have Archived status. And a lot of screens have archived label.

miigotu:
Yes, archived status, not quality. The functionality is improved now, the information/label did not make sense anymore. It didn't even work the way it was supposed to before these changes. It all works now and it makes sense.

We can't leave labels with incorrect names/descriptions just because people are used to it, its just incorrect then. Incorrect causes more issues and confusion than people having to understand something new that is correct.

miigotu:
Its hard to explain, but before you could have an 'archived' quality, but it could download a different quality that was not preferred. When you archive something, you want it to be in your preferred list.  Archive status is totally separate from picking qualities though. It's like before someone confused two totally different features.

PR #1960, opened 2015-06-15
Add private or public label next to providers

fernandog:
**Need to adjust the box size**

@SiCKRAGETV/moderators suggestions?

providers

### DIFF
--- a/gui/slick/interfaces/default/config_providers.tmpl
+++ b/gui/slick/interfaces/default/config_providers.tmpl
@@ -93,7 +93,11 @@ var show_nzb_providers = #if $sickbeard.USE_NZBS then "true" else "false"#;
                           <li class="ui-state-default" id="$curName">
                             <input type="checkbox" id="enable_$curName" class="provider_enabler" #if $curProvider.isEnabled() then "checked=\"checked\"" else ""#/>
                             <a href="<%= anon_url(curProvider.url) %>" class="imgLink" rel="noreferrer" onclick="window.open(this.href, '_blank'); return false;"><img src="$sbRoot/images/providers/$curProvider.imageName()" alt="$curProvider.name" title="$curProvider.name" width="16" height="16" style="vertical-align:middle;"/></a>
-                            <span style="vertical-align:middle;">$curProvider.name</span>
+                            #if $hasattr($curProvider, 'api_key') or $hasattr($curProvider, 'password') or $hasattr($curProvider, 'passkey'):
+                            <span style="vertical-align:middle;">$curProvider.name (Private)</span>
+                            #else
+                            <span style="vertical-align:middle;">$curProvider.name (Public)</span>
+                            #end if
                             #if not $curProvider.supportsBacklog then "*" else ""#
                             #if $curProvider.name == "EZRSS" then "**" else ""#
                             <span class="ui-icon ui-icon-arrowthick-2-n-s pull-right" style="vertical-align:middle;"></span>


Arcanemagus:
Not a moderator, but I would say an icon denoting their status would be better.

MGaetan89:
I like the idea of the icon, but I think it would be too much icons in each row (one for the provider, one to sort the providers and one for public/private).

fernandog:
I tried to add color but isnt possible (AFAIK).

Adding icone needs a lot more coding and Im not expert.

Arcanemagus:
As it stands the text method is breaking the UI (See the KickAssTorrents line).

fernandog:
see my first comment in this PR

@Thraxis do you know a better way? I'm not an expert in html

Thraxis:
Could use a CSS to change the color of the box / text if it is public / private. Otherwise it's increase the size of the box and / or use an icon of some kind  (padlock?)

fernandog:
@Thraxis I couldnt increase size. Any hint?

Too complicated to add the padlock for private?
is it possible to order by public/private first then to name?

miigotu:
.change the color of the box. Matches the idea of the rest of the theme then like episode status'

Thraxis:
Something like this?
image

image

Thraxis:
It's not 100% accurate as not all private trackers have a 'username' field (You'll notice BTN and Rarbg for example)

fernandog:
Maybe instead of check username check for password. Some are just 'user' I think

fernandog:
Rarbg is public

Thraxis:
BTN just has API Key

miigotu:
Maybe you could add a field to the nzb and torrent classes in generic.py, like access_type = 'public', and set it to 'private' in the init of each private indexer/tracker

miigotu:
Then you just check against that in the template, and set the css class by it.

fernandog:
@Thraxis do you want to try or should I fix the box size and merge it ?

fernandog:
@Thraxis 
updated my code to handle passkey and api_key

Thraxis:
@fernandog The problem with sorting is the purpose of that screen is to allow the user to order the providers as to how they want them searched (my understanding )

I think we need to do like miigotu suggested.  There needs to be a property in the provider module to tell if it is public or private.

You can have public providers that use API Keys, etc.

miigotu:
Are enabled providers still moved to the top of the list in the user order?

fernandog:
@miigotu neved saw that behavior. it's does not auto move AFAIK. user must move top

miigotu:
I could have swore it used to, after you clicked save and refresh. That is easy to add btw. Keeps the user order, but all enabled are at the top.

fernandog:
@miigotu sorry. yes. only when you save. the code "str(int(x.isEnabled()))" order by enabled

miigotu:
Except I just checked, and it doesnt do it anymore xD.

fernandog:
not working. strange

fernandog:
@miigotu 

"add a field to the nzb and torrent classes in generic.py,"

how we will get this information from template file?

miigotu:
it would be like $curProvider.name except it would be $curProvider.isPrivate or something.

fernandog:
@miigotu 

here? https://github.com/SiCKRAGETV/SickRage/blob/develop/sickbeard/providers/rarbg.py#L66

miigotu:
There would work, but I think it would be cleaner to do it in providers/generic.py Line46 as an argument to __init__, and modify any calls to init in every provider and in class NZBProvider and class TorrentProvider. This way you get an error if you missed a provider or if someone adds a provider in the future which is missing the value.

```
def __init__(self, name, public):
    self.public = public
```

miigotu:
Then in the css there should be probably 4 new attributes, .publicNZB, .privateNZB, .publicTorrent, .privateTorrent, each with a unique color.

Then you set the class by providerType and public

fernandog:
@Thraxis are you still interested in doing this?? thanks!

fernandog:
@Thraxis please??

miigotu:
I think hes waiting on the public privat fields in the providers

fernandog:
But we will do two PRs for that? One for fields and one for template?

miigotu:
I want to make a feature branch here, add the private/public attribute to the providers myself, and Thraxis or someone else can PR to that feature branch after. That is how more than one person works on a feature and adds commits without PR each on develop (and having half done work in develop).

PR #1968, opened 2015-06-17
Remove "Default Episode Status" from editshow

fernandog:
because it's only usefull when adding a show

https://github.com/SiCKRAGETV/sickrage-issues/issues/1483

### DIFF
--- a/gui/slick/interfaces/default/editShow.tmpl
+++ b/gui/slick/interfaces/default/editShow.tmpl
@@ -63,14 +63,16 @@ This will <b>affect the episode show search</b> on nzb and torrent provider.<br 
 #include $os.path.join($sickbeard.PROG_DIR, "gui/slick/interfaces/default/inc_qualityChooser.tmpl")
 <br />
 
+<!--
 <b>Default Episode Status:</b><br />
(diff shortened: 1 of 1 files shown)


miigotu:
We can remove this for now, but it's good to leave the code as we can make it work later how it was originally intended if we want to.

Paused should not update the show at all, and you wont see if there are new episodes
New episode default status should update info/new episodes, but you could set them to ignored or skipped by default.

fernandog:
we need to add another option. the existing one is for already aired. we need on for unaired

fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/1842

adampetrovic:
Umm. I can't see a good reason for why this was removed. What is provided in the Add Show screen is for existing, already aired episodes. What you have removed is for every show airing after the show was added to Sickrage. This is not a viable alternative. Please add this option back.

miigotu:
@adampetrovic The option in editshow did not work to change default status of new episodes after the show is already added. It is commented out until we can code it to work there also.

During add a show, it is so that SR doesnt try to snatch old episodes we dont want. But once the show is added, all new episodes are automatically wanted. Pause the show for now to not get new aired episodes.

PR #1972, opened 2015-06-17
Add Trakt empty token error message

fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/1838

### DIFF
--- a/lib/trakt/trakt.py
+++ b/lib/trakt/trakt.py
@@ -74,9 +74,12 @@ class TraktAPI():
         if None == headers:
             headers = self.headers
         
-        if not None == sickbeard.TRAKT_ACCESS_TOKEN:
-            headers['Authorization'] = 'Bearer ' + sickbeard.TRAKT_ACCESS_TOKEN
-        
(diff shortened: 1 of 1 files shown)


fernandog:
@miigotu can I just "return" instead of raise an exception?

miigotu:
Actually you want to, but i'm not sure if that change fixes the problem.

traktException just passes

fernandog:
@miigotu pass and continues or pass as ignore all next code ?

fernandog:
done

miigotu:
Nuuuu, I meant you were right that you need to use a continue, as the exception is just set to be ignored in lib/trakt/exceptions.py

fernandog:
Why others exceptions dont have return?

miigotu:
Idk, someone else handles the trakt code xD

fernandog:
@miigotu I didnt added the return. Or its ok as it is?

miigotu:
The raise makes it return from there AFAIK.

miigotu:
Should be fine, otherwise we will know in a few hours lol.

PR #1973, opened 2015-06-18
Remove html tags from ui.notifications

fernandog:
@miigotu only these two lines in html tag in ui.notification

### DIFF
--- a/sickbeard/webserve.py
+++ b/sickbeard/webserve.py
@@ -1540,7 +1540,7 @@ class Home(WebRoot):
             if not sickbeard.traktRollingScheduler.action.updateWantedList(showObj.indexerid):
                 errors.append("Unable to force an update on wanted episode")
 
-        ui.notifications.message('<b>%s</b> has been %s' % (showObj.name,('resumed', 'paused')[showObj.paused]))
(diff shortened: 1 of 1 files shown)


fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/1832

miigotu:
Failed, also there are a lot more html tags in other notifications in webserve.py

fernandog:
But they are not UI notifications. UI that is the problem isnt?

miigotu:
They are ui notifications, but searching doesnt show some of them because they are split on 2 lines, like line 3372.

ui.notifications.error and ui.notifications.message, search for those two strings and look at each one.

miigotu:
`'<br >\n'.join(errors)`

miigotu:
Ok, so I just did some tests, and notifications can have a few select html tags in the message, but not the title.

Lists work <ul><li></li></ul>
Bold kinda works <b></b> makes *Text*

Images do not aparently, can you verify this works: https://github.com/fernandog/SickRage/blob/html_notification/sickbeard/webserve.py#L2023

fernandog:
I only removed bold. That on works then?

PR #1974, opened 2015-06-18
Allowing non-existing HTTP date header

trick77:
See https://github.com/SiCKRAGETV/sickrage-issues/issues/1720


miigotu:
We can't do this in cachecontroller, . Open an issue on https://github.com/ionrock/cachecontrol

Besides if you set current_age = 0 the cache will never be invalidated and never refresh with a new request.

trick77:
The bug has been fixed in cachecontrol 3 months ago, you need to pull a recent version.

miigotu:
I will update it tonight then, thanks.

PR #1981, opened 2015-06-21
Multiple fixes

miigotu:
Fix SiCKRAGETV/sickrage-issues#1862
Fix SiCKRAGETV/sickrage-issues#1825 (Hopefully)
Fix Fix SiCKRAGETV/sickrage-issues#1858 (Might not catch it still)
Strip '. -' after removeWords.

regexes need updated but are quite complicated. failing regex is the only reason we need removeWords, and even after they are not perfect. I have a perfect working regex without removing subrls suffix but atm it breaks down if there is no release group at all in the name.

### DIFF
--- a/gui/slick/interfaces/default/history.tmpl
+++ b/gui/slick/interfaces/default/history.tmpl
@@ -213,16 +213,10 @@
 					#for $action in sorted($hItem["actions"]):
 						#set $curStatus, $curQuality = $Quality.splitCompositeStatus(int($action["action"]))
 						#if $curStatus == DOWNLOADED:
-							#set $match = $re.search("\-(\w+)\.\w{3}\Z", $os.path.basename($action["resource"]))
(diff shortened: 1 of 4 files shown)


fernandog:
@miigotu why don't we remove the https from rarbg api url? there's no sensitive data

Hellowlol:
That a depends on who you ask. For some, what they download might be sensitive for them. I think sr (or any other program for that matter) should always use https aslong as that's an option.

miigotu:
@Hellowlol I agree. But for getting the token at least, http is enough because all that is returned is a token.

These modifications are in their early stage anyways, and I am working on making some mass provider changes atm, so a lot could change for the better soon.

rarbg:
@Hellowlol @miigotu 
If they want to sniff you they will sniff your https also. Its not that hard :)

PR #1989, opened 2015-06-22
If these are too high, they are being used in places they don't belon…

miigotu:
…g! Don't change them

This is likely to help with a lot of banning/throttling/blocking of SR. Thanks @thezoggy


thezoggy:
typo: https://github.com/SiCKRAGETV/SickRage/blob/develop/sickbeard/scene_numbering.py#L501
should be XEM not XEN.

PR #1992, opened 2015-06-23
Why use requests if we arent going to verify anyways? Maybe requests …

miigotu:
…is broken with verify off


miigotu:
@fernandog what do you think about this?  I'm hoping to get some people with ssl problems to test this.

fernandog:
@miigotu really don't know. let's try. Cant' remember why this was added

miigotu:
The only reason to move from urllib2 to requests is to be able to verify an ssl certificate on Py2.

Since we have requests, and all of the requirements, there is no reason to not verify.

Also, verify=False may expose some bug in requests that causes some of our issues, because 99.99% of people using requests leave verify alone.

PR #2014, opened 2015-06-27
Updated search stings with airdate

daarcm:
Updated search stings with airdate from "|" to " " to narrow searches

"|" is used as an and function, this searches all entries that include the month and date specified, replaced with a space instead to bring only relevant entries to create successful searches.

### DIFF
--- a/sickbeard/providers/iptorrents.py
+++ b/sickbeard/providers/iptorrents.py
@@ -116,12 +116,12 @@ class IPTorrentsProvider(generic.TorrentProvider):
         if self.show.air_by_date:
             for show_name in set(allPossibleShowNames(self.show)):
                 ep_string = sanitizeSceneName(show_name) + ' ' + \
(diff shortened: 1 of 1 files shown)


daarcm:
log before
https://drive.google.com/file/d/0B8UA2p6mIRyOcmZHaVkyTzNCaU0/view?usp=sharing
log after
https://drive.google.com/file/d/0B8UA2p6mIRyOckZGdHp0X2tOcWM/view?usp=sharing

miigotu:
only ipt? xD

daarcm:
I only use IPT, the search function may be different on other providers,  I assume it was originally created from a template,  and not optimized

miigotu:
@daarcm mostly someone poached it from some other fork I assume, and didn't verify the original code before adding it very well =P.

Anyways, thanks for this, saves me one step as I work my way through reworking the providers

PR #2017, opened 2015-06-28
Update subliminal

miigotu:
Add dogpile and pysrt depends for new subliminal

Required because of new guessit.


fernandog:
@abeloin was updating SR with new subliminal version. We are using a very old version. Dont know how much still needs to be done to upgrade.

miigotu:
@fernandog the problem is that all of the libraries have dropped python 2.6. Either we need to also drop python 2.6, or patch every library as errors pop up.

Hellowlol:
Any reason to keep 2.6 at all?

miigotu:
@Hellowlol Only if we have users who have no choice but to use Python 2.6
SiCKRAGETV/sickrage-issues#1915

abeloin:
@fernandog Here's the branch were I was working on subliminal: https://github.com/abeloin/SickRage/tree/feature-subliminal080. It maybe missing 1 or 2 commits I had on my PC(which is in storage until I move back in mid-July).

It is partially functional, some function have changed in subliminal requiring some code change in sickrage(not completed):
* https://github.com/abeloin/SickRage/commit/646f5aa27b0dcd8843679e9d5a84767b680156c7?w=1
* https://github.com/abeloin/SickRage/commit/1d4a1d8cfeba81f856196d1b441f855e0423cbc8?w=1

miigotu:
@abeloin Yes I saw that and have been partially referring to it. Locally I am down to figuring out how to enable/disable the providers in the new subliminal, and changing the actual download line. Also, babelfish.Language(x) doesnt like alpha2 codes, doesnt find it unless its alpha3 -.-

miigotu:
Wouldnt even have started on this is new guessit worked with old subliminal lol.

abeloin:
@miigotu In the end, I was thinking of changing the whole country stuff we use to detect the language(which is a utterly broken concept IMHO) and simply use the language list from the selector widget I used for the tvdb language chooser. This way we can use babelfish.Language and don't have to do hoops to make it work

miigotu:
@abeloin This all is a work in progress and should not have gotten merged so quickly, but I see many changes to this that will help.

Eventually, I planned on it being able to accept any language code type in ini/db. ei: ietf, alpha2, alpha3 etc. And I want to stop converting languages back and forth between codes and instances, there are lots of methods unexplored/not utilized by us in the new subliminal.

I also want to move more of the subtitle handling into subtitles.py and class it. Since the subliminal major update was a complete rewrite for now I just want to make it all work since we had to make so many changes anyways (very close now).

miigotu:
SR is doing a lot more work to handle subtitles than it needs to, I have cut alot out, but there is alot more optimizing that can be done.

abeloin:
> Eventually, I planned on it being able to accept any language code type in ini/db. ei: ietf, alpha2, alpha3 etc.

Nice that would be better.

I'm looking forward to that feature branch :)

PR #2032, opened 2015-06-30
Patch disable add paused if trakt rolling

gborri:


gborri:
I've tried to squash also this PR but have no success in doing it. There is a lot of mixed commit and the procedure i have adopted for the other doesn't work. Anyone can help?

miigotu:
Do you have a separate remote called `upstream` for the main repo?

miigotu:
By the way, shouldn't a user be able to add a show paused regardless of trakt rolling is enabled? trakt rolling should just not touch the show if it is paused.

gborri:
@miigotu you are right. but the point is: would you  like to add show from trakt watchlist in active or paused mode? if active sickrage starts to downloa, for each show, "number of episode" configured episode.

gborri:
This PR have just cosmetic change asked by @fernandog

gborri:
@miigotu @fernandog i don't know how but i got it. suqashed on #2036

PR #2033, opened 2015-07-01
Subliminal continue

miigotu:
Update enzyme, make subtitles not throw errors, mostly. Could not verify if a subtitle actually downloaded, but the thread runs.


fernandog:
@neoatomic can you test subtitles please?

neoatomic:
@fernandog Already did. https://github.com/SiCKRAGETV/sickrage-issues/issues/1913#issuecomment-117660183

PR #2035, opened 2015-07-01
Icon for podnapisi.net was missing in the new Subliminal.

neoatomic:
Icon for podnapisi.net was missing in the new Subliminal.


neoatomic:
@miigotu I just note there is an icon but not named correctly. (ant therefor not showing)  podnapisiweb.png needs to be podnapisi.png

PR #2039, opened 2015-07-01
Add more debug log message to showrefresh

fernandog:


miigotu:
@fernandog from now on, regardless of how the other log lines are in a file, I would like to be adding/changing them all to use c-style replacement tags. I also like to line break after the % and indent 8 extra spaces.
```
logger.log(u"%s: Found files: %s" %
        (self.indexerid, mediaFiles), logger.DEBUG)
```
This eliminates needing to wrap things in str() and int() because it does it implicitly, thus shortening the line up, and making the line easier to search for in the code.

Also, whenever we are referring to an episode, try to use it like this:
```
logger.log(u"%s: Processing S%02dE%02d" %
        (self.show.indexerid, self.season, self.episode), logger.DEBUG)
```

I had a branch going where I was going over all of the log lines and making them more uniform but I got off track and started fixing other things. I would really like to make the logging more concise. Like only log what we need, and when we do we should be as descriptive as possible.

PR #2000, opened 2015-06-23
Add provider BTDIGG

matigonkas:
Feedback/testing is welcome
(back info) This provider ,scrappes torrents from the swarm ,so it will return alot of torrents that aren' t on the "surface" torrent sites
Also, If this gets added it would be better to get an api key exclusive to sickrage, as the PR is still using the one from the original fork 

PS:2000th PR, wooot

PR #2045, opened 2015-07-04
Add provider BTDIGG

miigotu:
Rebase of https://github.com/SiCKRAGETV/SickRage/pull/2000


matigonkas:
@miigotu  ive noticed something now ,when daily search runs it trows an error 
```
Error while searching BTDigg, skipping: TypeError("'NoneType' object has no attribute '__getitem__'",)
```

miigotu:
lol. @matigonkas you used tabs for indents?

matigonkas:
yeah

miigotu:
Always use 4 spaces for indents in python.

Also, the RSS cache is not implemented in that file (should still work for backlog i think). Working on it now.

miigotu:
@matigonkas #2050

matigonkas:
Thanks for fixing it @miigotu ,sorry for it being such a piece of s*** it seems OK now .But on the bright side since i updated it i got already like 20 new downloads that it didnt found previously on any of the public provoviders and some of the private obra só its a big plus :) 
Also do i use tab or 4 spaces? (Didnt quite understod)

miigotu:
Use 4 spaces. Tabs have no place in source code.
Just as a note, rss/daily is useless with btdigg, it is only ok as as backlog provider.

PR #2046, opened 2015-07-04
Trim all trailing spaces from lines in python files, using 's/ &//g'

miigotu:
Trim all trailing spaces from lines in templates, using 's/ &//g'

### DIFF
--- a/SickBeard.py
+++ b/SickBeard.py
@@ -170,7 +170,7 @@ class SickRage(object):
 
         if not hasattr(sys, "setdefaultencoding"):
             reload(sys)
-  
+ 
         if sys.platform == 'win32':
             if sys.getwindowsversion()[0] >= 6 and sys.stdout.encoding == 'cp65001':
                 sickbeard.SYS_ENCODING = 'UTF-8'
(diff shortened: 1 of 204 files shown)


miigotu:
@abeloin You think it is ok do do this now while we have few open PR?  Less ppl that need to rebase the better, but this annoys me.

abeloin:
I wouldn't touch the lib folder(except for a few exceptions) as those are from upstream.

I think is is better to create this PR and commit it right after a merge develop->master, it should resolve the rebasing issue

miigotu:
Hmm, was hoping it would be ok to just trim trailing on libs, as you can diff with ignoring whitespace.  I already didn't apply the pep8 to lib because that would totally fuxor diffs.

I will hold off on these until after a develop->master  merge like you suggested.

abeloin:
Regarding the libs, I gave you my personal opinion it doesn't mean it's set in stone. We are the advantages and disadvantages of doing that?

miigotu:
Not really a whole lot of advantage in trim trailing on libs besides making the source smaller a bit, but makes an extra step when updating/diffing libs and makes it harder to follow history before/after via github website ^. I will revert libs and squash before this gets merged.

PR #2047, opened 2015-07-04
autopep8 --in-place --max-line-length 160 --aggressive --aggressive

miigotu:
@abeloin this one too, touches everything but helps prepare for Py3 readyness. Should do this and the trim-trailing at same time, no?


abeloin:
See my comment in the other PR: https://github.com/SiCKRAGETV/SickRage/pull/2046#issuecomment-118471923

PR #2058, opened 2015-07-07
Use a temp hard link to get better subs matches

miigotu:
@fernandog @neoatomic 
This is what I mean about using a link named as release name (with extension), scan the video, then remove the link.

Then later name the subs right.

This works beautifully btw. It got every subtitle I had configured super quick, doesnt error or leave junk, and it has nearly every piece of imformation it could get. I suggest we use this method for now, and ask subliminal for a new paramter.

### DIFF
--- a/sickbeard/tv.py
+++ b/sickbeard/tv.py
@@ -1448,38 +1448,47 @@ class TVEpisode(object):
                 languages.add(babelfish.Language.fromietf(language))
 
             if not languages:
-                logger.log(u"%s: No missing subtitles for S%02dE%02d" % (self.show.indexerid, self.season, self.episode), logger.DEBUG)
+                logger.log(u'%s: No missing subtitles for S%02dE%02d' % (self.show.indexerid, self.season, self.episode), logger.DEBUG)
                 return
 
             providers = sickbeard.subtitles.getEnabledServiceList()
-            videos = subliminal.scan_videos([self.location], subtitles=not force, embedded_subtitles=not sickbeard.EMBEDDED_SUBTITLES_ALL or not force)
-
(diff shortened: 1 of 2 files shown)


sickbeard/tv.py
```diff
@@ -1448,38 +1448,47 @@ def downloadSubtitles(self, force=False):
...
+                        tmp += os.path.splitext(self.location)[-1]
+
+                    create_link = not ek.ek(os.path.exists, tmp)
+                    if create_link:
+                        ek.ek(helpers.link, self.location, tmp)
+                    vname = tmp
+                except:
+                    create_link = False
+                    vname = self.location
+                    pass
```
miigotu: @abeloin I have it ignore the hard link if it fails, and it just falls back to using the existing file for the scan. This should still work for unraid then right?

abeloin: Yes as it send an OSError: (38, 'Function not implemented').

Also you should use `except Exception:` instead of `except:` as you don't want to silence the KeyboardInterrupt(Ctrl+C to stop the program).

miigotu:
@abeloin changed the exception as you suggested. See anything else that might be off?

miigotu:
Should I be using ek.ek for os.path.split and os.path.splitext? Wasnt sure if that would handle a tuple well.

abeloin:
It should handle it well, I would change them to ensure we don't get weird encoding issue.

miigotu:
@abeloin I just figured we only needed to ek.ek for anything using stat in the background (because stat is funky with filename encoding).  Anyways, done.

abeloin:
Ok good to know

PR #2066, opened 2015-07-08
Fix TypeError on tntvillage provider

gborri:
Error log:
2015-07-08 10:50:37 ERROR    SEARCHQUEUE-BACKLOG-248835 :: [TNTVillage] :: Failed parsing TNTVillage Traceback: Traceback (most recent call last):
  File "/home/mediacenter/src/sickbeard/sickrage/sickrage/sickbeard/providers/tntvillage.py", line 436, in _doSearch
    if not self._is_italian(result) and not self.subtitle:
  File "/home/mediacenter/src/sickbeard/sickrage/sickrage/sickbeard/providers/tntvillage.py", line 309, in _is_italian
    if not re.search(sub, name, re.I):
  File "/usr/lib/python2.7/re.py", line 142, in search
    return _compile(pattern, flags).search(string)
TypeError: expected string or buffer
Traceback (most recent call last):
  File "/home/mediacenter/src/sickbeard/sickrage/sickrage/sickbeard/providers/tntvillage.py", line 436, in _doSearch
    if not self._is_italian(result) and not self.subtitle:
  File "/home/mediacenter/src/sickbeard/sickrage/sickrage/sickbeard/providers/tntvillage.py", line 309, in _is_italian
    if not re.search(sub, name, re.I):
  File "/usr/lib/python2.7/re.py", line 142, in search
    return _compile(pattern, flags).search(string)
TypeError: expected string or buffer

### DIFF
--- a/sickbeard/providers/tntvillage.py
+++ b/sickbeard/providers/tntvillage.py
@@ -302,6 +302,9 @@ class TNTVillageProvider(generic.TorrentProvider):
 
         name = str(span_tag)
 
+        if not name:
+            return 0
+
         subFound=0
 
         for sub in self.sub_string:


miigotu:
Checking for specific languages in providers is ridiculous btw, thats what required words are for.

Nevertheless, this PR doesnt fix that error.

miigotu:
TypeError is from when span_tag is None.

undo the changes you made in the last commit, then on line 304 do:
```
if not name:
    return 0
```

gborri:
@miigotu the problem is that on tntvillage there are no reference on title regarding language, so required words is not working. I agree with you that is something that is not related to provider but if every provider has different language management we need to make a language management and define for every provider the language adopted for every torrent with the particularly info managed by each torrent. I think is a big rewrite.

gborri:
closed for error.

gborri:
@miigotu done as per your suggestion

miigotu:
k

miigotu:
btw, wouldnt just re.search(" ita " work there? I saw on the ise it seems they are all spaced " eng ita "

gborri:
@miigotu they are spaces by the search procedure find also "ita", i think is not necessary to make it " ita " with spaces.

miigotu:
Just saying, you wouldn't have to search more than once, and wouldn't need self.sub_string

gborri:
@miigotu i need self.sub_string because the word referencing subtitle is more than once (sub, softsub)
the ITA is referenced either for subtitle or for audio language. So i have to split it in order to catch the self.sub_string and then looking for ita as audio language.

PR #2076, opened 2015-07-09
Try unicode bare first in case of empty string, then try chardet

miigotu:
SiCKRAGETV/sickrage-issues#1992
SiCKRAGETV/sickrage-issues#1830

@fernandog I want to check if there are any other places bad encode/decode happens, but this should catch most of it.


fernandog:
with have this one:
https://github.com/SiCKRAGETV/sickrage-issues/issues/937

miigotu:
This should be fixed now @fernandog

miigotu:
With chardet, now it will decode the string to unicode with the detected encoding (with many fallbacks). On encode, it will try a few different encodes before trying with 'replace' and then with 'ignore' errors.

PR #2078, opened 2015-07-09
Fix default episode status. Shows are added with set initial default …

miigotu:
…status,

then the status is set to wanted for future episodes. User can then edit show
or mass edit, and change default status for future episodes, and it will be
honored by dailysearcher.
SiCKRAGETV/sickrage-issues#2000
SiCKRAGETV/sickrage-issues#1925
SiCKRAGETV/sickrage-issues#1483
SiCKRAGETV/sickrage-issues#1875

### DIFF
--- a/gui/slick/interfaces/default/config_notifications.tmpl
+++ b/gui/slick/interfaces/default/config_notifications.tmpl
@@ -39,7 +39,7 @@
                     </div>
                     <fieldset class="component-group-list">
                         <div class="field-pair">
-                            <label class="cleafix" for="use_kodi">
+                            <label class="clearfix" for="use_kodi">
                                 <span class="component-title">Enable</span>
                                 <span class="component-desc">
                                     <input type="checkbox" class="enabler" name="use_kodi" id="use_kodi" #if $sickbeard.USE_KODI then "checked=\"checked\"" else ""# />
--- a/gui/slick/interfaces/default/editShow.tmpl
+++ b/gui/slick/interfaces/default/editShow.tmpl
@@ -63,16 +63,14 @@ This will <b>affect the episode show search</b> on nzb and torrent provider.<br 
 #include $os.path.join($sickbeard.PROG_DIR, "gui/slick/interfaces/default/inc_qualityChooser.tmpl")
 <br />
 
-<!--
 <b>Default Episode Status:</b><br />
-(this will set a default status of already aired episodes)<br />
+(this will set the status for future episodes)<br />
 <select name="defaultEpStatus" id="defaultEpStatusSelect" class="form-control form-control-inline input-sm">
     #for $curStatus in [$WANTED, $SKIPPED, $ARCHIVED, $IGNORED]:
     <option value="$curStatus" #if $curStatus == $show.default_ep_status then 'selected="selected"' else ''#>$statusStrings[$curStatus]</option>
     #end for
 </select><br />
 <br />
--->
 
 <b>Info Language:</b><br />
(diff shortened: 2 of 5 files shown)


fernandog:
@miigotu should we add another field to "add show" so users don't get confused?

one for already aired (this one is temp as it would only be used to created ep objects)
another for unaired. (this is the setting itself default_ep_stauts)

https://sickrage.tv/forums/forum/help-support/general-support/25028-marking-episodes-skipped

miigotu:
no, because that would only confuse it further. After the initial wave of questions about it are over ti wont be such a hassle

fernandog:
What if I want everything SKIPPED?

miigotu:
You didn't have that option before, and to do that would require a db change.  As it is now you can just change it right after the show is added.

fernandog:
why not set another list box setting this?


        # After initial add, set back to WANTED.
        self.show.default_ep_status = WANTED

to

        # After initial add, set back to WANTED.
        self.show.default_ep_status = NEW LIST BOX FROM GUI

https://github.com/SiCKRAGETV/SickRage/pull/2078/files#diff-8e48fe44d6097ae0b4a7ec42fc45cda2R424

miigotu:
And where are you going to store that value up until that point?

miigotu:
Trust me, I wanted to do it that way, but after I changed a hell of a lot of code trying to get that value there I realized it required a database entry in order to do that.

fernandog:
Store in the db setting that already exist.
The temp value is for already aired. We only need to store for unaired. Or am I wrong?

miigotu:
It would have to be stored for both settings, because of threading.

miigotu:
None of this confusion is about how SR adds new shows, it is about shows that were added before the change which do not have wanted state now because SR did not change not to wanted automatically before.

miigotu:
Actually it might be able to be done, I just had a hell of a time getting it to work. I'll look at it again.

fernandog:
No rush on this.

itsrumsey:
I recently rebuilt a server and switched to unRAID and something about this feature has caused me an enormous headache.  When I added all my shows I set "Status for previously aired episodes" to skipped.  I realized now this also set the "Status for all future episodes" to skipped.  That option wasn't even in the add show menu on the version of SR that was out that day.

So now all of my shows are set to skip, there is no option to change that setting in mass update.  Is there any way to mass change all my shows "Status for all future episodes" to wanted, or do I have to remove every show and readd them all again?

miigotu:
Default status in mass update __IS__ the default status for future episodes. After the show is added the default status for past episodes is not even stored anywhere because it has no use once the show is added.

nevafuse:
I thought "pausing" a show was how people were supposed to "skip" recently added shows.  Took me a while to realize this pull request is why a bunch of shows weren't downloading recently.  The better option would be to remove the "default ep status" from the UI since it was only meant to be used when adding a show to SR/B.

miigotu:
@nevafuse I could not change the default episode status of existing shows in people's database, because they may have wanted future episodes skipped. Also, If I made the code auto set their unpaused shows' default episode status to wanted, how could they manually set it to skipped? next reboot would set it back to wanted.

The behavior of this change is to fix how it was ORIGINALLY supposed to work, before there ever was an option to pause shows.

Regardless, it works as intended. There will be a short period of transition where users need to read the news, forum, commit log, pull requests, wiki, or IRC to find out what's up. (the information is EVERYWHERE)

PR #2084, opened 2015-07-10
Scene exceptions list overwite results in multiple identical search url's

SerialShadow:
All items in the list referred to the same object, so when this object was mutated all the items in the list got mutated.
This resulted in exactly the same search url multiple times when we had a list of scene exceptions.
Appended a copy of the original params dictionary object to the list, instead os the original object.

### DIFF
--- a/sickbeard/providers/newznab.py
+++ b/sickbeard/providers/newznab.py
@@ -198,7 +198,7 @@ class NewznabProvider(generic.NZBProvider):
             if add_string:
                 params['q'] += ' ' + add_string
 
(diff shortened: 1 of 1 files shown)


miigotu:
@SerialShadow I assume this is tested well and ready to merge? It seems trivial enough.

SerialShadow:
@miigotu I gave it a couple of testruns and stepped through it with the debugger. Also the debug log file shows the right search url's after the modification. So I think it is ready to merge.

PR #2089, opened 2015-07-12
Better integration in Chrome for Android

MGaetan89:
This PR makes SickRage looks a bit better in Chrome for Android (don't know if it works on iOS too).

Before:
Before

After:
After (dark theme)

After (light theme)

### DIFF
--- a/gui/slick/interfaces/default/inc_top.tmpl
+++ b/gui/slick/interfaces/default/inc_top.tmpl
@@ -7,6 +7,13 @@
 		<meta charset="utf-8">
 		<meta http-equiv="X-UA-Compatible" content="IE=edge">
 		<meta name="viewport" content="width=device-width, initial-scale=1">
+
+		<!-- These values come from css/dark.css and css/light.css -->
+		#if $sickbeard.THEME_NAME == "dark":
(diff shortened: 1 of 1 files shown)


fernandog:
Menu will be always visible even if you resize?

MGaetan89:
This behavior is still the same as in the current version. So if you resize the page, you will have to scroll to access the menu. I didn't change anything about that.

However, I created an ticket about the initial scale issue: SiCKRAGETV/sickrage-issues#2079

fernandog:
Is it possible to put menu in the left?

MGaetan89:
You mean on the right? Or move the menu to the left?

fernandog:
Left. Everytime I zoom in. I need to scroll right to access the menu

MGaetan89:
If we move it to the left, we still have to scroll if we are at the right side of the page.
The best solution would be make the navigation bar not-zoomable, but I don't know if it is feasible.

fernandog:
Dont know if is but wouldbe nice!

MGaetan89:
I will have a look to see what can be done and give a feedback in SiCKRAGETV/sickrage-issues#2079.

PR #2090, opened 2015-07-12
Add an API method to get a network logo

MGaetan89:
For SiCKRAGETV/sickrage-issues#1763.

This adds a new API method to get the network logo of a show. I also updated the web part.
I think it would be nice to merge some code between the API and the web part. Right now the code to get a show banner/poster/network logo is duplicated between those two parts.

### DIFF
--- a/sickbeard/webapi.py
+++ b/sickbeard/webapi.py
@@ -74,7 +74,7 @@ result_type_map = {RESULT_SUCCESS: "success",
 
 class ApiHandler(RequestHandler):
     """ api class that returns json results """
-    version = 5  # use an int since float-point is unpredictible
+    version = 5  # use an int since float-point is unpredictable
     intent = 4
 
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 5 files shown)


sickbeard/webapi.py
```diff
@@ -254,6 +254,8 @@ def showPoster(self, show=None, which=None):
         # Redirect initial poster/banner thumb to default images
         if which[0:6] == 'poster':
             default_image_name = 'poster.png'
+        elif which[0:6] == 'fanart':
```
MGaetan89: I added this from `webserve.py`. I'm not sure if it is require in the API or not.

miigotu:
I'll have to test this one before I merge it.
About the duplicate code, yes it annoys me too, but before we go making big changes like that would require, we need to get to a more stable point and work some other stuff down. The same issue exists in the providers, which is higher priority than webserve vs api dupe code.

miigotu:
All looks good.

PR #2091, opened 2015-07-12
Small visual improvement in server status

MGaetan89:
This change the formatting of the the *Start time*, *Cycle time* and *Next run* columns in the *Scheduler* table.

Before:
Before

After:
After


MGaetan89:
It seems that my changes in `helpers.py` were not merged in `master` with the changes in `status.tmpl`.
Is it on purpose?

PR #2094, opened 2015-07-13
Change log messages to format

fernandog:
@miigotu rarbg is ok but still is showing bad chars:

[Rarbg] :: Unable to connect to Rarbg provider: error ('Connection aborted.', gaierror(-2, 'Nome ou servi\xc3\xa7o desconhecido'))
https://github.com/SiCKRAGETV/SickRage/blob/master/sickbeard/providers/rarbg.py#L114

this pull was made because of rarbg but it code already uses .format

### DIFF
--- a/sickbeard/clients/generic.py
+++ b/sickbeard/clients/generic.py
@@ -52,20 +52,19 @@ class GenericClient(object):
             self.response = self.session.__getattribute__(method)(self.url, params=params, data=data, files=files,
                                                                   timeout=120, verify=False)
         except requests.exceptions.ConnectionError, e:
-            logger.log(self.name + u': Unable to connect ' + str(e), logger.ERROR)
+            logger.log(u'{0}: Unable to connect {1} '.format(self.name,str(e)), logger.ERROR)
             return False
         except (requests.exceptions.MissingSchema, requests.exceptions.InvalidURL):
             logger.log(self.name + u': Invalid Host', logger.ERROR)
             return False
         except requests.exceptions.HTTPError, e:
-            logger.log(self.name + u': Invalid HTTP Request ' + str(e), logger.ERROR)
+            logger.log(u'{0}: Invalid HTTP Request {1} '.format(self.name,str(e)), logger.ERROR)
             return False
         except requests.exceptions.Timeout, e:
-            logger.log(self.name + u': Connection Timeout ' + str(e), logger.WARNING)
+            logger.log(u'{0}: Connection Timeout {1}'.format(self.name,str(e)), logger.ERROR)
             return False
         except Exception, e:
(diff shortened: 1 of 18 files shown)


miigotu:
just import ex and do ex(e) instead of str(e), that should fix it

fernandog:
But rarbg already does that and its one of the providers that has special chars in error message


[Rarbg] :: Unable to connect to Rarbg provider: error ('Connection aborted.', gaierror(-2, 'Nome ou servi\xc3\xa7o desconhecido'))

miigotu:
```python
                logger.log(u'{name} provider: No token found'.format(name=self.name), logger.ERROR)
                logger.log(u'{name} provider: No token found: {error}'.format(name=self.name, error=ex(e)), logger.DEBUG)
```
You can't format and ex(), just ex()

miigotu:
Seems ex is messed up atm.

miigotu:
@fernandog before merging this, check for the encoding problem with just ex() again with my new commits. ex/ss/ek was incorrectly detecting your encoding for serviço as TIS-620. My changes before updating chardet fixed that because I try them in a better order, and now also the updated chardet itself correctly guesses it as utf-8 rather than TIS-620.

miigotu:
Old chardet:
```python
>>> chardet.detect('serviço')
{'confidence': 0.99, 'encoding': 'TIS-620'}
```
New chardet:
```python
>>> chardet.detect('serviço')
{'confidence': 0.505, 'encoding': 'utf-8'}
```

fernandog:
why not get sys encoding first? mine is utf-8. after that you can try guess it

miigotu:
That's what it does now. Order = sys_encoding , utf-8, latin1, chardet.

miigotu:
Don't need to do this now right? Can you check? I rather fix the encoding than band-aid the log messages =P

fernandog:
will close it for now then

fernandog:
@miigotu still getting unicode chars:

```
Connection error to getURL http://thexem.de/map/all?id=260485&origin=tvdb&destination=scene. Error: ('Connection aborted.', error(104, 'Conex\xc3\xa3o fechada pela outra ponta'))
```

miigotu:
@fernandog because you still have it using .format.
logger.log(u"Connection error to getURL {0}. Error: {1}".format(url,e.message), logger.WARNING)

Also, e.message is deprecated since python 2.5.

logger.log(u"Connection error to getURL %s. Error: %s" % (url, ex(e)), logger.WARNING)

fernandog:
ok! thanks! a lot of changes =)

miigotu:
Also, anywhere you use str(somevar) in a log line is a potential unicode problem. you have to use ek.ek, ex, or ek.ss

PR #2102, opened 2015-07-14
Re-add namecache creation to SickBeard.py

fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/2097

### DIFF
--- a/SickBeard.py
+++ b/SickBeard.py
@@ -367,6 +367,9 @@ class SickRage(object):
 
         # Fire up all our threads
         sickbeard.start()
+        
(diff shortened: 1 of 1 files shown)


miigotu:
+1 Didn't realize that would cause issues, it is supposed to be checked on searches. I had removed it when debugging the scene exceptions resource hogging.

fernandog:
no problem!
without this, episodes would only get snatched in backlog not in dailysearch

miigotu:
This might be the post process problem also.

PR #2103, opened 2015-07-15
Use SxxExx pattern in log (2)

fernandog:
@miigotu sorry about zfill. This was done yesterday. I just missed the PR because I want to make sure travis build wont fail.

### DIFF
--- a/sickbeard/tv.py
+++ b/sickbeard/tv.py
@@ -263,8 +263,8 @@ class TVShow(object):
                 episode = int(sqlResults[0]["episode"])
                 season = int(sqlResults[0]["season"])
                 logger.log(
-                    "Found episode by absolute_number:" + str(absolute_number) + " which is " + str(season) + "x" + str(
-                        episode), logger.DEBUG)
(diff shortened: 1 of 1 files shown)


miigotu:
Still switching to zfill? I'm confused

fernandog:
i did this yesterday together with the other. but I had to create an PR to make sure it builds ok. the other one I did in UI

miigotu:
I see, just making sure you know zfill is bad. too much extra processing for each log line (and we have a lot of debug logs) it can impact performance. With my way you don;t even need to cast from into to str or str to int.

`"Found episode by absolute_number: %s which is S%02dE%02d" % (absolute_number, season, episode), logger.DEBUG)`

Is so much simpler than:
`"Found episode by absolute_number:" + str(absolute_number) + " which is " + "S" +  str(season).zfill(2) + "E" + str(episode).zfill(2), logger.DEBUG)`

fernandog:
wont do it know. need to redo all logs messages. zfill was easier to change. closing this then. will do in a couple of days

PR #2106, opened 2015-07-15
Add FrenchTorrentDB and Libertalia providers

adaur:
French private trackers


maxcore:
Work great

miigotu:
I was following you writing this, and at first glance it all looks good except one thing.

It needs to use requests instead of urllib2. This provides automatic session management and cookie handling as well as ssl verifications and SNI (when possible)

miigotu:
I'm going to mark it DNM until we can work through that.

maxcore:
@adaur @raver2046  you can do this ?

miigotu:
Matter of fact, I will be rewriting every provider to follow a standard format soon, so I need these to reference. No such a big deal that they use urllib2 for now.

adaur:
Thanks for the merge! I can rewrite those if need be.

miigotu:
@adaur If you want to make the changes to make them use self.getURL and requests then go ahead and PR. 

After I make the basic template for the new providers, you could also help there if you'd like. They will be designed to be movie/music/season/episode ready.

adaur:
Why not! I'll try to do this :)

I remember that I wanted to use getURL instead of self.opener but I couldn't use it within _doLogin since getURL checks if we are logged or not.

PR #2109, opened 2015-07-16
Fix old subtitles codes in DB

miigotu:


miigotu:
@fernandog RAWR! I can just change the time stamp next time I need to change subliminal codes!

PR #2113, opened 2015-07-16
Add Cpasbien provider

kylliox:
French public provider


miigotu:
This was added in another PR.

PR #2116, opened 2015-07-16
Fix detection of already downloaded subtitles when using a custom sub…

miigotu:
…title dir


miigotu:
@Rouzax no need to open an issue, this fixes the subtitles dir and detection of already downloaded subs.

PR #2117, opened 2015-07-17
Fix news url due to blacklisting

skmagiik:
following the rawgit instructions fixed the url to prevent blacklisting.


miigotu:
Negative. CDN version is cached forever, there would never be any update to the page yet. For now I have made it point to the news.md in source until I figure it out.

PR #2122, opened 2015-07-18
Set episodes with airdate as 'never' to UNAIRED status

fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/2184

@miigotu agree?

### DIFF
--- a/sickbeard/tv.py
+++ b/sickbeard/tv.py
@@ -1787,8 +1787,8 @@ class TVEpisode(object):
         if not ek.ek(os.path.isfile, self.location):
 
             # if it hasn't aired yet set the status to UNAIRED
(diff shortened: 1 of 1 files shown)


fernandog:
how something is SKIPPED is was never aired? Code didn't check for aidate = 1, only airdate > today

miigotu:
== not = lol, maybe this will be ok but i'm not sure.

fernandog:
damn! will fix it

miigotu:
Don't worry about silly syntax errors, I mis `:` after method definitions, ifs, and for statements all the time because of C/C++ lol

miigotu:
Btw, are you sure the self.airdate of airdate 'never' is 1 and not `datetime.date.fromordinal(1)` ?

fernandog:
DB value is 1

PR #2128, opened 2015-07-18
Fix search indexers api

MGaetan89:
Currently, if you use the `sb.searchtvdb` and `sb.searchtvrage` API commands, they both returns the search results for all indexers (TVDb and TVRage).

With this PR, the behavior is as follow:
- `sb.searchindexers` returns the results for all indexers
- `sb.searchtvdb` returns the results for TVDb only
- `sb.searchtvrage` returns the results for TVRage only


miigotu:
Sweet, I noticed that when checking the api results, but didn't investigate. Good find.

PR #2129, opened 2015-07-18
Add TitansOfTV provider

adaur:
Original file from SickGear


fernandog:
@adaur did you checked with them if is ok to use their code?

JackDandy:
No he didn't @fernandog, our position hasn't changed, this chap has done a blatant kang.

adaur:
I didn't, because SickGear is released under the GNU Licence and because they are given credits in the file. What is a blatant kang?

adaur:
OK, from what I've understood, "kang-ing" means stealing code without giving proper credits. This is not the case here since SickGear is credited:

# Originally written for SickGear

Besides, the GNU License gives all freedom to fork as long as the original author is mentionned.

http://stackoverflow.com/questions/4475319/how-to-rebrand-copyright-a-forked-project-gnu-gpl

I can change the credit line if you're not happy with it.

adaur:
Nevermind, I'll rewrite it myself.

fernandog:
Already closed. Dont delete so others can read when want to copy your code... 

Also SR donations has new home now =)

miigotu:
I 100% disagree. I have been in the open source community for a long time and this is always something people try and do.

1. Nothing is direct copy and paste from SG to SR
2. Its open source, they have no rights to deny usage of the code. If you don't want snippets of code from SG in SR, remove all SR/SB original code and make it closed source.

As long as proper credits and licensing is included with the PR, I am going to merge it every time.

neoatomic:
Actually i wanted to stay out of this discussion, but i need to voice my support for miigotu's view.
Also i find the tone that is used not really appropriate, its just short from giving orders... Think that can be done in a more civilized way.?

miigotu:
This happens often in the open-source world when there are contributors who do not believe in the core values of GPL. The GPL partially addresses specifically this, that when you fork or use code it cannot be for profit and that the new code must be the same license. It is also partially intentional to discourage forks and encourage the contribution directly back.

Open source projects are not to make money, and this is not like a guild sharing loot from a raid either. Donations are used for the betterment and maintenance of the project.

The only thing I've ever noticed people porting back are providers, which require modification to even work in SR properly, and which I will be rewriting when I soon rewrite all providers for the new format.

I have noticed commits in SG with changes I have made as well, and I don't give a shit. There is never any credit there, so at least we do the proper thing of giving credit.

I personally won't be seeking out code from them to port back because it's a waste of my time, but I also won't be seeking out whether every PR is ported from them when it is all fair game to begin with.

miigotu:
BTW, for future reference, our policy to contributors is that it's not required that they to ask some other project for permission, they simply need to leave the license intact with the original creators info (although they may add themselves to the license as modifying it).

Crediting in the commit/PR is also not required/necessary (but they can if they wish), the license handles that.

If the license allows it, it is fair game.

PR #2135, opened 2015-07-19
Simplify status in tv.py

miigotu:
@fernandog There is no reason to ever ignore an episode with airdate "never" anymore.

People usually had set these from skipped to ignored because they used to show up as missing episodes when they were set to skipped. They no longer get auto set to skipped, and unaired is already not included in missing episodes.

### DIFF
--- a/sickbeard/tv.py
+++ b/sickbeard/tv.py
@@ -1780,44 +1780,19 @@ class TVEpisode(object):
             return
 
         if self.location:
-            logger.log(str(self.show.indexerid) + u": Setting status for " + str(season) + "x" + str(
-                episode) + " based on status " + str(self.status) + " and existence of " + self.location, logger.DEBUG)
+            logger.log(u"%s: Setting status for S%02dE%02d based on status %s and location %s" %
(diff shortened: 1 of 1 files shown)


miigotu:
@fernandog tell me this doesnt look so much easier to follow?

```python
        # if we don't have the file
        if not ek.ek(os.path.isfile, self.location):

            # if it hasn't aired yet set the status to UNAIRED
            if self.airdate >= datetime.date.today() or self.airdate == datetime.date.fromordinal(1):
                logger.log(u"Episode airs in the future or has no airdate, marking it UNAIRED", logger.DEBUG)
                self.status = UNAIRED
            # if we don't have the file, but it has already aired
            else:
                # we arent changing skipped or ignored here
                if self.status == UNAIRED or self.status == UNKNOWN:
                    if self.season > 0:
                        self.status = self.show.default_ep_status
                    else:
                        self.status = SKIPPED
                else:
                    logger.log(
                        u"Not touching status because we have no ep file, the airdate is in the past, and the status is " + str(
                            self.status), logger.DEBUG)
```

PR #2138, opened 2015-07-19
Fix error: Filesize would require ZIP64 extensions

fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/2210

@miigotu you asked user to do this in forum
https://sickrage.tv/forums/forum/help-support/general-support/24316-updating-zip-creation-error-filesize-would-require-zip64-extensions

### DIFF
--- a/sickbeard/helpers.py
+++ b/sickbeard/helpers.py
@@ -1117,7 +1117,7 @@ def makeZip(fileList, archive):
     'archive' is the file name for the archive with a full path
     """
     try:
-        a = zipfile.ZipFile(archive, 'w', zipfile.ZIP_DEFLATED)
+        a = zipfile.ZipFile(archive, 'w', zipfile.ZIP_DEFLATED, allowZip64=True)
         for f in fileList:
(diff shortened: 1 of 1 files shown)


miigotu:
@fernandog I'm not sure if ZIP64 is available on all systems, thats is why it isn't enabled by default in the library, or I would have added it =P

Should we try to verify all modern systems support it or just merge and find out? lol
That guy just has an insanely large cache/library

fernandog:
Whats the best solution? Fallback?

miigotu:
Lets just merge it and see if anyone has problems with it?

fernandog:
@miigotu "ZIP64 extensions are disabled by default because the default zip and unzip commands on Unix (the InfoZIP utilities) don’t support these extensions"

https://docs.python.org/2/library/zipfile.html

so restore won't work

miigotu:
Lets just wait and see.

miigotu:
noone is running unix btw ^

PR #2139, opened 2015-07-19
Add post subtitle download scripts

miigotu:
SiCKRAGETV/sickrage-issues#1075

Argument 1 is the absolute path to the video file
Argument 2 is the absolute path to the new subtitle downloaded
Argument 3 OpenSubtitles 3 letter language code (Or should it be the full name of the language?)
Argument 4 Series name
Argument 5 Season
Argument 6 Episode
Argument 7 Episode title

Currently:
Gets called once for each subtitle downloaded for each video, so if a video downloads 10 subtitles, it is ran 10 times for that video. (I don't like this, I'd rather pass all of the subtitles at once, but we can decide that maybe after it gets some use)

Only gets called if embeded subtitles are allowed/not disabled (to prevent a race condition if your script deletes the srt after embedding it and SR keeps trying to re-download the subs)
Only gets called for mkv and mp4 (No other formats to embed subs in right?)

### DIFF
--- a/sickbeard/tv.py
+++ b/sickbeard/tv.py
@@ -1477,6 +1477,9 @@ class TVEpisode(object):
                     helpers.chmodAsParent(subpath)
                     helpers.fixSetGroupID(subpath)
 
+            if not sickbeard.EMBEDDED_SUBTITLES_ALL and sickbeard.SUBTITLES_EXTRA_SCRIPTS and self.location.endswith(('mkv','mp4')):
+                subtitles.run_subs_extra_scripts(self, foundSubs)
+
         except Exception as e:
             logger.log("Error occurred when downloading subtitles for: %s" % self.location)
             logger.log(traceback.format_exc(), logger.ERROR)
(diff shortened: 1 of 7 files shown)


sickbeard/tv.py
```diff
@@ -1480,6 +1480,9 @@ def downloadSubtitles(self, force=False):
 
             subliminal.save_subtitles(foundSubs, directory=subs_new_path, single=not sickbeard.SUBTITLES_MULTI)
 
+            if not sickbeard.EMBEDDED_SUBTITLES_ALL and sickbeard.SUBTITLES_EXTRA_SCRIPTS and self.location.endswith('mkv'):
```
fernandog: @miigotu why only mkv? WIP?

miigotu: Only mkv supports embedding subtitles in the container right?

miigotu: It should be working/complete, but I want the original requester to try it first.

PR #2142, opened 2015-07-19
Add Default Ep Status to WebAPI CMD_ShowAddNew

Hiltronix:
This adds an optional argument to set future episode status.

WebAPI CMD_ShowAddNew was not updating "default_ep_status" when adding new shows to function "addShow" in  "show_queue.py".  So new shows were having "default_status_after" set to None.  Leaving new shows added via webapi not being set to "wanted" or anything useful, requiring a manual episode status change to "wanted".

### DIFF
--- a/sickbeard/webapi.py
+++ b/sickbeard/webapi.py
@@ -2139,7 +2139,8 @@ class CMD_ShowAddNew(ApiCall):
                  "lang": {"desc": "the 2 letter lang abbreviation id"},
                  "subtitles": {"desc": "allow search episode subtitle"},
                  "anime": {"desc": "set show to anime"},
-                 "scene": {"desc": "show searches episodes by scene numbering"}
+                 "scene": {"desc": "show searches episodes by scene numbering"},
+                 "future_status": {"desc": "status of future episodes"}
              }
     }
 
@@ -2173,6 +2174,8 @@ class CMD_ShowAddNew(ApiCall):
         self.scene, args = self.check_params(args, kwargs, "scene", int(sickbeard.SCENE_DEFAULT), False,
                                              "int",
             [])
+        self.future_status, args = self.check_params(args, kwargs, "future_status", None, False, "string",
+                                              ["wanted", "skipped", "archived", "ignored"])
 
         # super, missing, help
         ApiCall.__init__(self, args, kwargs)
@@ -2237,6 +2240,22 @@ class CMD_ShowAddNew(ApiCall):
                 return _responds(RESULT_FAILURE, msg="Status prohibited")
             newStatus = self.status
 
+        # Assume default future status as wanted.
+        default_ep_status = WANTED
+        if self.future_status:
+            # convert the string status to a int
+            for status in statusStrings.statusStrings:
+                if statusStrings[status].lower() == str(self.future_status).lower():
+                    self.future_status = status
+                    break
+            # TODO: check if obsolete
+            if not self.future_status in statusStrings.statusStrings:
+                raise ApiError("Invalid Status")
+            # only allow the status options we want
+            if int(self.future_status) not in (3, 5, 6, 7):
+                return _responds(RESULT_FAILURE, msg="Status prohibited")
+            default_ep_status = self.future_status
+
         indexerName = None
         indexerResult = CMD_SickBeardSearchIndexers([], {indexer_ids[self.indexer]: self.indexerid}).run()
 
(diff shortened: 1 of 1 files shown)


MGaetan89:
Is this different from the existing `status` parameter?

Hiltronix:
Yes.  The existing status param is for previously aired episodes.  What I added is for future episodes.

sickbeard/webapi.py
```diff
@@ -2173,6 +2174,8 @@ def __init__(self, args, kwargs):
         self.scene, args = self.check_params(args, kwargs, "scene", int(sickbeard.SCENE_DEFAULT), False,
                                              "int",
             [])
+        self.future_status, args = self.check_params(args, kwargs, "future_status", None, False, "string",
```
miigotu: Shouldn't the None here be:

sickbeard.STATUS_DEFAULT_AFTER

sickbeard/webapi.py
```diff
@@ -2237,6 +2240,22 @@ def run(self):
                 return _responds(RESULT_FAILURE, msg="Status prohibited")
             newStatus = self.status
 
+        # Assume default future status as wanted.
+        default_ep_status = WANTED
```
miigotu: default_ep_status_after = WANTED

sickbeard/webapi.py
```diff
@@ -2237,6 +2240,22 @@ def run(self):
...
+            # convert the string status to a int
+            for status in statusStrings.statusStrings:
+                if statusStrings[status].lower() == str(self.future_status).lower():
+                    self.future_status = status
+                    break
+            # TODO: check if obsolete
+            if not self.future_status in statusStrings.statusStrings:
+                raise ApiError("Invalid Status")
+            # only allow the status options we want
+            if int(self.future_status) not in (3, 5, 6, 7):
```
miigotu: Please use the variables for status definition here, in case they ever change:

`not in (WANTED, SKIPPED, ARCHIVED, IGNORED):`

sickbeard/webapi.py
```diff
@@ -2270,7 +2289,7 @@ def run(self):
         sickbeard.showQueueScheduler.action.addShow(int(indexer), int(self.indexerid), showPath, newStatus,
                                                     newQuality,
                                                     int(self.flatten_folders), self.lang, self.subtitles, self.anime,
-                                                    self.scene)  # @UndefinedVariable
+                                                    self.scene, default_status_after=default_ep_status)  # @UndefinedVariable
```
miigotu: `default_status_after=default_ep_status_after`

miigotu:
Just a few small changes, need to stick to conformity.

Hiltronix:
I was following the same code style and format as was already written for the "status" argument.  It seemed to default to None in __init__

        self.status, args = self.check_params(args, kwargs, "status", None, False, "string",
                                              ["wanted", "skipped", "archived", "ignored"])

and then set the default in run

        # use default status as a failsafe
        newStatus = sickbeard.STATUS_DEFAULT

But I'm not aware that there is a system default for "default_ep_status" so my personal preference was to hard set it to WANTED.  But I'm sure that's open for debate, and can understand if someone else thinks the default should be None.

miigotu:
There is a default episode status app wide, as well as default episode status after. They are two separate application defaults, do not confuse the two.

My comments on this piece of code are the correct way, if the other part for the initial default status are different, then they are wrong.

miigotu:
@Hiltronix perfect, thanks.

MGaetan89:
This PR breaks things for me.
I'm not using the new (optional) parameter (`future_status `), but when I try to add a new show, I have the following response:
```javascript
{
    "data": {},
    "message": "param: 'future_status' with given value: '3' is out of allowed range '['wanted', 'skipped', 'archived', 'ignored']'",
    "result": "error"
}
```

The request I send is:
```
/?cmd=show.addnew&indexerid=292124&status=wanted&lang=en&anime=0&subtitles=0
```

Hiltronix:
Yes, I'm getting the same result when "future_status" is not used.

I'm thinking this is due to the variable "sickbeard.STATUS_DEFAULT_AFTER" not having a value set in line 2244.

	default_ep_status_after = sickbeard.STATUS_DEFAULT_AFTER

Or I am mistaken to assume that it has the same values as "sickbeard.STATUS_DEFAULT".

I originally had this hard coded to "WANTED" as the default, if "future_status" was not used.  But it was my understanding from miigotu that he wanted me to use "STATUS_DEFAULT" and "STATUS_DEFAULT_AFTER" based on his comments.  So I assumed that there would be a default value set to "sickbeard.STATUS_DEFAULT_AFTER".

Do you suggest we hard code the default in this function (which will solve the problem), or see why it isn't being set to a default value elsewhere?  Or maybe it doesn't use the same values as "sickbeard.STATUS_DEFAULT"

miigotu:
sickbeard.STATUS_DEFAULT_AFTER always has a value, as does sickbeard.STATUS_DEFAULT

Will look at what needs fixed in a bit, have to take my wife to work soon.

PR #2143, opened 2015-07-19
Don't allow chained call with command show.getnetworklogo

MGaetan89:


miigotu:
Should be for fanart also?

MGaetan89:
There is no command to display a fanart. I only added it to [`showPoster`](https://github.com/MGaetan89/SickRage/blob/159f52a7d2831a5ab1e7af90e913efe62f431919/sickbeard/webapi.py#L253) in one of my latest commit to match the [`webserve.py`](https://github.com/SiCKRAGETV/SickRage/blob/master/sickbeard/webserve.py#L382) version.

miigotu:
Ah ok, might be nice for you to be able to grab fanart via api and show it in your app.

MGaetan89:
I agree. It shouldn't be too difficult to add this, as `getPoster` already support this.
I will work on that :smile:

PR #2145, opened 2015-07-20
Remove zoink.ch

duramato:
removed zoink.ch as it appears to be hijacked by the ones running eztv ,as it is redirecting to their website


fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/2220

PR #2147, opened 2015-07-20
178.73.198.210 torrent caching website

duramato:
a torrent caching website running the torrage os project afaik, to fill the void of zoink.ch


miigotu:
Have no clue who is running that.
Not sure if I feel safe adding a site that isn't used by any trackers themselves.

neoatomic:
This looks to be another alternative/option.? http://thetorrent.org/

miigotu:
@neoatomic we can try to use some of those, thanks. I did add btcache.me but it was failing and I forget why so I removed it.

fernandog:
btcache was giving code 200 when it was not 200. remember we talked about?

go sleep! =)

duramato:
and what about itorrents.org ,why was it removed?

miigotu:
Thetorrent and btdig were added.

PR #2151, opened 2015-07-20
Migrate travis to a container-based infrastructure

MGaetan89:
As per the documentation http://docs.travis-ci.com/user/migrating-from-legacy/

This should make the build start faster on Tavis (event if they are not taking a lot of time right now :wink:).


fernandog:
@miigotu what do you think?

PR #2157, opened 2015-07-21
Stop building for 2.6, and set readme requirements to 2.7.x

miigotu:


MGaetan89:
Why not add Python 3 in `allow_failures`? Even if it is not supported, it helps seeing what's not working.

miigotu:
Because we already know 100% cheetah does not support python3.

Once we can replace cheetah with mako, we will add it in travis with allow_failures.

MGaetan89:
Ok makes sense.

PR #2165, opened 2015-07-22
Remove old clients files

fernandog:
@miigotu

### DIFF
--- a/SickBeard.py
+++ b/SickBeard.py
@@ -147,6 +147,30 @@ class SickRage(object):
         help_msg += "                --noresize          Prevent resizing of the banner/posters even if PIL is installed\n"
 
         return help_msg
+        
+    def fix_clients_nonsense():
+    
+        files = ["sickbeard/clients/download_station.py",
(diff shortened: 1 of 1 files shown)


fernandog:
@neoatomic i think this don't work in NAS. Some users reported having the last update but error persists

can you please:
stop SR
create a file called /SickRage/sickbeard/clients/rtorrent.py
create a file called /SickRage/sickbeard/clients/rtorrent.pyc
start SR
make sure that file was deleted

neoatomic:
@fernandog I dont use torrents so i hadn't any *.pyc in the folder. Except for generic.py.
So i removed those 2 files with empty dummy's. But after the restart they where still present. (and still empty.)

fernandog:
So this code does not work for NAS...

miigotu:
generic.py(c) is supposed to be there.

miigotu:
Check my commits just after, I had to hotfix very quickly.

fernandog:
@miigotu still nas users are reporting the same error. The don't work on NAS
@neoatomic confirmed.

PR #2168, opened 2015-07-22
Add GITUSER to IRC

fernandog:
@miigotu ideia!

maybe we can add two input fields in SR settings to user add user/pass of IRC. 
First we try to use them, if don't exists, use GIT, if don't exist, use "srforum|?"


miigotu:
Good enough for now.

PR #2171, opened 2015-07-22
Hide remote IP from log

fernandog:
@duramato or you want to hide local IP?


fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/2201

Arcanemagus:
Are you sure you want to include a gigantic library like that when a simple regex will do just as well for the simple use case you have in there?

fernandog:
@Arcanemagus would you like to make a PR for that? I couldn't make regex to check if IP is in those intervals

neoatomic:
Have to agree. That is a lot of code for hiding an ip in the log. :P

Arcanemagus:
This file has essentially the logic you want: https://github.com/ecleese/plexWatchWeb/blob/dev/datafactory/get-user-ip-stats.php

I'm at work so I can't dig into this codebase to apply it here right now.

fernandog:
@Arcanemagus ok! thanks

miigotu:
I agree with everyone else, you don't even need regex, let alone a big ass library =P

miigotu:
http://stackoverflow.com/a/28532296/1338570

if apikey and ip is not local, mask it
else do not mask.

PR #2174, opened 2015-07-23
Crop support image

fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/2279


duramato:
i belive this would be better it where changed to something like "Donate" ,as just support might be missleading

fernandog:
@duramato was checking if fixes user's issue

fernandog:
@duramato new image suggestion?

duramato:
@fernandog maybe something like the paypal donate botton ,i will try tinker and see if i come with something pleasing to the eye

fernandog:
and small ! =)

duramato:
@fernandog  maybe something like this ? i choose red as its a color that suits blue and white for both themes
(dimensions arent right yet)


donate

PR #2175, opened 2015-07-23
Check if show is mapped in XEM before query site

fernandog:
Related: https://github.com/SiCKRAGETV/sickrage-issues/issues/2300

Avoid check every show for http://thexem.de/map/all?id=%s&origin=%s&destination=scene
Probably need to remove some code.

https://github.com/SiCKRAGETV/SickRage/blob/develop/sickbeard/scene_numbering.py#L473

### DIFF
--- a/sickbeard/show_queue.py
+++ b/sickbeard/show_queue.py
@@ -30,6 +30,8 @@ from sickbeard import generic_queue
 from sickbeard import name_cache
 from sickbeard.exceptions import ex
 from sickbeard.blackandwhitelist import BlackAndWhiteList, short_group_names
+import json, requests
+
 
 
 class ShowQueue(generic_queue.GenericQueue):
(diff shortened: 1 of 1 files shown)


fernandog:
don't reduce query to xem. only changed the URL
I need to cache 
http://thexem.de/map/havemap?origin=tvdb
http://thexem.de/map/havemap?origin=rage
http://thexem.de/map/havemap?origin=anidb

before start the update. not in show_queue =(

fernandog:
@miigotu can I cache those here?

https://github.com/SiCKRAGETV/SickRage/blob/develop/sickbeard/showUpdater.py#L78

miigotu:
Why would you do this in the show_queue instead of inside the xem_refresh, where it already has a session?

any call to getURL that doesn't already have a session (like providers) need session=requests.Session() added to the getURL call.

fernandog:
because we need to cache those sites results before the loop th. shows
if we do in xem_refresh it will access the site N times(shows)

fernandog:
new PR coming

PR #2179, opened 2015-07-23
Handle xem_refresh better

miigotu:


fernandog:
thanks!

fernandog:
@miigotu 

still not caching page.
https://gist.github.com/fernandog/423a8a7423a9068b36d3

miigotu:
@fernandog just because you see that log line does not mean it is actually requesting direct to xem.

CacheControl is handled internally inside requests session, if the cache is not expired, it returns the cached result instead of opening a socket to xem.

You should not have added logs on successfule requests btw, thats crazy.

fernandog:
as we talked. I only added that to debug the new cache sites. will remove it

fernandog:
@miigotu we need to update this code too:

https://github.com/SiCKRAGETV/SickRage/blob/develop/sickbeard/scene_exceptions.py#L297

we only changed scene_numbering.py

fernandog:
scratch that. I think does not need to

miigotu:
nope that one is fine.

PR #2182, opened 2015-07-24
Some cleanup

miigotu:


fernandog:
Minor issue. Missed dot in some lines: %s Error:

%s. Error:

miigotu:
Because if we add a . after a url or path it is hard to see if the path has a . at the end or not when it is failing.

miigotu:
https://github.com/SiCKRAGETV/SickRage/pull/2182#issuecomment-124289739.
/opt/SickRage/data/somefile.

miigotu:
Still a lot of work needs done for log lines, they should all be done the same way. This is just a start

PR #2187, opened 2015-07-24
Text factory decoding order change - Fixes SiCKRAGETV/sickrage-issues…

miigotu:
…#2319

@fernandog


fernandog:
Will this fix that windows locale issue?

miigotu:
I think so

PR #2195, opened 2015-07-26
Fix bugs preventing valid torrents from passing data checks

joshuahiggins:
Torrents that had 0 seeders or leechers would fail due to the falsey check that `if not seeders` is doing, preventing valid torrents from being passed through for quality checks. Changed it to explicitly look for the default value of `None`.

To reproduce this bug, use the HD-Torrents provider and attempt to download a TV show that is available but has 0 leechers. Review the debug logs and you will not find any "[HDTorrents] :: Found result" logs, indicating that all available torrents failed the initial checks before the quality check.


joshuahiggins:
Closing to reopen from dev branch.

PR #2196, opened 2015-07-26
Fix bug preventing torrents with 0 leechers from being downloaded

joshuahiggins:
Torrents that had 0 seeders or leechers would fail due to the falsey check that if not seeders is doing, preventing valid torrents from being passed through for quality checks. Changed it to explicitly look for the default value of None.

To reproduce this bug, use the HD-Torrents provider and attempt to download a TV show that is available but has 0 leechers. Review the debug logs and you will not find any "[HDTorrents] :: Found result" logs, indicating that all available torrents failed the initial checks before the quality check.

### DIFF
--- a/sickbeard/providers/hdtorrents.py
+++ b/sickbeard/providers/hdtorrents.py
@@ -232,7 +232,7 @@ class HDTorrentsProvider(generic.TorrentProvider):
                 except (AttributeError, TypeError, KeyError, ValueError):
                     continue
 
-                if not title or not url or not seeders or not leechers or not size or \
+                if not title or not url or seeders == None or leechers == None or not size or \
                         seeders < self.minseed or leechers < self.minleech:
                     continue
 


neoatomic:
Thanks for contributing, its much appropriated. ;)

miigotu:
Thanks, I see the bug here, but you're change introduces another bug also. Can you change it so that the seeders part stays the same and only change the leechers part?

Also, just for sake of being pythonic, please use `is` for `==` and `is not` for `!=`.

We absolutely do not want 0 seeder torrents, 0 leechers is fine.

`if not title or not url or not seeders or leechers is None or not size or \`

miigotu:
@joshuahiggins I merged this via command line with the change I requested, if you disagree with my change or it isn't working let me know.

Closing this.

joshuahiggins:
I don't disagree with you, but since you give the user the option to set their own minseed and don't prevent them from setting 0, technically preventing 0 from passing here would be a bug. Instead of silently failing here, I'd prevent the user from setting 0 on the search provider settings page.

miigotu:
The '0' setting just means there is no preference, but I can see where the misconception would be. We don't want to pick dead torrents in any situation.

joshuahiggins:
I personally have no interest in 0 seeder torrents, but I wouldn't say there's no value to allowing them. Most sites purge old torrents according to their own timeouts, so torrents with 0 seeders aren't actually too common as is. Looking at HDT, there's ~800 torrents with 0 seeders, out of the ~39,000 overall torrents.

There's 2 scenarios where these 0 seeder torrents tend to commonly exist:

- New torrents. Sometimes a seeder starts a torrent but leaves before any leechers are able to finish. This is very unlikely on most well maintained sites with scene releases, but on some of the smaller communities I wouldn't rule this out.
- Low interest torrents that have yet to be purged. Some sites, like HDT, allow users to submit reseed requests to torrents that no longer have seeds but haven't been purged yet. In the past I received a few dozen reseed requests when I uploaded a full season rip of a less popular show and eventually stopped seeding it (as I was the only one).

These are more fringe cases, but are meant to illustrate why a user may set `minSeed = 0` and actually mean it. I know I put mine at 1 because I had assumed that's what it meant. In my opinion, there's value to allowing 0 to pass through for some users.

Granted, the drawback is that making this change to HDT (or other sites) potentially opens up a current SickRage user to snagging a 0 seeder torrent when previously they wouldn't have. In my opinion, I'd consider that happening far more of a fringe thing when compared to a user actually wanting to be able to pull 0 seeder torrents. However, if you're following semver then that would still qualify for a minor version upgrade as at this point the bug in the code has become the intended feature, hah.

Anyway, food for thought. Ultimately, for my personal use I only cared about the leechers half of the bug as it was causing pretty much everything in my backlog that was more than a couple years old to fail. Cheers!

fernandog:
@joshuahiggins I can't download from HDT. can you? I tried manual search and I always get no results found. Wondering if they have more than one server with different html code page.

joshuahiggins:
@fernandog Not sure what you mean? Through SickRage or on the actual HDT site? I've never had a problem with their site. The only problem I had with SickRage is what was fixed here, regarding torrents with 0 leechers being blocked.

miigotu:
@fernandog make sure you are searching for an episode that exists on hdt.  Most shows only have season packs on hdt, and you cannot manual search for a season pack because the code only allows searching for seasons when the search includes more than 1 episode.

PR #2199, opened 2015-07-26
Use issues even if they are closed for issue submitter

miigotu:
This PR will make issues submitted from inside SR always use an existing issue whether it is opened or closed, instead of creating a new issue. It will only create a new issue if one does not exit with the same title.


fernandog:
Is it possible to post in the newer one? Otherwise it could post in random issues with same title

miigotu:
It should always post in the newest one, because the results are returned in newest->oldest and when looping through results SR stops at the first match.

PR #2217, opened 2015-07-28
Check if SR is updated before submitting an issue

fernandog:
@miigotu I couldn't not test properly. I got: 

`DEBUG::Thread-53 :: git didn't return newest commit hash`

the logic seems ok. what do you think? No compile errors.

### DIFF
--- a/sickbeard/logger.py
+++ b/sickbeard/logger.py
@@ -152,9 +152,22 @@ class Logger(object):
             sys.exit(1)
 
     def submit_errors(self):
-        if not (sickbeard.GIT_USERNAME and sickbeard.GIT_PASSWORD and len(classes.ErrorViewer.errors) > 0):
-            self.log('Please set your GitHub username and password in the config, unable to submit issue ticket to GitHub!')
+        if not (sickbeard.GIT_USERNAME and sickbeard.GIT_PASSWORD and sickbeard.DEBUG and len(classes.ErrorViewer.errors) > 0):
+            self.log('Please set your GitHub username and password in the config and enable debug. Unable to submit issue ticket to GitHub!')
             return
+          
+        try:
+            from versionChecker import CheckVersion
+            checkversion = CheckVersion()
+            needs_update = checkversion.check_for_new_version()
+            commits_behind = checkversion.updater.get_num_commits_behind()
+        except:
+            self.log('Could not check if your SickRage is updated, unable to submit issue ticket to GitHub!')
+            return
+
+        if commits_behind is None or commits_behind > 0:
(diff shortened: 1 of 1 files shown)


sickbeard/logger.py
```diff
@@ -155,6 +155,19 @@ def submit_errors(self):
         if not (sickbeard.GIT_USERNAME and sickbeard.GIT_PASSWORD and len(classes.ErrorViewer.errors) > 0):
```
miigotu: Should make this line:
`if not all([sickbeard.DEBUG, sickbeard.GIT_USERNAME, sickbeard.GIT_PASSWORD, classes.ErrorViewer.errors]):
(would need to test it to make sure it behaves as expected, if not could just add the sickbeard.DEBUG part in the line above.

sickbeard/logger.py
```diff
@@ -155,6 +155,19 @@ def submit_errors(self):
...
             self.log('Please set your GitHub username and password in the config, unable to submit issue ticket to GitHub!')
```
miigotu: and add `and enable debug` to this log line.

MGaetan89:
Isn't checking the value of `needs_update` (which is unused at the moment) enough?
Otherwise, the logic looks good to me too. Don't know why you have this error.

sickbeard/logger.py
MGaetan89: Isn't there an issue to automatically turning `debug` on before sending an issue?

fernandog:
@MGaetan89 i didn't check because if git update check fails it says 'update not needed' - which is wrong!

sickbeard/logger.py
miigotu: If we automatically turn debug on just before submitting, the error still is not logged with debug in the logs. It has to be enabled and then retrigger the error.

fernandog:
@miigotu any suggestion to avoid submit issue when couldn't check for update?

sickbeard/logger.py
WebSpider: Well, needs_update is only set if SR knows it needs an update ... If it doesn't, because checkversion doesn't run for instance, we should complain about that instead ;)

-----Original Message-----

miigotu:
Maybe just allow it when the update check fails? That would still stop most non-updated bug reports.
I'm not sure how exactly we should handle that situation.
Some version checks fail because of people using things like py2exe and not having git at all.

fernandog:
@miigotu merge it?

miigotu:
Yep, looks solid. we will find out if it isn't =P.

fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/2372

PR #2222, opened 2015-07-28
Mako

MGaetan89:
Moving forward with Mako migration.

- I created all `.mako` files (they are just a copy of the current `.tmpl` files). So now when someone work on the Mako migration, we will see the actual changes in the file, and not just a big new file because it wasn't there yet. The issue is that it is now hard to know what was done and what still needs to be done. Two solutions:
  - Delete the `tmpl` file as we are done with them
  - Use GitHub tasks list markdown format to track the progress (harder to maintain):
    - [ ] fileA.tmpl
    - [x] fileB.tmpl
- I updating the `apiBuilder.mako` file to the new format. To see the changes in the file: https://github.com/MGaetan89/SickRage/commit/ee03e01f6bbda5182cf69935d99751036aeb1f2b

I don't know if something inside SickRage was already done to use Make template files, so I did not try my changes.

### DIFF
--- a/gui/slick/interfaces/default/IRC.mako
+++ b/gui/slick/interfaces/default/IRC.mako
@@ -1,0 +1,15 @@
+#import sickbeard
+#from sickbeard.common import *
+
+#import os.path
+#include $os.path.join($sickbeard.PROG_DIR, "gui/slick/interfaces/default/inc_top.tmpl")
+
+#if $sickbeard.GIT_USERNAME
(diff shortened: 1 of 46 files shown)


miigotu:
Need to rebase. I had to rebase and force a push to the mako branch, where I just commented the commit that did both trim trailing and tab replace in one commit, and I did the trim trailing and tab replace in develop.

Was having conflicts trying to keep it in sync

PR #2225, opened 2015-07-29
Fix equality vs identity problem in catching requests status codes fo…

miigotu:
…r trakt

Catch the status codes correctly so we can debug SiCKRAGETV/sickrage-issues/issues/2370


fernandog:
using 'is' = pep8?

miigotu:
Not always. Sometimes you must use ==.

```python
>>> code = 404
>>> code is 404
False
>>> code == 404
True
>>> code in 404
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
TypeError: argument of type 'int' is not iterable
>>> code in (403,404)
True
>>> code in [403,404]
True
>>>
```

PR #2226, opened 2015-07-29
[DO NOT CLOSE] Started converting to mako

OmgImAlexis:
@miigotu  Don't pull until more is done.


OmgImAlexis:
Pretty sure I accidently replaced some ``if not`` statements with ``ifnot``. Needs to be fixed.

OmgImAlexis:
@miigotu that should work now. I'll start on the other less important files now.

OmgImAlexis:
Heading to bed, @miigotu if you don't work on this in the next 12 - 24 hours I'll finish this off along with any errors I made.

WebSpider:
Will this be wrapped in one big commit later on?

OmgImAlexis:
@WebSpider probably, I'm trying to get all of the templates to mako and then @miigotu and I are going to work on taking out all of the SQL and such from there to where it's meant to be. I think for now we're just getting this to the stage where mako can be completely used instead of cheetah.

WebSpider:
Ok ill read up on Mako if we're ditching Cheetah altogether.

OmgImAlexis:
What's going on with the first 3 lines in SickRage/gui/slick/interfaces/default/config.mako, do we just dump them in the code block?

Also lines 24 - 34 with the cpu comments, etc. are they needed anymore?

gui/slick/interfaces/default/IRC.mako
```diff
@@ -1,15 +1,17 @@
-#import sickbeard
-#from sickbeard.common import *
+<%
+    import sickbeard
+    from sickbeard.common import *
 
-#import os.path
-#include $os.path.join($sickbeard.PROG_DIR, "gui/slick/interfaces/default/inc_top.tmpl")
+    import os.path
+    include $os.path.join($sickbeard.PROG_DIR, "gui/slick/interfaces/default/inc_top.tmpl")
```
MGaetan89: Do we need the `$` here?

gui/slick/interfaces/default/IRC.mako
```diff
@@ -1,15 +1,17 @@
...
+    else
+        username = "SickRageUI|?"
+    endif
+%>
 
-<iframe id="extFrame" src="https://kiwiirc.com/client/irc.freenode.net/?nick=$username&theme=basic#sickrage" width="100%" height="500" frameBorder="0" style="border: 1px black solid;"></iframe>
+<iframe id="extFrame" src="https://kiwiirc.com/client/irc.freenode.net/?nick=${username}&theme=basic#sickrage" width="100%" height="500" frameBorder="0" style="border: 1px black solid;"></iframe>
 
-#include $os.path.join($sickbeard.PROG_DIR, "gui/slick/interfaces/default/inc_bottom.tmpl")
+% include $os.path.join(sickbeard.PROG_DIR, "gui/slick/interfaces/default/inc_bottom.tmpl")
```
MGaetan89: And `$` here too.

gui/slick/interfaces/default/IRC.mako
OmgImAlexis: Nope, mako only needs it for inline HTML. Check the next commit. https://github.com/SiCKRAGETV/SickRage/commit/4348f6f8d429237f79ec97a5428cc7a28795f929

MGaetan89: So they can be removed here too :)
I was going over all the changes to see if every thing was updated.

gui/slick/interfaces/default/IRC.mako
OmgImAlexis: Jump on IRC on #sickrage it'll be easier to fix these as I'm working on them as we speak.

OmgImAlexis:
That's all I can do for now, 5 hours of solid typing kills your arms. :sob: 

I'll be back on tomorrow.

neoatomic:
@OmgImAlexis Thanks for contributing, keep up the good work.!

gui/slick/interfaces/default/IRC.mako
miigotu: http://docs.makotemplates.org/en/latest/syntax.html#include

OmgImAlexis: So we'd be looking at something like this?
``include file=os.path.join(sickbeard.PROG_DIR, "gui/slick/interfaces/default/inc_bottom.mako")``

OmgImAlexis:
@miigotu 

Oh no... I've been writing the if statements backwards.
Is true or false first in this?

``${(' checked="checked"', '')[sickbeard.LAUNCH_BROWSER == True]}``

OmgImAlexis:
Just going through the files and doing a once over with the # to % and then I'll sort out the inline if statements.

OmgImAlexis:
@miigotu is there a reason why displayShow.mako and displayShow.tmpl look like  completely different files?

OmgImAlexis:
Is there any point in converting the trendingShows since I don't think that's used anymore, right?

miigotu:
@OmgImAlexis yes, trending shows is only visible if you have trakt enabled.

OmgImAlexis:
Ah, that would make sense. Trakt was playing up like 6 months ago and I never reactivated it, oops.

OmgImAlexis:
@miigotu how would I go about testing the mako files?
Has SR actually been changed on the mako branch to use them or is it still using the tmpl files?

miigotu:
@OmgImAlexis It's sort of a chicken/egg situation with testing.

As soon as the initial conversion is done I will make the changes again in webserve.py to use mako. Also the readme and whatnot need references to cheetah removed and mako added.

I might try the switch tonight. It's just that inc_top, home, and inc_bottom need to be 100% done or there is no easy way to debug the changes I make in webserve.py.

OmgImAlexis:
Ah that's a pain, well if you want to do that I'll do another couple of checks on those to make sure they're completely done. They should be mostly working, the only thing that may not would be the imports. I'm pretty sure the way I have it now it should work. I'll have a look more at home now. I think that's the one I missed.

miigotu:
There is a funky line in inc_top that I need to work on too, I had skipped it.

https://github.com/OmgImAlexis/SickRage/blob/mako/gui/slick/interfaces/default/inc_top.mako#L269

It has two ifs/else, but it seems like it is completely wrong before the change lol. Unless I'm reading it wrong.

OmgImAlexis:
I'll fix that now. I think I know what it's meant to be.

WebSpider:
You will be using the mako branch for initial testing, i presume? So we dont break develop in a horrible way when something goes wrong with mako :)

-----Original Message-----

OmgImAlexis:
@WebSpider, considering what just happened I really hope so.

miigotu:
@WebSpider yes, it will stay in mako branch until it is all working and well tested. And then we need to make a note/deprecation warning in develop and master for at least a week saying that mako is going to be replacing cheetah and they must install mako to work with the new version.

There will still surely be a mass of complaints and issue reports and questions when the actual swap happens, but that is the cost of major changes.

If anyone has better ideas about how to handle the change-over period let me know. Im open to any suggestions about how to make the transition go smoother for our userbase.

WebSpider:
Suggestion:  When we have a somewhat workable mako solution, set the template engine in develop as configurable using a config.ini setting, not exposed to the ui.

Encourage develop users to start using mako and fix bugs in develop. 

When all is deemed stable, notice master and do the flip.

-----Original Message-----

OmgImAlexis:
@WebSpider add comments to the new pull.

https://github.com/SiCKRAGETV/SickRage/pull/2229

From the way things are going if @miigotu can get the webserve stuff sorted most of the templates should be completely converted and fixed within a day. I'm not sure if there'd be any point in adding in code for both but he may disagree.

PR #2245, opened 2015-08-01
Slicker Theme

Goeny:
Not finished, but it works


miigotu:
Wow 1400 changed files.

working on porting slick to mako now, maybe we can work on porting slicker to mako next and then finish it.

Goeny:
Most real changes are in the tmpl and css files. Some small changes in .js files. Also some new files, and as i mentioned before in the forum, a lot of old files are still there which are not used. All the images are in the theme aswell. Probably would be better to hardcode the img src link to the slick images folder. But for some reason i can't get that to work so i decided to leave it like this for now and look into it later.

miigotu:
I think we should separate most of the images from the theme.
flags and provider/network icons do not need to be part of the theme itself.

Goeny:
Agree!
Calling the flags, provider and network images should be the same way as i.e. show posters, banners and fanart. Would make maintaining of different themes easier.

I will stop working on the theme for now.
Once Mako is up and running i will resume working on it (depending on the time i have available).

For now, the theme is available for everyone who is interested to try out, or to work on it.

miigotu:
Ok. BTW, mako branch is about half done. config pages arent working yet but show list, display show, news, irc, changelog, help&info, restart, episode status management, history, log pages working

OmgImAlexis:
@Goeny I'll start work on moving this to mako now. Should be done in a few days. I'll mention you in the PR once it's open.

OmgImAlexis:
https://github.com/SiCKRAGETV/SickRage/pull/2324

PR #2250, opened 2015-08-02
Mako conversion pt3

OmgImAlexis:


OmgImAlexis:
Configs can't be fixed until imports are sorted.

I'm getting ``CompileException: 'import *' is not supported, since all identifier names must be explicitly declared.  Please use the form 'from <modulename> import <name1>, <name2>, ...' instead. in file '/Users/xo/SickRage/gui/slick/interfaces/default/config_general.mako' at line: 2 char: 1`` which I'm guessing won't allow me to see any other erorrs until it's fixed.

OmgImAlexis:
Inline if then else for checked and selected should be changed to the ${} version like the other files.

PR #2255, opened 2015-08-02
Adding IMDB Popular View for discovering new shows.

jonathanendersby:


WebSpider:
How would one configure USE_IMDB_MOVIEMETER?

jonathanendersby:
@WebSpider USE_IMDB_MOVIEMETER is simply a boolean which determines whether or not to show the "View Popular Shows" button on the "Add Show" page.

WebSpider:
@jonathanendersby you haven't added code to save/read the value from config.ini to make it persistant across restarts.

Check sickbeard/__init__.py from line 1632 and webserve.py for inspiration :)

jonathanendersby:
Adding a screenshot so that people can see what the heck I'm on about.

image

I'm not sure if I'm going to have time in the next few days to add the save/read config stuff, so if anyone else wants to do that, please do.

fernandog:
I don't think we need an option to show/hide this.  Just show it.

WebSpider:
I will add the option to config.ini, and by default enable it, so we can add the dropdown in config menu when users start complaining they hate IMDB and only use SR for Anime .... Im sure there are users like that :)

Should be done in about 12h.

-----Original Message-----

jonathanendersby:
:+1: @WebSpider 

PS. The one issue that should be raised is that the images (and data) are not cached, and it is hotlinking to the IMDB images, which technically could be viewed as "privacy leakage"... but I personally don't have too much of an issue with that. Not sure what the status quo is?

miigotu:
I think this needs to be completed.  @WebSpider whichever changes you make to it can you make the PR to @jonathanendersby repo on the imdb_popular branch?  He can merge it, then it will show here, then when it is complete we can merge this.

WebSpider:
@miigotu offcourse!

WebSpider:
@jonathanendersby I've pulled your branch, however the IMDB view isn't showing up on a default install

adrianmoisey:
\o/

jonathanendersby:
@WebSpider What does the /home/addShows/ page look like? It should look like this:

image

Is USE_IMDB_MOVIEMETER set to True?

miigotu:
@jonathanendersby would you mind porting this to the mako templates now? since mako has been merged to develop this is out of date, and I would like to get this added soon.

sickbeard/imdbPopular.py
```diff
@@ -0,0 +1,55 @@
+import requests
+from bs4 import BeautifulSoup
+import re
+
+url = "http://www.imdb.com/search/title?at=0&sort=moviemeter&title_type=tv_series&year=2014,2016"
```
Hellowlol: Don't  hardcode the years ;)

PR #2261, opened 2015-08-03
Fix usenet crawler searchstring paramaters

KontiSR:
Fix usenet crawler searchstring paramaters 
'ep' and 'rid' don't work anymore. I've replaced these with 'episode' and 'tvrage' specific for usenet-crawler. Searching series/anime should work now, while other newznab based providers should not be effected.

### DIFF
--- a/sickbeard/providers/newznab.py
+++ b/sickbeard/providers/newznab.py
@@ -147,7 +147,7 @@ class NewznabProvider(generic.NZBProvider):
 
         # search
         rid = helpers.mapIndexersToShow(ep_obj.show)[2]
-        if rid:
+        if rid and self.name not in "Usenet-Crawler":
             cur_params['rid'] = rid
         elif 'rid' in cur_params:
             cur_params.pop('rid')
@@ -172,20 +172,21 @@ class NewznabProvider(generic.NZBProvider):
             return [params]
 
         params['maxage'] = (datetime.datetime.now() - datetime.datetime.combine(ep_obj.airdate, datetime.datetime.min.time())).days + 1
-
+        
+        episode = 'ep' if self.name not in "Usenet-Crawler" else 'episode'
         if ep_obj.show.air_by_date or ep_obj.show.sports:
             date_str = str(ep_obj.airdate)
             params['season'] = date_str.partition('-')[0]
-            params['ep'] = date_str.partition('-')[2].replace('-', '/')
+            params[episode] = date_str.partition('-')[2].replace('-', '/')
         elif ep_obj.show.anime:
-            params['ep'] = "%i" % int(ep_obj.scene_absolute_number if int(ep_obj.scene_absolute_number) > 0 else ep_obj.scene_episode)
+            params[episode] = "%i" % int(ep_obj.scene_absolute_number if int(ep_obj.scene_absolute_number) > 0 else ep_obj.scene_episode)
         else:
             params['season'] = ep_obj.scene_season
-            params['ep'] = ep_obj.scene_episode
+            params[episode] = ep_obj.scene_episode
 
         # search
         rid = helpers.mapIndexersToShow(ep_obj.show)[2]
-        if rid:
+        if rid and self.name not in "Usenet-Crawler":
             params['rid'] = rid
         elif 'rid' in params:
             params.pop('rid')
@@ -202,9 +203,9 @@ class NewznabProvider(generic.NZBProvider):
 
             if ep_obj.show.anime:
                 paramsNoEp = params.copy()
-                paramsNoEp['q'] = paramsNoEp['q'] + " " + paramsNoEp['ep']
-                if "ep" in paramsNoEp:
-                    paramsNoEp.pop("ep")
+                paramsNoEp['q'] = paramsNoEp['q'] + " " + paramsNoEp[episode]
+                if episode in paramsNoEp:
+                    paramsNoEp.pop(episode)
                 to_return.append(paramsNoEp)
 
         return to_return


fernandog:
what about tvdb?

sickbeard/providers/newznab.py
```diff
@@ -141,9 +141,11 @@ def _get_season_search_strings(self, ep_obj):
 
         # search
         rid = helpers.mapIndexersToShow(ep_obj.show)[2]
+        #added exception for usenet-crawler, as it seems they have changed something within their api
+        tvrage = 'rid' if self.name <> "Usenet-Crawler" else 'tvrage'
```
miigotu: why `<>` and not in ?

KontiSR: I'm just doing to many different languages at the moment. <> works the same as !=. <> is supported, but i just read it's deprecated. If you'd like I can change it to !=.

Personally i only use "not in", with lists or tuples. But also that's possible, if you prefer, I can also use that.

fernandog:
@KontiSR merge conflicts

Step 1: From your project repository, check out a new branch and test the changes.
git checkout -b KontiSR-dev-fix-usenet-crawler-searchstring-para develop
git pull git://github.com/KontiSR/SickRage.git dev-fix-usenet-crawler-searchstring-para

Step 2: Merge the changes and update on GitHub.
git checkout develop
git merge --no-ff KontiSR-dev-fix-usenet-crawler-searchstring-para
git push origin develop

sickbeard/providers/newznab.py
```diff
@@ -162,22 +164,25 @@ def _get_episode_search_strings(self, ep_obj, add_string=''):
 
         if not ep_obj:
             return [params]
+        
+        episode = 'ep' if self.name <> "Usenet-Crawler" else 'episode'
```
autronix: This is not unique to Usenet-Crawler, I encounter the same behavior wih nzb.su; could possibly apply to others - maybe looping through a list of providers would be a better choice than setting a single provider string

sickbeard/providers/newznab.py
```diff
@@ -195,11 +200,11 @@ def _get_episode_search_strings(self, ep_obj, add_string=''):
                 params['q'] = helpers.sanitizeSceneName(cur_exception)
                 paramsNoEp = params.copy()
                 
-                paramsNoEp['q'] = paramsNoEp['q'] + " " + str(paramsNoEp['ep'])
-                if "ep" in paramsNoEp:
-                    paramsNoEp.pop("ep")
+                paramsNoEp['q'] = paramsNoEp['q'] + " " + str(paramsNoEp[episode])
```
autronix: My understanding is that paramsNoEp['episode'] is already a string formated by "%i" % ... in line 171. Also worth noting that for both providers I've tried (usenet-crawler and nzb.su) having an episode without a 0 padding when when in single digits does not yield a result: q=Show.Name+3 => no result while q=Show.Name+03 has result

miigotu: There should be no ep # in the q string if its added as ep/episode parameter. Thats why there is params and paramsNoEp

sickbeard/providers/newznab.py
miigotu: To be pyhtoninc you should only use != (cstyle comparison operators) when comparing integers

I would rather see 
```python
tvrage = 'rid' if self.name not in "Usenet-Crawler" else 'tvrage' 
```
or
```python
tvrage = ('rid', 'tvrage') ["Usenet-Crawler" in self.name] 
```

KontiSR: Aaargggg, you'r right. Crap. okay, i'm just going to pop the paramater

KontiSR:
Aah it's the same for ep as it is for rid. Both paramaters are not functioning. Leaving them out has the same effect as adding the 'episode' and 'tvrage'. Pff such a mess. If I also leave episode out, sickrage will get back an enormous set of results to parse, each time it's doing a daily/hourly/manual search.

I'm really in doubt if we should want to fix this in SR.

autronix:
Same behavior appears on nzb.su, not only Usenet-Crawler.
It might be ideal to have a list rather than just a string matching Usenet-Crawler

sickbeard/providers/newznab.py
```diff
@@ -147,7 +147,7 @@ def _get_season_search_strings(self, ep_obj):
 
         # search
         rid = helpers.mapIndexersToShow(ep_obj.show)[2]
-        if rid:
+        if rid and self.name not in "Usenet-Crawler":
```
autronix: We would need to have something like 
```
newznab_providers = ["Usenet-Crawler", "NZB.su", ....]
if rid and self.name not in newznab_providers:
[.....]

```

sickbeard/providers/newznab.py
```diff
@@ -202,9 +203,9 @@ def _get_episode_search_strings(self, ep_obj, add_string=''):
 
             if ep_obj.show.anime:
                 paramsNoEp = params.copy()
-                paramsNoEp['q'] = paramsNoEp['q'] + " " + paramsNoEp['ep']
-                if "ep" in paramsNoEp:
-                    paramsNoEp.pop("ep")
+                paramsNoEp['q'] = paramsNoEp['q'] + " " + paramsNoEp[episode]
```
autronix:
```python
if int(paramsNoEp[episode]) < 10:
    paramsNoEp['q'] = paramsNoEp['q'] + " 0" + paramsNoEp[episode]
else:
    paramsNoEp['q'] = paramsNoEp['q'] + " " + paramsNoEp[episode]
```

thezoggy:
you dont do rid + query. you do one or the other.... notice that this 'issue' doesnt exist in sonarr/sb

miigotu:
Calm down @thezoggy I'm working on it now.

KontiSR:
We've already realized that the way it has been implemented is nog ideal. Im also werking on a rewrite, where i dump the unnecessary paramaters all together for newznab.

miigotu:
@KontiSR I have it half way there already. Just need to decide if I should use season/ep params with the q string and no rid, or if I should put the ep/season in the q string.

Debating just using season, ep, and rid and skipping scene name exceptions for newznab (no q searches)

I promise if I do that, and they still have some complaint, then they are full of shit. It's one thing that I take care of everything they point out and then some every time, but I don't like how it's made out to be that someone is an idiot or something. Like seriously, even SB is far from perfect.

thezoggy:
@miigotu i'm plenty calm, i dont use sr :)

sickbeard/providers/newznab.py
thezoggy: http://stackoverflow.com/questions/339007/nicest-way-to-pad-zeroes-to-string

autronix: @thezoggy thanks - I haven't done python in a very long time
```python
paramsNoEp['q'] = paramsNoEp['q'] + " " + paramsNoEp[episode].zfill(2)
```

Taloth:
@miigotu I was warned of possible issues with Usenet-crawler by zoggy, so I tested it a bit yesterday on Better Call Saul. The api seems to be working like other nn indexers.
Basically you should do `rid+season+ep` or `q+season+ep` depending on whether you have the tvrageid yourself. IIRC for SB they did the `q` search only if either the tvrageid is 0 or no results were found.
The problem with `rid+q+season+ep` is that, if the indexer doesn't have the tvrageid associated with any release, you'll get no results at all, and if there are associated releases, the `q` parameter will simple filter further instead of adding additional releases by scene exception. To put it simple, all those parameters use `and` operators.

I hope that helps. Feel free to bug me if you have questions.

KontiSR:
@miigotu Tnx for giving this some attention! I think we all have a pretty good understanding on how to get this fixed. As this is not really the ideal place to discus this (we already have have the open issue), im going to close this one.

PR #2274, opened 2015-08-06
Add new Deluge Daemon Client

PaulW:
Add new client method to support directly connecting to the Deluge Daemon service, without the requirement to go via the WebUI.

Existing WebUI Client method is still retained, so the option for either still exists.

Helpful for fully headless clients, or minimal installs of Deluge, or where access via WebUI is not available.


miigotu:
@PaulW Thanks. Although, there are 3 problems with the PR. You need to base it on develop, and you need to squash commits, and make the PR to develop

fernandog:
@PaulW can you merge into develop please?

PaulW:
Hi

I'll get the commit & change processes into the dev branch tonight.  I've
had a mad few days recently, so not had chance to catch up.

PR #2277, opened 2015-08-06
Only delete show if is not in any queue/action

fernandog:
@miigotu how can I add all those actions in a variable and call  _isInQueue(showObj,actions) ?

### DIFF
--- a/sickbeard/webserve.py
+++ b/sickbeard/webserve.py
@@ -1619,12 +1619,27 @@ class Home(WebRoot):
                 logger.log("Trakt: Unable to delete show: {0}. Error: {1}".format(showObj.name, ex(e)),logger.ERROR)
                 return self._genericMessage("Error", "Unable to delete show: {0}".format(showObj.name))
 
-        showObj.deleteShow(bool(full))
-
-        ui.notifications.message('%s has been %s %s' %
+        try:
+            from sickbeard.show_queue import ShowQueue
+            ShowActions = ShowQueue()
+            queue_actions = (ShowQueueActions.ADD, ShowQueueActions.UPDATE, ShowQueueActions.FORCEUPDATE, ShowQueueActions.REFRESH, ShowQueueActions.RENAME, ShowQueueActions.SUBTITLE)
+            if not ShowActions._isInQueue(showObj, queue_actions) and not ShowActions._isBeingSomethinged(showObj, queue_actions):
+                #Delete show if not in queue or doing something with it
+                showObj.deleteShow(bool(full))
+                ui.notifications.message('%s has been %s %s' %
                                  (showObj.name,
                                   ('deleted', 'trashed')[sickbeard.TRASH_REMOVE_SHOW],
                                   ('(media untouched)', '(with all related media)')[bool(full)]))
+            else:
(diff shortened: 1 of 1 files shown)


sickbeard/webserve.py
```diff
@@ -1619,12 +1619,26 @@ def deleteShow(self, show=None, full=0):
                 logger.log("Trakt: Unable to delete show: {0}. Error: {1}".format(showObj.name, ex(e)),logger.ERROR)
                 return self._genericMessage("Error", "Unable to delete show: {0}".format(showObj.name))
 
-        showObj.deleteShow(bool(full))
-
-        ui.notifications.message('%s has been %s %s' %
+        try:
+            from sickbeard.show_queue import ShowQueue
+            ShowActions = ShowQueue()
```
miigotu: queue_actions = (ShowQueueActions.UPDATE, ShowQueueActions.FORCEUPDATE, ShowQueueActions.REFRESH, ShowQueueActions.RENAME, ShowQueueActions.SUBTITLE)

sickbeard/webserve.py
```diff
@@ -1619,12 +1619,26 @@ def deleteShow(self, show=None, full=0):
...
+            if not ShowActions._isInQueue(showObj,(ShowQueueActions.UPDATE, ShowQueueActions.FORCEUPDATE, ShowQueueActions.REFRESH, ShowQueueActions.RENAME, ShowQueueActions.SUBTITLE)) and not ShowActions._isBeingSomethinged(showObj, (ShowQueueActions.ADD, ShowQueueActions.UPDATE, ShowQueueActions.FORCEUPDATE, ShowQueueActions.REFRESH, ShowQueueActions.RENAME, ShowQueueActions.SUBTITLE) ):
```
miigotu: if not ShowActions._isInQueue(showObj,queue_actions) and not ShowActions._isBeingSomethinged(showObj, queue_actions):

miigotu:
Not sure if I like this approach. Why can't add another queue action that deletes the show. That way you know no other actions are running using that show, and you can remove any actions in the queue involving that show after the remove action. set the delete action as high priority.

fernandog:
will need a lot more rewrite. 
we have delete in webserve, webapi. and template must pass the show to webserve.
there is must be a reason for that code not being at show_queue

WebSpider:
Alternatively, when deleting a show, delete show related queue items first, and then actually delete the show itself?

-----Original Message-----

fernandog:
No. Just tell user to try again later.
I still couldnt reproduce because the delete button gets disabled in show page

I need to try with mass delete and webapi

fernandog:
@miigotu 
Can't delete via API or mass update (shows no message, click the button and nothing happens). Can you try please?
So probably our theory was wrong in that issue

/api/******/?cmd=show.delete&indexerid=248962&removefiles=0
{
    "data": {}, 
    "message": "Show can not be deleted while being added or updated", 
    "result": "failure"
}

fernandog:
can't find the code that hides buttons in show page when full-updating

fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/1291

miigotu:
@fernandog replacing this with a proper queue item.

PR #2280, opened 2015-08-09
Starting sql and js move from mako

OmgImAlexis:


OmgImAlexis:
Does [this](https://github.com/OmgImAlexis/SickRage/blob/move-sql-js-from-mako/gui/slick/interfaces/default/config_notifications.mako#L1457) need to be removed? It's currently commented out.

OmgImAlexis:
Use ``'', 'checked="checked"'`` not ``"", "checked=\"checked\""`` so it doesn't have to be escaped.

OmgImAlexis:
The last few lines of inc_bottom need looking at they have a ``## FIXME`` on them.

miigotu:
I'll look at that inc_bottom tonight.

miigotu:
Merged in #2284

PR #2283, opened 2015-08-09
Added support for support HEVC (x265)

ipsq:
by changing all 264 with (264|265)

### DIFF
--- a/sickbeard/common.py
+++ b/sickbeard/common.py
@@ -232,27 +232,27 @@ class Quality:
             else:
                 return Quality.UNKNOWN
 
-        if checkName(["(pdtv|hdtv|dsr|tvrip).(xvid|x264|h.?264)"], all) and not checkName(["(720|1080)[pi]"], all) and\
-                not checkName(["hr.ws.pdtv.x264"], any):
+        if checkName(["(pdtv|hdtv|dsr|tvrip).(xvid|x(264|265)|h.?264)"], all) and not checkName(["(720|1080)[pi]"], all) and\
+                not checkName(["hr.ws.pdtv.x(264|265)"], any):
             return Quality.SDTV
-        elif checkName(["web.dl|webrip", "xvid|x264|h.?264"], all) and not checkName(["(720|1080)[pi]"], all):
+        elif checkName(["web.dl|webrip", "xvid|x(264|265)|h.?(264|265)"], all) and not checkName(["(720|1080)[pi]"], all):
             return Quality.SDTV
-        elif checkName(["(dvdrip|b[rd]rip|blue?-?ray)(.ws)?.(xvid|divx|x264)"], any) and not checkName(["(720|1080)[pi]"], all):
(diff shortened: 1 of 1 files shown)


ipsq:
Wrong commit message, sorry for that.

ipsq:
Should I also add HEVC itself to the list? Some release groups prefer to use HEVC over x265 so this won't pick those up.

miigotu:
@ipsq we cant add HEVC/H265 this way, it would need it's own quality. We have discussed it and the biggest problem is that many users use raspberry pi/pi2 and support for hevc/x265 is limited. IE: they need to be able to distinguish HD/FULLHD from HDHEVC/FULLHDHEVC

ipsq:
But SickRage already downloads files with x265 so it would only be an indicator inside SickRage itself. I don't get why this would  reduce compatibility with Raspberry Pi?

miigotu:
Most embedded devices cannot play hevc well. If you change it this way there is no indication in the gui that it is an HEVC video.

Lets see what some others think about this first.

fernandog:
@miigotu I add x265 and hevc to global ignore words to don't download them.

if this PR only adds the possibility to "read" files with x265 I guess it's no a problem.

We can add an option (default off) that auto adds x265 and hevc to global ignore words.
I don't see we rewriting all qualities to allow/disallow x265. Instead we should filter it.

miigotu:
Ok, I just checked too that hevc gets downloaded but just detected as the wrong quality. Still don't like it that we can't distinguish easily.

I'm worried that something will break =P so if something goes boom I kill you! lol

miigotu:
Create PR to develop -.-

fernandog:
@miigotu wait. just my opinion lol. don't need to merge. we were discussing lol

fernandog:
@miigotu there's no need to point that file is x264 or x265. we don't do that with xvid. why do that with x265

ipsq:
@fernandog My SickRage client did download a file containing x265?

fernandog:
@miigotu or do you think we should add codec in quality? Like 1080p.HDTV.x264 and 1080p.HDTV.x265 ?

In rpi 1080p x265 dont play well. Even 720p

ipsq:
Shouldn't this become an option then? There are a lot of people who **don't** run SickRage on an RPi.

fernandog:
SR downloads OK x265 files.

miigotu:
@ipsq its not about what you run SR on, its about what you play your media on, and I'm better half our users play media on a raspberry pi, pi2 or other embedded board.

We aren't saying that something should not be done, we just have been discussing options for a proper solution. Usually the first most obvious and easy "workaround" is not the best route, and it is better to make a well thought out decision.

It just needs incubated/discussed a bit

piejanssens:
IMO adding the codec to the quality is going to get messy.
Getting to a point where x265 is widely adopted and compatible (like x264 is today) will take some time.
Just a simple "Enable HEVC/x265 releases" checkbox in the Episode Search Settings is easy to configure, easy to find for the end user and doesn't require any filtering words.

miigotu:
Need to have this now, with the hachoir usage now x265 was getting detected as unknown.

PR #2293, opened 2015-08-12
Update thepiratebay.py

creamers:
Corrected Root TLD

### DIFF
--- a/sickbeard/providers/thepiratebay.py
+++ b/sickbeard/providers/thepiratebay.py
@@ -59,7 +59,7 @@ class ThePirateBayProvider(generic.TorrentProvider):
 
         self.cache = ThePirateBayCache(self)
 
-        self.urls = {'base_url': 'https://thepiratebay.gd/'}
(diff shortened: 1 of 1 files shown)


miigotu:
1. you never create a PR to master
2. the gd domain is correct.

creamers:
The GD domain is a redirect, according to my sickrage logs when ever it tries to query the GD domain it files. I corrected the domain on my local install and everything is working.

As for commiting wrong, I have never used GIT I am just now learning it.

Arcanemagus:
Guidelines for contributing can be found in [contributing.md](https://github.com/SiCKRAGETV/SickRage/blob/master/contributing.md).

miigotu:
@creamers it doesnt matter which domain you set it to, it will redirect based on the closest dowmain depending on which country you are in, to one that is not blocked. .gd is used for north america.

creamers:
@miigotu true just seems odd to me that my logs no longer have errors from TPB once the url is changed.

miigotu:
try using .se, as that is the main redirect domain.

PR #2296, opened 2015-08-13
Fixed invalid syntax error when starting sickrage

roh85:
_camel_word2_set is a dictionary but uses incorrect syntax, comma to
seperate elements instead of ':'. Changed the dict to a list. because I dont think a dict is needed here.

### DIFF
--- a/lib/guessit/textutils.py
+++ b/lib/guessit/textutils.py
@@ -247,8 +247,8 @@ def find_first_level_groups(string, enclosing, blank_sep=None):
     return split_on_groups(string, groups)
 
 
-_camel_word2_set = {'is', 'to'}
-_camel_word3_set = {'the'}
+_camel_word2_set = ['is', 'to']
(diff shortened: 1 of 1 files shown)


roh85:
Tested on synology DS411

WebSpider:
We may also want to send this upstream?

roh85:
The PR is now closed but it is not merged? Does this mean that the solution is not accepted? Please give some feedback.Thx!

WebSpider:
Well there are a few things here:

- The pr was based off master, Always base your pr's off develop
- The code seems to be part of the Python guessit module. It would be better to have the guys from guessit solve this
- Maybe @fernandog has more to add

miigotu:
1. SickRage only supports python 2.7., only 2.6 throws the error.
2. We do not patch libs, fixes get PR upstream.
3. We do not accept PR to master.

PR #2297, opened 2015-08-13
FIX: Subtitles Extra Scripts

eucalyptus30:
> Unable to run subs_extra_script: execv() arg 2 must contain only strings
> +ADD epObj.show.indexerid arguments


miigotu:
Next time make PR to develop or I will kill it,

PR #2300, opened 2015-08-13
Add new Deluge Daemon Client

PaulW:
As requested from originally submitted pull request https://github.com/SiCKRAGETV/SickRage/pull/2274 to submit to develop not master:

Add new client method to support directly connecting to the Deluge Daemon service, without the requirement to go via the WebUI.

Existing WebUI Client method is still retained, so the option for either still exists.

Helpful for fully headless clients, or minimal installs of Deluge, or where access via WebUI is not available.


miigotu:
@PaulW Perfect, thanks!

PR #2301, opened 2015-08-14
Closes SiCKRAGETV/sickrage-issues/issues/2461

miigotu:


sickbeard/common.py
```diff
@@ -40,6 +40,8 @@
               ]
 
 INSTANCE_ID = str(uuid.uuid1())
+if len(sickbeard.CUR_COMMIT_HASH) == 40:
+    INSTANCE_ID += '-' + sickbeard.CUR_COMMIT_HASH[:7] # Append github style short hash
```
WebSpider: sickbeard config variables aren't available here yet, if i read the code correctly?

miigotu:
@WebSpider yes I saw that, and I cant even get them to there either....

WebSpider:
Yeah i was having similar troubles with causing circular imports..

miigotu:
I think this might be because of sickbeard/__init__.py importing like `from sickbeard import` where is should be `from . import` or just `import `

miigotu:
eh, because of `__init__` importing from common. maybe need to mod the USER_AGENT later, this one is a pita.

PR #2303, opened 2015-08-14
Adds option to stop daily execution of show updates

WebSpider:
Closes SiCKRAGETV/sickrage-issues#1626

Combined with Update on Startup/Snatch this prevents daily calls to indexers while most shows don't change daily.

### DIFF
--- a/gui/slick/interfaces/default/config_general.mako
+++ b/gui/slick/interfaces/default/config_general.mako
@@ -103,6 +103,16 @@
                             </label>
                         </div>
                         <div class="field-pair">
+                            <label for="no_showupdate">
+                                <span class="component-title">Disable automatic show updates</span>
+                                <span class="component-desc">
+                                    <input type="checkbox" name="no_showupdate" id="no_showupdate" ${('', 'checked="checked"')[bool(sickbeard.NO_SHOWUPDATE)]}/>
+
+                                    <p>shows will not get updated automatically, ie. we will not find new seasons/episodes as they air. Makes next option on timing irrelevant. 
+                                </span>
+                            </label>
+                        </div>
+                        <div class="field-pair">
                             <label for="showupdate_hour">
                                 <span class="component-title">When to update shows</span>
                                 <span class="component-desc">
--- a/sickbeard/__init__.py
+++ b/sickbeard/__init__.py
@@ -90,7 +90,7 @@ events = None
 # github
 gh = None
 
-# schedualers
+# schedulers
 dailySearchScheduler = None
 backlogSearchScheduler = None
 showUpdateScheduler = None
@@ -179,6 +179,7 @@ ACTUAL_CACHE_DIR = None
 ROOT_DIRS = None
 UPDATE_SHOWS_ON_START = False
 UPDATE_SHOWS_ON_SNATCH = False
+NO_SHOWUPDATE = False
 TRASH_REMOVE_SHOW = False
 TRASH_ROTATE_LOGS = False
 SORT_ARTICLE = False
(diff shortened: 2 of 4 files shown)


miigotu:
Update on snatch will be going away soon. Maybe also changing this to be able to set > 24 hours between show updates is better?

miigotu:
Actually, setting it up so that each show updates a few hours before each airdate/time would be the best solution. It solves the episode naming problems, and weekly shows would get updated once a week, daily once a day, and the load/strain on SR server and meta indexers would be spread out over the day.

The option for when to update shows then becomes "Update show x number of hours before airtime" minumum 1 hour, maximum 24 hours.

Then I think update/refresh on startup could be taken away on top of the "update on snatch" option also.

You think you could make all of that happen? It is hands down the best solution, and removes at least 2 settings, reduces tvdb/tvrage strain, and still provides the benefits of the other options (and works better anyways).

fernandog:
but how this would work when new season or new show has no airdate yet? how to update it?
what about shows that have a season split in two like TWD ?

miigotu:
@fernandog we already cache what info there is. This is just about show updater which runs every day now, and does not need to because most shows only have a new episode or are changed once a week. If it has no airdate yet, we just update that show everyday, as it will surely be a small transfer.

I want rid of "update on snatch" and to spread the load around. My suggestion gives a real solution for all of the problems we've had previously, and eliminates some unnecessary configuration while not working the indexers as hard.

miigotu:
Basically this would make it work like for example, you have 42 shows, 6 air each day, only 6 shows are updated each day instead of 42, and they are not updated back to back but throughout the day depending on the time of day they air.

fernandog:
Ok!

WebSpider:
@miigotu yeah that's what I was thinking when entering the FR on daily search intelligence as well. May make it into something event-based instead of timer nased

WebSpider:
*based

neoatomic:
I see some problems with this.
It happens often that for a new seasons a rumored/fictional date is added. 
Lets say they set the date 31 December and some time later they change it to 1 August as official air-dates gets released. SR wont be aware of this change if there are no frequent updates.

Removing `update/refresh on startup` i dont think is a good idea. We have users that dont run SR 24/7 and for them to have that option prevents outdated air-dates etc.

miigotu:
@neoatomic So we can set the max time in between refreshes to maybe 2 weeks for each show. Most shows are bi-weekly or less, and most shows with a phony airdate should have their airdate updated with a real one at least 2 weeks before the real airdate.

@WebSpider so if airdate is unknown/1/never update every 2 weeks, if airdate is > 2weeks in the future, update every 2 weeks. Otherwise update the show a set x number of hours before air time.

neoatomic:
Find 2 weeks to long... If we do the update every 2 days it would be a reduction of 50% of the load. if we do every 3 days its 75%. Every thing more than that doesn't give a real advantage...

Also if we can do a check if an update has run/occurred within 24 hours an update on startup is blocked/cant run. That should prevent users from updating x times a day if they install updates/restart with the function "update on startup" enabled. 
Think that would be a better solution.

WebSpider:
Don't forget update on startup interaction and rate of updates in master.

In reality, you will most likely restart SR once a week...

-----Original Message-----

neoatomic:
Well if you use a NAS/Server i agree.:) But dont forget the Windows users, they start SR on demand. So that can be multiple times a day. And they generally use "update on startup" to be sure they are up to date.

Also you have talk shows/sports etc that air daily, and i dont know what consequences it will have for those. Thats why i have some concerns, think its best not to rush this, but give it some more thoughts.

miigotu:
"Also you have talk shows/sports etc that air daily," Exactly why each show updates specifically a few hours before each airdate. If it is a daily show SR knows this already, and would update it a few hours before airtime each day. If it is a weekly show, Sr knows this and updates it a few hours before the time it airs each week. If it is bi-weekly, same thing. If it is more than 2 weeks in the future, we will just check it every two weeks in case the airdate changes on the meta indexer.

We have all of the info about when a show airs and how often it airs, and when we last updated shows in db. There is no sense updating a show every day or 2 or 3 when we know it doesn't air again for another month.

miigotu:
Changes like this are exactly the kind of things I have been preaching about with removing unnecessary config options. We can remove 2 config options and improve the behavior with a little creative thinking.

If 1 request can be avoided by doing something a little different, than we aren't doing it correctly yet.

Update for one show has to be at least 10 requests (I havent counted) but is probably many more.
Times that by 50 shows average and you have 500+ requests per show update queue as it currently sits. Add in a restart or a few snatches and you can multiply that by say 10x a day. 5000 requests.

With my proposed behavior, not only does it remove the 2 extra config options, show update queue blocking search/snatch while waiting for show update to finish, issue deleting show while it is being updated, 12 hour show update queue etc, you can divide the number of requests to the indexer by at least 100, and have maybe 50 requests to the meta indexer a day.

flightlevel:
I'm no developer but don't TVDB and TVRage API both let you enter a time and it returns the shows that have had changes since the that time. This way you'd only ever update shows that have changed. Save the last time you checked and you could even update more often, as the majority of the time your shows probably wouldn't have a change. 

The methods above are guessing when you might have a change, this way you only update when a change occurs and it would get to SR faster than waiting a few hours before airtime (makes coming episodes work better the earlier you know)
http://thetvdb.com/wiki/index.php?title=Programmers_API
eg.
http://thetvdb.com/api/Updates.php?type=all&time=1439721344

neoatomic:
@miigotu Getting a better picture of how you want to implement it now. And that could work.

@flightlevel Constructive idea's/comments are alway more than welcome. And it might not even be a bad idea to look into this feature.

miigotu:
@WebSpider and I will be working on switch over to api based show queue updates. It is the better option, and some parts from the other ideas may make it in. Switching will fix hella issues and FR anyways.

PR #2322, opened 2015-08-18
Dont allow forcing daily/backlog search on startup

miigotu:
forcing searches on startup is no good, and defeats the purpose of a timer between searches. We must set the interval in relation to last daily/backlog time set in the database instead.


miigotu:
run_delay may need removed or set by last_backlog if this doesnt work efficiently, but backlog on startup and daily on startup are BAD ideas when considering strain on indexers and providers both.

PR #2323, opened 2015-08-18
Set up cache control for static content (images)

WebSpider:
We use Tornado's default v= cache breaker.
This will instruct the browser to cache content until we change commit hash or 10 years, whichever comes first


miigotu:
Going to merge this until we have a better solution worked out.

PR #2324, opened 2015-08-19
Changing from tmpl to mako

OmgImAlexis:


OmgImAlexis:
A lot of testing needs to be done to make sure I didn't accidently remove anything from the theme.

@Goeny do you remember which files you changed so I can just replace the ones you didn't with already converted mako files?

Goeny:
All tmpl files were changed.

Also some js files were changed, and some new js libraries were added. (For the full page backgrounds -> backstretch, For the scrollbars -> mCustomScrollbar, For the coverflow -> coverflow.js, i think i used raterater,js for the star rating, and i am pretty sure i edited scripts.js)

In sickrage itself was only 1 minor change, but that was merged a few months ago. (sickbeard/webserve.py)

Sorry if i can't be of any more assistance. Just too busy busy busy with other stuff.

OmgImAlexis:
@Goeny this may need to be completely rewritten I don't think a conversion will work at all now especially after https://github.com/SiCKRAGETV/SickRage/pull/2403

PR #2333, opened 2015-08-22
Let tornado handle etags for browser caching~

miigotu:
Still is a 302 redirect for poster and probably network icon, however it is redirecting to a cached image.


miigotu:
@OmgImAlexis this makes a huge difference. using etags as cache breaker, built into tornado.

Which method would you use for lazy loading of the images?

OmgImAlexis:
I'll be on IRC later. I'm thinking just some jQuery plugin that loads them as soon as the DOM is finished since we already know their size that'll be fine.

miigotu:
@OmgImAlexis atm with this PR, DOM is taking 424ms and full load is 450ms total, locally. I still see the blip of whacky styling.

OmgImAlexis:
Currently going to the Apple Store. When I get back I want to try something. I'm pretty sure with lazy loading we can get the style sheet to load before the Dom finishes load so it styles as its rendered.

PR #2339, opened 2015-08-24
Add AmActive to schedulers

fernandog:
Obs: Subtitle scheduler has def __init__(self)

https://github.com/SiCKRAGETV/sickrage-issues/issues/2552

### DIFF
--- a/sickbeard/subtitles.py
+++ b/sickbeard/subtitles.py
@@ -125,6 +125,8 @@ class SubtitlesFinder():
     """
 
     def run(self, force=False):
+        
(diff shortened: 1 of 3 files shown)


fernandog:
@miigotu This will fix the "N/A" in server status

image

fernandog:
@miigotu didn't fix subtitles finder. Can't understand why.

miigotu:
You need to add an `__init__()` that initializes self.amActive = False for subtitle finder @fernandog

miigotu:
Going to do it now...

PR #2340, opened 2015-08-24
Exit function properly on exception

Hiltronix:
When "metadata.extract(parser)" raises an exception, the function exits without returning "None", which misleads the calling function that it exited with usably data, and it then interprets the result incorrectly.

The end result is that image_cache.py > fill_cache(self, show_obj) exits early on an exception when attempting to retrieve images for the cache from a meta data provider, instead of continuing to the second step of attempting to retrieve images for the cache from the indexer online.  

Without this fix the log gets the generic error message "bad descriptor" for an unknown error from "metadata.extract(parser)", and the secondary cache image retrieval attempt from the indexer is never made.

### DIFF
--- a/lib/hachoir_metadata/metadata.py
+++ b/lib/hachoir_metadata/metadata.py
@@ -284,6 +284,7 @@ def extractMetadata(parser, quality=QUALITY_NORMAL):
         metadata.extract(parser)
     except HACHOIR_ERRORS, err:
         error("Error during metadata extraction: %s" % unicode(err))
+        return None
     except Exception:
         return None
     if metadata:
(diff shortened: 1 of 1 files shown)


fernandog:
Why not return none in the previous exception?

Hiltronix:
I tried that, it didn't solve the problem.  The previous exception is targeting the specific HACHOIR_ERRORS errors, which aren't the problem when any other exception is raised.  Which was the case in my testing.  When an exception is raised that doesn't fall into HACHOIR_ERRORS then the function will not exit with None.

Hiltronix:
Quite honestly I personally don't see the value to the exception handling of:

    except HACHOIR_ERRORS, err:
        error("Error during metadata extraction: %s" % unicode(err))

as these errors are not being generated in the sickrage.log file.  I don't know where they are being written to, but it isn't the main log.

miigotu:
This cannot be merged here. It must be submitted to the hachoir repo.
https://bitbucket.org/haypo/hachoir

Hiltronix:
@miigotu  My PR has been merged on hachoir.  Does this hachoir library get updated into SR on each update you put out?

miigotu:
It gets updated periodically, or when I am notified that there is a beneficial change for us in it.

I will update the library shortly.

PR #2346, opened 2015-08-25
Initialize amActive for SubtitlesFinder

miigotu:
@fernandog 
SiCKRAGETV/SickRage/pull/2339


fernandog:
Should we move amactive after 'if not sickbeard.USE_SUBTITLES' ?

miigotu:
Um yes, otherwise the thread will show as active forever after the first run when users don't have subtitles enabled.

PR #2348, opened 2015-08-25
Fix imdb popular

jonathanendersby:
Better handling of exceptions.
Added caching for Images. (bypasses referrer banning).
Converted imdbPopular into a class.
Small style changes.
Small bug fixes.

### DIFF
--- a/gui/slick/interfaces/default/home_popularShows.mako
+++ b/gui/slick/interfaces/default/home_popularShows.mako
@@ -2,14 +2,20 @@
 
 <%include file="/inc_top.mako"/>
 
+<h2>Popular Shows</h2>
+<br />
+
 % if not popular_shows:
     <h3>Fetching of IMDB Data failed. Are you online?</h3>
+
+    <strong>Exception:</strong>
+    <p>${imdb_exception}</p>
 
 % else:
     % for cur_result in popular_shows:
         <div class="popularShow">
             <div class="left">
-                <img class="coverImage" src="${cur_result['image_url_large']}" />
+                <img class="coverImage" src="${sbRoot}/cache/${cur_result['image_path']}" />
             </div>
             <div class="right">
                 <h3>${cur_result['name']}</h3>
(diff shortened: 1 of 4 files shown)


miigotu:
https://github.com/SiCKRAGETV/SickRage/commit/28d7f48ff513ae739aebc337ce0005f76f9a4acd

fernandog:
@jonathanendersby @miigotu too complex to check if show already is in SR and not suggest it or disable the "add" button ?

WebSpider:
Have to agree here, seeing shows in the overview that are already in SR is a bit annoying.

fernandog:
I like to see my shows are 'popular' - others may not
But there is no point to show add button

jonathanendersby:
It's going to be quite tricky (maybe even impossible) to 100% accurately match up show names (ie strings) with TVDB/TVRage IDs in order to compare the popular list to which shows you already follow on SR. Basically we'd have to do a search against TVDB and the guess which show from the results it the right one... 

That all said, we could decide that if the two name strings (IMDB vs SR) match perfectly, then we grey out that show... But I don't think IMDB follows Scene naming conventions closely enough for this not to get confusing.

Thoughts?

miigotu:
I think its more processing and potential bugginess to attempt to modify the list.  We return the list just as it is on IMDB and there is no problem with that, you can compare how many shows on that list you have vidually.

fernandog:
@jonathanendersby 
all shows have the imdb_id in the database (tv_shows table). Just use that.

miigotu:
@fernandog isn't it more important to get it working without a 20 seconds page load than to worry about disabling a button?

jonathanendersby:
I'll check if the ID in the URL is the actual IMDB ID, and if it is I'll see what I can do re greying out buttons...

But I agree with @miigotu that the initial cache delay is painful and feels broken. Can we not add a call to  imdbPopular.__init__() at startup? Is there a thread for Miscellaneous?

miigotu:
@jonathanendersby there is not currently.

PR #2350, opened 2015-08-25
Remove obsolete Google Code and Windows binary distributions

WebSpider:
Google Code is dead.
So is binary distributions.


miigotu:
@WebSpider sweet, have been meaning to do this, and maybe try and find a way to move the majority of SickBeard.py into the module to it is properly distributable...

Will need to write a proper setup.py sometime maybe xD.

PR #2351, opened 2015-08-26
Fixed up deluge daemon handling of torrents that are already added.

jalim:
Will resume a paused torrent if that is the default.


miigotu:
This adds unused imports, print calls, and unpauses torrents that the user has manually paused. Cannot accept this.

PR #2353, opened 2015-08-26
Fixed up deluge daemon handling of torrents that are already added.

jalim:
Removed superfluous code that was duplicating hash extraction.
Cleaned up unused imports & print calls.

### DIFF
--- a/sickbeard/clients/deluged_client.py
+++ b/sickbeard/clients/deluged_client.py
@@ -1,4 +1,5 @@
 # Author: Paul Wollaston
+# Contributions: Luke Mullan
 #
 # This client script allows connection to Deluge Daemon directly, completely
 # circumventing the requirement to use the WebUI.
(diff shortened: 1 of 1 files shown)


miigotu:
We don't always have the torrent hash either.

jalim:
The code in _check_torrent was basically identical to that in generic.py (_get_torrent_hash) which adds the hash to the result before sending it to the client script.  If there's no hash found by generic.py then _check_torrent isn't going to have any more luck as far as I can tell and it doesn't make sense to repeat this.  Unless I'm missing something?

miigotu:
AFAIK get_torrent_hash is only for when we have a magnet only. If we have a URL we don't have the hash.

jalim:
I might be wrong but it looks to me that it is called every time we send a torrent to the client regardless of whether it is for a magnet or a torrent url.  Obviously it has two different methods to find this hash depending on whether it is a magnet or torrent file but then it always appends a hash to the result before this result is sent to the client script.

miigotu:
@jalim I apoligize. Just looked into the code instead of from memory and you are correct.

I will review your PR again.

PR #2356, opened 2015-08-27
Fix for SiCKRAGETV/sickrage-issues#2578: Add a "Size" column on shows list

MGaetan89:
It needs to be tested for potential performance issue before being included in `develop`/`master`.
I only have 2 shows with files in my dev environment, so I am not sure about the result.

### DIFF
--- a/gui/slick/views/home.mako
+++ b/gui/slick/views/home.mako
@@ -126,14 +126,14 @@ $(document).ready(function(){
     });
 
     $("#showListTableShows:has(tbody tr)").tablesorter({
-        sortList: [[6,1],[2,0]],
+        sortList: [[7,1],[2,0]],
         textExtraction: {
             0: function(node) { return $(node).find("span").text().toLowerCase(); },
             1: function(node) { return $(node).find("span").text().toLowerCase(); },
             3: function(node) { return $(node).find("span").prop("title").toLowerCase(); },
             4: function(node) { return $(node).find("span").text().toLowerCase(); },
             5: function(node) { return $(node).find("span:first").text(); },
-            6: function(node) { return $(node).find("img").attr("alt"); }
+            6: function(node) { return $(node).data('show-size'); },
             7: function(node) { return $(node).find("img").attr("alt"); }
         },
         widgets: ['saveSort', 'zebra', 'stickyHeaders', 'filter', 'columnSelector'],
@@ -218,7 +218,7 @@ $(document).ready(function(){
     });
 
     $("#showListTableAnime:has(tbody tr)").tablesorter({
-        sortList: [[6,1],[2,0]],
+        sortList: [[7,1],[2,0]],
         textExtraction: {
             0: function(node) { return $(node).find("span").text().toLowerCase(); },
             1: function(node) { return $(node).find("span").text().toLowerCase(); },
--- a/gui/slick/views/home.mako
+++ b/gui/slick/views/home.mako
@@ -134,6 +134,7 @@ $(document).ready(function(){
             4: function(node) { return $(node).find("span").text().toLowerCase(); },
             5: function(node) { return $(node).find("span:first").text(); },
             6: function(node) { return $(node).find("img").attr("alt"); }
+            7: function(node) { return $(node).find("img").attr("alt"); }
         },
         widgets: ['saveSort', 'zebra', 'stickyHeaders', 'filter', 'columnSelector'],
         headers: {
@@ -143,7 +144,7 @@ $(document).ready(function(){
             4: { sorter: 'quality' },
             5: { sorter: 'eps' },
             % if sickbeard.FILTER_ROW:
-                6: { filter : 'parsed' }
+                7: { filter : 'parsed' }
             % endif
         },
         widgetOptions : {
@@ -224,7 +225,8 @@ $(document).ready(function(){
             3: function(node) { return $(node).find("span").prop("title").toLowerCase(); },
             4: function(node) { return $(node).find("span").text().toLowerCase(); },
             5: function(node) { return $(node).find("span:first").text(); },
-            6: function(node) { return $(node).find("img").attr("alt"); }
+            6: function(node) { return $(node).data('show-size'); },
+            7: function(node) { return $(node).find("img").attr("alt"); }
         },
         widgets: ['saveSort', 'zebra', 'stickyHeaders', 'filter', 'columnSelector'],
         headers: {
@@ -234,7 +236,7 @@ $(document).ready(function(){
             4: { sorter: 'quality' },
             5: { sorter: 'eps' },
             % if sickbeard.FILTER_ROW:
-            6: { filter : 'parsed' }
+                7: { filter : 'parsed' }
             % endif
         },
         widgetOptions : {
@@ -604,6 +606,7 @@ $(document).ready(function(){
             <th>Network</th>
             <th>Quality</th>
             <th>Downloads</th>
+            <th>Size</th>
             <th>Active</th>
             <th>Status</th>
         </tr>
@@ -612,6 +615,7 @@ $(document).ready(function(){
     <tfoot class="hidden-print">
         <tr>
             <th rowspan="1" colspan="1" align="center"><a href="${sbRoot}/home/addShows/">Add Show</a></th>
+            <th>&nbsp;</th>
             <th>&nbsp;</th>
             <th>&nbsp;</th>
             <th>&nbsp;</th>
@@ -775,6 +779,15 @@ $(document).ready(function(){
             <div class="progressbar hidden-print" style="position:relative;" data-show-id="${curShow.indexerid}" data-progress-percentage="${progressbar_percent}" data-progress-text="${download_stat}" data-progress-tip="${download_stat_tip}"></div>
             <span class="visible-print-inline">${download_stat}</span>
         </td>
+
+        <% show_size = 0 %>
+        % try:
+            <% show_size = sickbeard.helpers.get_size(curShow.location) %>
+        % except sickbeard.exceptions.ShowDirNotFoundException:
+            <% show_size = sickbeard.helpers.get_size(curShow._location) %>
+        % endtry
+
+        <td align="center" data-show-size="${show_size}">${sickbeard.helpers.pretty_filesize(show_size)}</td>
 
         <td align="center">
             <% paused = int(curShow.paused) == 0 and curShow.status == 'Continuing' %>
--- a/gui/slick/views/home.mako
+++ b/gui/slick/views/home.mako
@@ -780,13 +780,7 @@ $(document).ready(function(){
             <span class="visible-print-inline">${download_stat}</span>
         </td>
 
-        <% show_size = 0 %>
-        % try:
-            <% show_size = sickbeard.helpers.get_size(curShow.location) %>
-        % except sickbeard.exceptions.ShowDirNotFoundException:
-            <% show_size = sickbeard.helpers.get_size(curShow._location) %>
-        % endtry
-
+        <% show_size = sickbeard.helpers.get_size(curShow._location) %>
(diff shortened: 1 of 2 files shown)


MGaetan89:
Just noticed that the commit message is wrong. I will fix it later.

OmgImAlexis:
The load times seem okay, just tried this on my install with about 100 shows totally 4TB using a SMB mount and it loaded maybe 300ms slower.

MGaetan89:
Good to know. The overhead will be gone once the total size is saved in the DB.

OmgImAlexis:
Since we've removed about 500ms this week, I don't think that's too bad and you're right the db will help a LOT.

fernandog:
Fix size sorting and lets merge to make some tests

OmgImAlexis:
@MGaetan89 Did you want to work on the sorting for this or should I do it?

OmgImAlexis:
You also need to add `align="center"` to the td.

MGaetan89:
I can do it. I will work on it after work ;)

MGaetan89:
Commit message is fixed, and the cell is centered. I'm looking at the sort issue now.

MGaetan89:
The sort should be fix.
By doing that I think I have found a cleaner solution to sort the progress (https://github.com/SiCKRAGETV/SickRage/commit/3d2f382c1423e3ac5d06c30e88ea6a6813d6272d). I will investigate that.

fernandog:
Active column  swapped with size

MGaetan89:
Yes, because the "Size" column is before the "Active" column

fernandog:
@MGaetan89 

contents swapped
image

fernandog:
and still not working the sorting. related?

MGaetan89:
O.o I don't have that issue on my side. I tried both themes (`light` and `dark` to be sure), deleted cache, and enabled the filter row. In all cases it works fine.
I'm not sure where this might come form...

fernandog:
Will try delete cache

fernandog:
Deleted all /cache except image.
Git reset
Apply  .patch
Still the same

@miigotu any idea?

fernandog:
Also deleted browser cache

MGaetan89:
Same for me (about browser cache).
Which browser are you using? I tried with the latest version of Chrome.
Do you have some specific option enable maybe?

fernandog:
Latest chrome. Will try with windows when I get home

miigotu:
No one has some videos on fat32 partition to test this with?
Also need to test the "free disk space" feature of the status/info page with a path on fat32 specifically.

miigotu:
I see two js bugs xD.

OmgImAlexis:
@MGaetan89 if you're still looking for a cleaner way of doing https://github.com/SiCKRAGETV/SickRage/commit/3d2f382c1423e3ac5d06c30e88ea6a6813d6272d I'd suggest you look at the tablesorter and change where it's getting the text from so instead of something like this `5: function(node) { return $(node).find("span").text(); },` it'd be `5: function(node) { return $(node).find("span").eq(1).text(); },` so it selects the 2nd span.

OmgImAlexis:
I'm just taking a guess here but @fernandog's issue may be from only selecting certain columns, I haven't tried it but that's the only thing I can think of that may be causing it.

gui/slick/interfaces/default/home.mako
```diff
@@ -222,7 +224,7 @@ $(document).ready(function(){
             4: { sorter: 'quality' },
             5: { sorter: 'eps' },
             % if sickbeard.FILTER_ROW:
-            6: { filter : 'parsed' }
+                7: { filter : 'parsed' }
             % endif
         },
```
miigotu: Also add a # 6 header/sorter here ??

gui/slick/interfaces/default/home.mako
```diff
@@ -814,6 +818,15 @@ $(document).ready(function(){
             </script>
         </td>
 
+        <% show_size = 0 %>
+        % try:
+            <% show_size = sickbeard.helpers.get_size(curShow.location) %>
+        % except sickbeard.exceptions.ShowDirNotFoundException:
+            <% show_size = sickbeard.helpers.get_size(curShow._location) %>
+        % endtry
+
```
miigotu: I think you also need to check os.apth.exists here, because sometimes the show can be moved/removed or have the wrong location in db.

gui/slick/interfaces/default/home.mako
```diff
@@ -131,7 +132,7 @@ $(document).ready(function(){
             4: { sorter: 'quality' },
             5: { sorter: 'eps' },
             % if sickbeard.FILTER_ROW:
-                6: { filter : 'parsed' }
+                7: { filter : 'parsed' }
             % endif
         },
```
miigotu: Also add a # 6 header/sorter here ??

MGaetan89:
@OmgImAlexis I did it in https://github.com/SiCKRAGETV/SickRage/pull/2368 ;)
@miigotu What are the 2 JS bugs?

gui/slick/interfaces/default/home.mako
MGaetan89: No, because column 6 is the show size, and I'm just using an integer ([see here](https://github.com/SiCKRAGETV/SickRage/pull/2356/files#diff-e00affd852c5d93c6d5e2d8e92320048R216)). So I don't need to provide a sorting method.

gui/slick/interfaces/default/home.mako
MGaetan89: I took the code from [here](https://github.com/SiCKRAGETV/SickRage/blob/master/sickbeard/webserve.py#L1239) for the show location and [here](https://github.com/SiCKRAGETV/SickRage/blob/master/gui/slick/interfaces/default/displayShow.mako#L252) for the display.
So maybe that check needs to be added there too?

gui/slick/interfaces/default/home.mako
miigotu: those are used by the column selector though, and reording the columns no? I would assume this could be the cause of the issue @fernandog had with testing this, because he has different columsn selected and/or a different order of the columns.

MGaetan89: I didn't know we could change the order of the column...
But if you look a couple of lines above my change, you can see that column 3 is missing too.
Should just add `6: {}` as I don't need anything special?

gui/slick/interfaces/default/home.mako
miigotu: I think all of those are wrong, because: https://github.com/SiCKRAGETV/SickRage/blob/master/sickbeard/helpers.py#L1458 doesn't throw `sickbeard.exceptions.ShowDirNotFoundException`

gui/slick/interfaces/default/home.mako
miigotu: hmm, I thought we could change column order before. either I am mistaken or we lost that ability somehow.

what are the values in `sortList: [[6,1],[2,0]],` for?

gui/slick/interfaces/default/home.mako
MGaetan89: Ok, I will change that.

gui/slick/interfaces/default/home.mako
MGaetan89: I didn't look into it, but I guess it's the default sort: column 6 ascending or descending (not sure what 1 means), column 2 ascending or descending (the opposite of 1).
So sort by name and active status. Maybe I should change that too, as active is column 7 now?

miigotu: I think so, also with Line 117?

gui/slick/interfaces/default/home.mako
MGaetan89: The exception if thrown [here](https://github.com/SiCKRAGETV/SickRage/blob/master/sickbeard/tv.py#L189), that's what the check is for.

gui/slick/interfaces/default/home.mako
MGaetan89: Yes, will change both.

MGaetan89: Changed.

MGaetan89:
I updated the default column sort index. @fernandog can you try again to see if it works for you now?

@miigotu can you detail the two JS bugs you talked about in https://github.com/SiCKRAGETV/SickRage/pull/2356#issuecomment-135884352?

OmgImAlexis:
@MGaetan89 since I said I was going to redo a lot of the js, what are your thoughts on converting this over to using the data element for sorting? Obviously I won't change anything now but I think it'd be a better option than using this method of trying to select the correct piece of text, etc.

miigotu:
@MGaetan89 I was referring to the item index and the sortlist item.

MGaetan89:
It's a good idea. And the sort options are duplicated (one for the show table, one for the anime table). So they could be extracted in a common variable to prevent issues when a change is made.

@miigotu Ok. I thought there were something else.

gui/slick/interfaces/default/home.mako
miigotu: `if ek.ek(os.path.isdir, self._location):
            return self._location
        else:
            raise exceptions.ShowDirNotFoundException("Show folder doesn't exist, you shouldn't be using it")
`

If the exception is thrown, `curShow._location` doesnt exist.

MGaetan89: You're right! I will look into that more carefully.

miigotu: My kids have a saying... "Daddy always wins!!!" lol

Not trying to scrutinize so hard, just being super vigilant to try and keep a lease on any potential new bugs. This is especially important at this stage because I am looking to make a major release (v4.1.0) once we have a severely stable base.

That release will mark the start of allowing some major changes again.

fernandog:
working here! thanks!

MGaetan89:
@fernandog Thanks for the feedback. I will address @miigotu comment and then it should good to merge.
I don't know if someone was able to test on fat32?

gui/slick/interfaces/default/home.mako
MGaetan89: Kids are always right :)

I understand your point. What about trying to PEP8ing all the codebase too before that release? There is almost no PR left, so it shouldn't break too many things.

MGaetan89: I wasn't really sure what I was supposed to do :/
https://github.com/MGaetan89/SickRage/commit/16b0a4f2421a8b0f32fb9d14a6a498e09e8abd89
- In `home.mako` I just request the size of the show, and delegate the check if the folder exists to `get_size`
- In `helpers.py#get_size()` I check if the path exists before looping through its children

fernandog:
is it possible to check if tv-show folder is fat32, then don't calculate it ?

WebSpider:
Yes, it is. https://github.com/giampaolo/psutil to detect the file system type.

-----Original Message-----

miigotu:
@WebSpider psutil is not an option, as it requires c extensions.

MGaetan89:
Maybe the easiest solution would be to add an option to enable/disable this column? Disabled by default, user would have to enable it if they want to display it.
The downside is that the feature will be hidden and not working the same way as the other columns...

fernandog:
what about this:
http://stackoverflow.com/a/7120061

miigotu:
@fernandog that SO answer is for our hard link check, and it is not a good solution anyways for even that as it does more processing than necessary to determine if two paths are on that same partition.

@MGaetan89 it just needs tested on fat32, because I'm not sure if it was getting file size, or determining free space on fat32 that takes a hella amount of time. If it is determining free space, then the slow one would be on the status page.

MGaetan89:
I guess I can setup a disk with fat32 and some shows, but I can do that until next week-end (11/12th September).

miigotu:
Conflicts now ^

MGaetan89:
Updated :)

miigotu:
Im just gonna merge this, and if someone's install barfs Im sure we will know pretty quick.

PR #2359, opened 2015-08-28
Fixes SiCKRAGETV/sickrage-issues#2584

OmgImAlexis:
Now using class for the `progressbar` instead of id with `showid` added as a data attribute of the div.


miigotu:
@OmgImAlexis conflicts ^^

PR #2365, opened 2015-08-28
Fixes progressbar sorting

OmgImAlexis:
Since @MGaetan89 added the print view we now get back a duplicate result like `33+1 / 3733+1 / 37` instead of `33+1 / 37` so we need to split the string in half.

### DIFF
--- a/gui/slick/interfaces/default/home.mako
+++ b/gui/slick/interfaces/default/home.mako
@@ -78,6 +78,8 @@ $.tablesorter.addParser({
         return false;
     },
     format: function(s) {
+        s = s.substring(s.length/2);
+
         match = s.match(/^(.*)/);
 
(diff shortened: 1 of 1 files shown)


MGaetan89:
Didn't think about that :o
Is it a library that provide this feature or custom code? If it's custom code, we could adapt it to use only the visible text in the cell.

OmgImAlexis:
For now the split is fine, it works just as it did before. When I go back over and redo all of the js this week I'll have a look more into this.

It's the jQuery tablesort plugin and it just returns the text of the div that's why it shows up twice.

MGaetan89:
Ok, if it's working now, we can improve this later ;)

fernandog:
Thanks

PR #2367, opened 2015-08-28
Update binsearch.py

fpancis:
added alt.binaries.multimedia, only place I could find a particular show


fernandog:
don't PR to master

PR #2369, opened 2015-08-29
More specific quality pills

VinceVal:
I'm guessing most users aren't as particular as I am about their quality settings, but I like to select the specific quality I like/want for each show. The downside of this is that everything shows up as "Custom" everywhere, which is kind of annoying. This PR shows the specific custom quality that is selected, or if both resolutions of a type (HDTV, WEB-DL, BluRay) are selected, it will display a combined quality pill. For anything more complex, it reverts back to displaying "Custom".

I added a Mako %def to display these pills and tried to use it in all places where quality is displayed to reduce the amount of duplicated code.

It also renames "HDTV" to "720p HDTV" which I think makes more sense since we now have 1080p HDTV, but feel free to rip that out. :)

Thoughts/suggestions?

### DIFF
--- a/gui/slick/css/style.css
+++ b/gui/slick/css/style.css
@@ -1301,6 +1301,11 @@ td.col-status {
 	text-align: center;
 }
 
+th.col-quality,
+td.col-quality {
+	width: 110px;
+}
+
 th.col-legend,
 td.col-legend {
 	width: 80px;
@@ -1892,6 +1897,17 @@ span.quality {
 	-webkit-border-radius: 4px;
 	-moz-border-radius: 4px;
 	border-radius: 4px;
+}
+
+span.any-hd {
+	background-color: #2672b6;
+	background: repeating-linear-gradient(
+		-45deg,
+		#2672b6,
+		#2672b6 10px,
+		#5b990d 10px,
+		#5b990d 20px
+	);
 }
 
 span.Custom {
--- a/gui/slick/interfaces/default/home.mako
+++ b/gui/slick/interfaces/default/home.mako
@@ -7,6 +7,7 @@
     import datetime
     import re
 %>
+<%namespace file="/inc_defs.mako" import="renderQualityPill"/>
 <%include file="/inc_top.mako"/>
 <%
     myDB = db.DBConnection()
@@ -578,11 +579,7 @@ $(document).ready(function(){
                 </td>
 
                 <td class="show-table">
-                    % if curShow.quality in qualityPresets:
-                        <span class="show-quality">${qualityPresetStrings[curShow.quality]}</span>
-                    % else:
-                        <span class="show-quality">Custom</span>
-                    % endif
+		    ${renderQualityPill(curShow.quality, overrideClass="show-quality")}
                 </td>
             </tr>
         </table>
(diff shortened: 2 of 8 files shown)


WebSpider:
Thanks for the contribution!
Could you please squash the commits?

fernandog:
Can you send a screenshot?

miigotu:
Thanks for your time on this. However Im not sure this is effective enough. The original FR for this was to be able to define several custom quality setups and name them whatever you wanted, like in CP.

VinceVal:
Here is an example of what my qualities look like before and after:

image image

Pills are also shown on the displayShow page for custom qualities:
image

The intention of this PR is not to implement a new quality selection system, but rather to show you more information than is currently shown whenever Custom is displayed.

miigotu:
@WebSpider @MGaetan89 @OmgImAlexis what do you guys think about this PR?

I can squash and merge it myself, but im curious if any info is mis-communicated by this in the UI...

gui/slick/css/style.css
```diff
@@ -1894,6 +1899,17 @@ span.quality {
 	border-radius: 4px;
 }
 
+span.any-hd {
+	background-color: #2672b6;
+	background: repeating-linear-gradient(
```
MGaetan89: According to [Can I use](http://caniuse.com/#feat=css-repeating-gradients), one browser still needs the `-webkit-` prefix here.

OmgImAlexis:
Not sure what's up with the stripped pills but other than that it looks alright. Really I'd like to have it so people can add custom titles for their qualities but I guess this is okay for now.

MGaetan89:
I like the idea too, but as @OmgImAlexis said, it's not easy to understand what the stripped pills means.

OmgImAlexis:
It'd be nice to have the list shown on hover of custom just like when you hover the download progress bar on the main page.

miigotu:
I think the stripes are to distinguish it from the identical existing qualities, to denote it is a custom quality.

OmgImAlexis:
@miigotu since `HDTV` is changed to `720p HDTV` on this should it be change on the rest of the UI to match?

miigotu:
That will change it everywhere, even in logs.

OmgImAlexis:
Ah, didn't notice it was in `common.py`.

MGaetan89:
I'm not using "Custom" quality, but I guess people using them should understand what's going on behind that new display.

OmgImAlexis:
@MGaetan89 ALL of mine are custom since I want any HD initially and then just 1080p WEB-DL for archiving, I wouldn't be surprised if quite a few people used a similar setup to that.

gui/slick/interfaces/default/home.mako
```diff
@@ -770,11 +767,7 @@ $(document).ready(function(){
         </td>
     % endif
 
-    % if curShow.quality in qualityPresets:
-        <td align="center"><span class="quality ${qualityPresetStrings[curShow.quality]}">${qualityPresetStrings[curShow.quality]}</span></td>
-    % else:
-        <td align="center"><span class="quality Custom">Custom</span></td>
-    % endif
+        <td align="center">${renderQualityPill(curShow.quality)}</td>
```
miigotu: does this need `overrideClass="quality"` ?

miigotu:
@OmgImAlexis in your case this would still show "Custom" I think. That is the biggest drawback from going down this road, as it only handles a few cases from what I see.

miigotu:
Aha the striped pills are when the quality is mixed 720p and 1080p, since one is green and the other blue.

PR #2388, opened 2015-09-01
Move submenu js to webserve

OmgImAlexis:
@miigotu @MGaetan89 When we can, more of the javascript needs to be moved to files instead of inline and added to the inc_bottom.mako instead of the top as it'll increase performance by loading the files after the page is done rendering. This won't change any functionality.

### DIFF
--- a/gui/slick/interfaces/default/inc_bottom.mako
+++ b/gui/slick/interfaces/default/inc_bottom.mako
@@ -66,11 +66,15 @@
         <!--
         <ul style="display: table; margin: 0 auto; font-size: 12px; list-style-type: none; padding: 0; padding-top: 10px;">
             <li><a href="${sbRoot}/manage/manageSearches/forceVersionCheck"><img src="${sbRoot}/images/menu/update16.png" alt="" width="16" height="16" style="vertical-align:middle;" />Force Version Check</a></li>
-            <li><a href="${sbRoot}/home/restart/?pid=${sbPID}" class="confirm"><img src="${sbRoot}/images/menu/restart16.png" alt="" width="16" height="16" style="vertical-align:middle;" />Restart</a></li>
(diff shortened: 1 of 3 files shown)


WebSpider:
Does this relate to the issue that was reported that in some browsers the layout partially loads, and only after time it completes?

I'm in the car now, cant look up the #

-----Original Message-----

OmgImAlexis:
Yep, this should fix it. I still want to move more of the js to the bottom though but I can't do that until we get rid of all of the inline js as most of it calls scripts and that'd end up calling the scripts before they're loaded.

MGaetan89:
:+1: I'm not even sure we need JS to do that. It looks wrong to me. But it would need some big changes to update that.

WebSpider:
Its SiCKRAGETV/sickrage-issues#2381

miigotu:
The js needs moved to a separate .js file, and the browser will fix it.

OmgImAlexis:
I've moved most of the icons to the webserve function thanks to @miigotu. The only ones that're left are ones that don't use the `ui-icon` class and the logout one that I can't find in webserve.

gui/slick/interfaces/default/inc_bottom.mako
```diff
@@ -66,11 +66,15 @@
         <!--
```
OmgImAlexis: Does anyone have any idea why this commented out section is in here?

MGaetan89: I wondered that too.
I guess these options were here before being moved in the top menu.
I'm not sure if it is necessary to keep it.

MGaetan89:
:+1: This PR is great :D

gui/slick/interfaces/default/inc_bottom.mako
OmgImAlexis: If no one can find any reason for it, that's more code we can remove. 

@WebSpider @neoatomic @fernandog @miigotu

miigotu: I think it just used to be part of the submenu, before it was added to the navbar.

PR #2390, opened 2015-09-02
Add support for non-ASCII chars in search string

nesler:
A lot of content on nextgen is danish, so adding support for danish characters seems obvious.


miigotu:
I would rather see a test for unicode:

```python
if isinstance(search_string, unicode):
    search_string = search_string.encode('utf-8')
```

All of the providers will do this when we rewrite providers soonish.

miigotu:
@nesler thanks for pointing this out, am going to fix it along with the rest of the providers in https://github.com/SiCKRAGETV/SickRage/pull/2395 though.

PR #2392, opened 2015-09-03
Use getURL to do post requests, so cache control is used, sessions pe…

miigotu:
…rsist, and global settings such as allow_redirects and SSL Verify are honored. SiCKRAGETV/sickrage-issues#2532

### DIFF
--- a/sickbeard/providers/alpharatio.py
+++ b/sickbeard/providers/alpharatio.py
@@ -89,7 +89,7 @@ class AlphaRatioProvider(generic.TorrentProvider):
         self.session = requests.Session()
 
         try:
-            response = self.session.post(self.urls['login'], data=login_params, timeout=30)
+            response = self.getURL(self.urls['login'],  post_data=login_params, timeout=30)
         except (requests.exceptions.ConnectionError, requests.exceptions.HTTPError), e:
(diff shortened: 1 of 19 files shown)


fernandog:
@miigotu I reverted this and the issue about recursions is gone

miigotu:
And without people cant login to their providers.

Reverting this is not the solution, fixing the actual recursion is.

fernandog:
so how everything worked before this? I never had issue with TD and HDT.
Only one provider had issues with login (issue 2532)

miigotu:
@fernandog just because it worked for you does not mean it worked for everyone else. This is why we have an assload of issues before this for ipt and scc and TL/TD.

If the users system did not work perfect for SSL they could not log into these sites for searches.

Be a little more patient and a little less demanding geez. We already work at a hella fast pace and get A LOT of shit done quick, and this will be fixed soon enough. Reverting it doesn't make sense because then a proper fix will likely be forgotten and never get done.

fernandog:
I'm patient. This issue is from 7 days ago only.
As I said in IRC, fall season is in the corner and we need a stable master (kind of) and keep all new features in develop or feature branch and keep them there until fully tested IMHO. Otherwise we will have a LOT more issues/complains about shows not being downloaded.

PR #2394, opened 2015-09-03
Dont allow setting to bare archived status, dont allow setting failed…

miigotu:
… status in webapi if FDH is disabled.

@fernandog

### DIFF
--- a/gui/slick/interfaces/default/displayShow.mako
+++ b/gui/slick/interfaces/default/displayShow.mako
@@ -282,8 +282,8 @@ $(document).ready(function(){
         <% availableStatus = [WANTED, SKIPPED, IGNORED, FAILED] %>
         % if not sickbeard.USE_FAILED_DOWNLOADS:
         <% availableStatus.remove(FAILED) %>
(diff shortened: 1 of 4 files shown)


fernandog:
Nice.
What about users that has 'archived' only?
Will they have problems?

miigotu:
It actually does still allow bare 'archived' but the compound archived status are available to be set now which is what they need. Its a migration in progress =P.

darklow:
Now after these changes and PR i cannot use api anymore as i was using it before.
I was using web API and updating status to Archived once i have processed it on my app (after post-processing script), and now i got error:
```
param: 'status' with given value: 'archived' is out of allowed range '['wanted', 'skipped', 'ignored', 'failed']
```
I tried enabling "Use Failed Downloads - Use Failed Download Handling"

But nothing changed and i still got same error, so looks like there must be a bug.
@miigotu Could you explain what do you mean by "compound archived statuts", how can i set one?
Thanks

miigotu:
`Its a migration in progress`

PR #2395, opened 2015-09-03
Get rid of unidecode, as it is unreliable. check and decode in getURL and getRSSFeed.

miigotu:
Going to make sure we are passing unicode strings everywhere. encode to utf-8 just once in helpers.getURL and tvcache.getRSSFeed

unicode sandwiches taste better every byte.


miigotu:
need to rewrite nearly each and every url, and pass url params into getURL instead of as part of the url string. PITA, but it will be worth it to stop having to go back into providers for silly changes and some providers not wanting to log in with funky chars in PW or search for episodes with funky chars.

PR #2400, opened 2015-09-05
Added handling of seed ratio and high priority for new releases.

jalim:
Also removed some error logging that was duplicating generic.py logging.
General tidying up.


WebSpider:
Thanks for this PR!
Small question: Why are you removing error logging?

jalim:
The few error logs that I have removed weren't actually working because traceback wasn't being imported.
Also, generic.py already logs all the errors I removed logging for so it seemed to me like duplication.

PR #2401, opened 2015-09-05
Fix history limit

MGaetan89:
Fix for https://github.com/SiCKRAGETV/sickrage-issues/issues/2649

### DIFF
--- a/sickbeard/webserve.py
+++ b/sickbeard/webserve.py
@@ -443,15 +443,6 @@ class WebRoot(WebHandler):
 
         return self.redirect("/history/")
 
-    def setHistoryLimit(self, history_limit):
-
-        if not int(history_limit):
-            history_limit = 100
-
-        sickbeard.HISTORY_LIMIT = history_limit
-
-        return self.redirect("/history/")
-
     def toggleDisplayShowSpecials(self, show):
 
         sickbeard.DISPLAY_SHOW_SPECIALS = not sickbeard.DISPLAY_SHOW_SPECIALS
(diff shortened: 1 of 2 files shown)


fernandog:
it's not saving to config.ini the last history value

MGaetan89:
Was it saved before?
I don't see where that was done.

fernandog:
Webserve

MGaetan89:
I only removed [setHistoryLimit](https://github.com/SiCKRAGETV/SickRage/pull/2401/files#diff-357efeabeb6d6921e8196af70c431bebL446), and I don't see how it affects `config.ini`. Can you be more specific about what I have to change?

sickbeard/webserve.py
```diff
@@ -443,15 +443,6 @@ def setHistoryLayout(self, layout):
 
         return self.redirect("/history/")
 
-    def setHistoryLimit(self, history_limit):
```
fernandog: @MGaetan89 mako called this def to save

MGaetan89: Except for the type check, I moved that part line 3584 (just below), no? Or did I forget something?

fernandog: Line 64 of mako is not calling that def

MGaetan89: Yes, because the logic is done directly in `/history` handling.
I don't understand how that could affect saving in `config.ini` nor how it was done before. This method only sets a value, it does not save anything.

fernandog:
not sure. history is working ok but it's not saving in config.ini
@miigotu any clue?

sickbeard/webserve.py
```diff
@@ -443,15 +443,6 @@ def setHistoryLayout(self, layout):
...
-
-        if not int(history_limit):
-            history_limit = 100
-
-        sickbeard.HISTORY_LIMIT = history_limit
```
miigotu: It set it here
`sickbeard.HISTORY_LIMIT = history_limit`
this value is then reused for any page loads after that, and saved to ini when SR shuts down or anything other method saves config before then.

fernandog:
so we should force save when changing history limit?

WebSpider:
Nah, it will save on graceful shutdown

-----Original Message-----

WebSpider:
SetHistoryLimit is what i introduced to save the config value.. ;)

-----Original Message-----

MGaetan89:
So, if I understood everything correctly: `sickbeard.HISTORY_LIMIT` has to be set, and it's value is saved to `config.ini` when SR shuts down or if the save method is called somewhere else?
So this PR is good like that, as I set `sickbeard.HISTORY_LIMIT` in `History#index()`, or do I have to change something?

WebSpider:
If you set sickbeard.HISTORY_LIMIT I change of the drop down, you should be fine.

-----Original Message-----

MGaetan89:
Ok, so this should be good :)

PR #2403, opened 2015-09-05
Change mako to inheritance model

OmgImAlexis:


MGaetan89:
Adding the related issue: https://github.com/SiCKRAGETV/sickrage-issues/issues/2657

OmgImAlexis:
@MGaetan89 referenced it just above that. :tongue:

OmgImAlexis:
@WebSpider @MGaetan89 @miigotu @fernandog if any of you are touching mako files please link the pr/commit/issue here so they can be copied as this changes EVERY mako file and won't be as easy as just merging.

OmgImAlexis:
For future reference script tags don't need to have the charset declared as we already declare the page's charset as utf-8 at the top and it only needs to be declared on the script if it differs from the page. Also you'll notice with these that all scripts and css are at the top, the way the files go is like so.

- <%inherit file="/layouts/main.mako"/>
- Cached imports using <%! %>
- <%block name="scripts">
- <%block name="css">
- <%block name="content">

OmgImAlexis:
We shouldn't be using multiple document.ready statements as [they slow the page down](https://jsperf.com/docready/10) and since we've now got the scripts block we can easily remove the others that aren't needed.

miigotu:
@OmgImAlexis not sure if utf-8 even needs declared, as we render directly in unicode.

MGaetan89:
I changed `history.mako` in https://github.com/SiCKRAGETV/SickRage/pull/2401/files.
I don't know if you integrate those changes yet.

OmgImAlexis:
@miigotu like @MGaetan89 said with HTML it's usually good to declare it especially when you're using foreign languages.

@MGaetan89 I think if this can be tested I can either add those changes or someone will need to fix it when we merge this.

miigotu:
https://github.com/SiCKRAGETV/SickRage/pull/2415

PR #2406, opened 2015-09-05
Update jQuery and use cdn

OmgImAlexis:


MGaetan89:
Why not using jQuery 2.1.4 for recent browsers and jQuery 1.11.3 for older browsers?
You could also delete the local `js/lib/jquery-1.11.2.min.js` if it is no longer used.

OmgImAlexis:
@MGaetan89 just want to see if this causes any broken crap and then we'll fully switch it. Also I'm looking at removing a lot of the stuff in js/lib as we're trying to move away from bootstrap and jQuery-ui.

Also I just copied this from someone's gist so that's why it's using these specific ones.

MGaetan89:
According to the jQuery 2.0 release notes, if we support jQuery 1.9, using jQuery shouldn't be an issue. As we currently use jQuery 1.11.2, we should be good for that change.
I didn't check for 2.1, so I'm not sure about that one.

OmgImAlexis:
The speed increase in 2.x is crazy though, that's why I added the if statement. If it were up to me we'd drop support for older IEs since most people using something like SR should be using a relatively new OS but @miigotu wants older support if possible.

miigotu:
If its just loaded based on IE version (if it is IE) like here, and is using cdn, there is kinda no reason not to support the old browsers. If another situation comes up that makes it more difficult to do that then be can think about not supporting the older ones as much.

PR #2409, opened 2015-09-05
Only keep /cache/images. Delete everything else...

fernandog:
when git clean and update

https://github.com/SiCKRAGETV/sickrage-issues/issues/2661

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 #  SR User Related   #
 ######################
-/cache/
+/cache/images/
 /Logs/
 /restore/
 /backup/


WebSpider:
We should also keep provider information cached, no?

fernandog:
rss?

WebSpider:
I was more thinking of indexers .. And on my NAS install i also see subtitle caches, but those maybe from ancient times.

MGaetan89:
I think this is wrong. This will ignore `/cache/images` in Git, but will track every other files/folders in `/cache/`.
I propose the following rules, instead of `/cache/images/`:
```
!/cache/
/cache/*
!/cache/images/
```

See: http://stackoverflow.com/a/5534865/1914223

fernandog:
@MGaetan89 I tried that too. it's erasing cache folder

miigotu:
maybe
```
/cache/
!/cache/mako/
!/cache/sessions/
```

fernandog:
don't work either

miigotu:
This wont work anyways, because git will show it as untracked files then. need to delete the with code instead, when restarting from the updater.

OmgImAlexis:
You might also want to delete directories such as `cache-20150908_213838` or add them to the .gitignore, they get created when you import a backup.

PR #2412, opened 2015-09-06
Use get instead of refreshing the page

OmgImAlexis:


fernandog:
Cursor wait is to avoid user to filter other things when SR is already searching somethin

OmgImAlexis:
@fernandog this doesn't need it as it changes as soon as they select something. Why would we need to wait?

fernandog:
If user has 10 logs. 10 mega each for example. It will take time to search all logs files.

OmgImAlexis:
@fernandog my logs are around 100MBs and I have about 5 of them and it searches within a second at most. I don't see why we'd need to change the cursor. Also the cursor shouldn't be changed for things like this.

fernandog:
Ok. NP

PR #2414, opened 2015-09-06
Move logic to load images (fanart/banner/poster/network logo) into reusable classes

MGaetan89:
Working on https://github.com/SiCKRAGETV/sickrage-issues/issues/2567

This PR creates reusable classes to load a show fan art, banner, poster and network logo.
I still need to implement the part about network logo, so don't merge it yet.

The goal of this PR is also to agree on the structure. It's easier to adapt things now, rather than in two month, when everything is done.

- In my opinion, SR is mature enough to move away from the `sickbeard` folder. So I created a `sickrage` folder where I put all the new files
- I use one file per class. Coming from Java, it's an obvious choice. I don't know what is the recommended way in Python
- Instead of simply doing `import somename`, I use `from somename import somename2`. I don't know how Python handles import internally, but I guess it's better to only import what is needed?
- I never really did object oriented programming in Python, so let me know if some things are wrong or could be done differently

Let me know what you think about that and if you agree with the way I try to go with this.

### DIFF
--- a/sickbeard/webapi.py
+++ b/sickbeard/webapi.py
@@ -27,6 +27,10 @@ import re
 import traceback
 
 import sickbeard
+from sickrage.media.ShowFanArt import ShowFanArt
+from sickrage.media.ShowNetworkLogo import ShowNetworkLogo
+from sickrage.media.ShowPoster import ShowPoster
+from sickrage.media.ShowBanner import ShowBanner
 
 from versionChecker import CheckVersion
 from sickbeard import db, logger, exceptions, history, ui, helpers
@@ -80,6 +84,7 @@ result_type_map = {RESULT_SUCCESS: "success",
                    RESULT_DENIED: "denied",
 }
 # basically everything except RESULT_SUCCESS / success is bad
+
 
 class ApiHandler(RequestHandler):
     """ api class that returns json results """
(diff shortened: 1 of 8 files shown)


sickbeard/webapi.py
```diff
@@ -258,41 +262,6 @@ def filter_params(self, cmd, args, kwargs):
...
-            if which == 'poster':
-                image_file_name = cache_obj.poster_path(show)
-            if which == 'poster_thumb' or which == 'small':
-                image_file_name = cache_obj.poster_thumb_path(show)
-            if which == 'banner':
-                image_file_name = cache_obj.banner_path(show)
-            if which == 'banner_thumb':
-                image_file_name = cache_obj.banner_thumb_path(show)
-            if which == 'fanart':
-                if not cache_obj.has_fanart(show):
```
MGaetan89: This condition only exists in `webapi`, but not in [`webserve`](https://github.com/SiCKRAGETV/SickRage/pull/2414/files#diff-357efeabeb6d6921e8196af70c431bebL399). Which one is correct?
For now I chose the `webserver` part as it is probably more up to date than the `webapi` part. And I don't see why we have this check for fan art only...

miigotu: That is legacy, the cache should already load it right?

I removed it in webserve as a test to make sure, havent heard anything about that yet.

MGaetan89: Yes, it should.
From what I saw in the code, the cache should be updated when a show is added or refreshed. So we shouldn't need this.

miigotu:
Sweet, bc I think rh was sometimes undefined (according to pylint). Still working on this or?

MGaetan89:
Yes, I still want to do a couple of things in here.

MGaetan89:
Should be good now.
It support show banner, fan art, network logo and poster.

OmgImAlexis:
@MGaetan89 YOU BROKE SR. :cry: it was all pretty and now we've got no images at all.

fernandog:
just notice that lol. I use simple laytou

OmgImAlexis:
I thought my jQuery stuff broke it but maybe not...

screen shot 2015-09-07 at 2 14 10 am

fernandog:
images are still there in /cache/images

MGaetan89:
O.o will check that. Sorry for breaking everything.

miigotu:
Bad kitty. xD

OmgImAlexis:
Sure you're sorry... :crying_cat_face:

MGaetan89:
My dev environment does not have all the images for my show. Maybe that's why I did not notice it.
Is there a way to get all the show images or do I have to refresh each show one by one?

fernandog:
or copy /cache/images or use the 'mass update' feature

MGaetan89:
Thanks.
It works on the API builder but not on the Web UI.
I will investigate and submit a fix soon.

MGaetan89:
I guess Chrome was using cache data and that's why it seemed to work for me.
I found the issue. Just cleaning all up and I make a PR.

miigotu:
Yeah I didnt see it until I added a new show and then manually refreshed.

OmgImAlexis:
@MGaetan89 1. you need to get on IRC sometime 2. Did you find the issue for network icons as well?

MGaetan89:
I will try to come on IRC.
Not yet. For some reason, the `showNetworkLogo` method in `webserve` is no longer called.

PR #2422, opened 2015-09-06
Shutdown/Restart have wrong reference

MGaetan89:
I don't know how it worked until now.


MGaetan89:
I forgot some references. See https://github.com/SiCKRAGETV/SickRage/pull/2423

PR #2424, opened 2015-09-06
Only cached imports work on the whole file

OmgImAlexis:
Fixes https://github.com/SiCKRAGETV/sickrage-issues/issues/2673


fernandog:
should we have to update others .mako ?

OmgImAlexis:
@fernandog I think this is the only one with that issue and the mako files only need cached imports in some files, others work better without it. @MGaetan89, @miigotu and I will go over it when we clean up them in the next week though.

PR #2426, opened 2015-09-06
restart_bare.mako isn't used

OmgImAlexis:


fernandog:
isn't using when restarting from webapi ?

OmgImAlexis:
@fernandog I scanned the whole project and couldn't find a single reference to the bare one and all it did was include the non bare one so I'm guessing it use to be like that until someone changed it.

gui/slick/views/restart.mako
```diff
@@ -1,3 +1,67 @@
...
+        curSBHttpsEnabled = sbHttpsEnabled
+        curSBHandleReverseProxy = sbHandleReverseProxy
+        themeSpinner = sbThemeName
+    except NameMapper.NotFound:
+        curSBHost = "localhost"
+        curSBHttpPort = sickbeard.WEB_PORT
+        curSBHttpsEnabled = "False"
+        curSBHandleReverseProxy = "False"
+        themeSpinner = sickbeard.THEME_NAME
+%>
```
miigotu: this python block is in the scripts block also

OmgImAlexis: It's needed for now as I'm not sure why caching <!% only works on some stuff and not on others. If you can figure that out then we can remove it otherwise neither section seems to be able to access the stuff in there.

fernandog:
@OmgImAlexis is it possible to add the "Backup SickRage" before the "Waiting for SickRage to shut down" ?

Todays there's only a UI notification showing backup started and backup ended. But I think it's better to add this in restart mako. If backup failed. just show the "x" instead of the green "V"

What do you think?

OmgImAlexis:
I have a feeling the "Backup SickRage" stuff is done in Python and not the mako files so you'd probably be best asking @MGaetan89 or @miigotu for that.

MGaetan89:
I don't know how that part works. If I have time between two refactoring, I can try, unles @miigotu does it first.

PR #2430, opened 2015-09-06
Move history logic into a reusable class

MGaetan89:
The code to get, clear and trim the history is now the same for the web UI and the API.
From what I saw, the failed history use almost the same code. So it should be easy to change that part too.

### DIFF
--- a/gui/slick/views/history.mako
+++ b/gui/slick/views/history.mako
@@ -6,13 +6,14 @@
     import re
     import time
 
-    from sickbeard import history
     from sickbeard import providers
     from sickbeard import sbdatetime
     from sickbeard.providers import generic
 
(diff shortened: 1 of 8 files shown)


miigotu:
Now we have a separate sickrage module? xD

MGaetan89:
It was explained [here](https://github.com/SiCKRAGETV/SickRage/pull/2414).
If you don't agree with that let me know so I can move everything back into `sickbeard`.
I though it was clearer for SickRage to have a `sickrage` module, and it is also easier to see what was refactored.

miigotu:
Aha I see. I missed that comment somehow.

While it does feel a little funky having two modules at top level, I see a big benefit of having it this way for now.

As we rewrite classes we can gradually move all of the code into there, sort of like a rolling rewrite. It has been hoped for for a long time to rename the module to sickrage, but I have avoided it. This way it has added benefits rather than just renaming the module and the imports.

PR #2437, opened 2015-09-07
Temporarily disable show updates from TVRage since it's down

WebSpider:
Fixes SiCKRAGETV/sickrage-issues#2680 and related

### DIFF
--- a/sickbeard/tv.py
+++ b/sickbeard/tv.py
@@ -51,6 +51,7 @@ from sickbeard import history
 from sickbeard.blackandwhitelist import BlackAndWhiteList
 from sickbeard import sbdatetime
 from sickbeard import network_timezones
+from sickbeard.indexers.indexer_config import INDEXER_TVRAGE
 from dateutil.tz import *
 
 from sickbeard import encodingKludge as ek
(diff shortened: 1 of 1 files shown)


MGaetan89:
If it doesn't come back up, @miigotu plans to remove TVRage from SickRage (https://github.com/SiCKRAGETV/sickrage-issues/issues/2652#issuecomment-138131858).
Not sure if this PR make any sense if TVRage is going to be removed. And it would need an other update to reenable TVRage.

WebSpider:
This enables people to get regular updates from tvdb/imdb again, as the event queues are full with timing out updates ...

-----Original Message-----

miigotu:
Ok, for now this will work until wet go through the wait period (hoping tvrage comes back up)

duramato:
Are you planning on adding a replacemet for it? As i can say personally i use mainly tvdb as i like to have my posters but i depend on tvrage becouse there are a lot of shows that arent on tvdb (not the well known ones like game of thornes but a lot of the discovery shows and nat geo)

miigotu:
@duramato I guess you should get them added to tvdb then. A replacement negates the positives of being able to remove tvrage.

duramato:
@miigotu well yeah i can next time i encounter i show like that ,is that i dont find tvdb that user friedly ,as in ,i ve tryed to edit a wrong show on both tvdb and tvrage ,tvrage it got accept pretty quicly as on tvdb it didnt

PR #2438, opened 2015-09-07
Move "Coming episodes" logic into a dedicated class

MGaetan89:
I am still working on this PR. But as I found a couple of differences between the API and the web UI, I decided to open this PR to track the changes and see which version is the best and validate the choices.

- Qualities definition:
 - API ([source](https://github.com/SiCKRAGETV/SickRage/blob/v4.0.55/sickbeard/webapi.py#L736)): `qualList = Quality.DOWNLOADED + Quality.SNATCHED + [ARCHIVED, IGNORED]`
 - Web UI ([source](https://github.com/SiCKRAGETV/SickRage/blob/v4.0.55/sickbeard/webserve.py#L499)): `qualList = Quality.DOWNLOADED + Quality.SNATCHED + [ARCHIVED, IGNORED]`
 - Solution: @miigotu confirmed that `Quality.ARCHIVED` should be used (https://github.com/SiCKRAGETV/sickrage-issues/issues/2567#issuecomment-138367874)
- Sorting methods:
 - API ([source](https://github.com/SiCKRAGETV/SickRage/blob/v4.0.55/sickbeard/webapi.py#L759)): only sort by the required field
 - Web UI ([source](https://github.com/SiCKRAGETV/SickRage/blob/v4.0.55/sickbeard/webserve.py#L523)): sort by the required field and then by time
 - Solution: I chose the web UI option. However the `date` sort might need to use `airdate` rather than `localtime`. What do you think?
- Date time format:
 - API: `%Y-%m-%d %H:%M` (for `future` and `show` commands)
 - Web UI: `%Y-%m-%d %H:%M:%S`
 - Solution: Use the Web UI format (`%Y-%m-%d %H:%M:%S`) everywhere (https://github.com/SiCKRAGETV/SickRage/pull/2438#issuecomment-139912848)

The rest of the logic looks similar, except in the SQL queries. The API only select what is needed (at least it looks like it does), while the web UI selects **everything**. I will try to improve that to select only what we need.

https://github.com/SiCKRAGETV/sickrage-issues/issues/2567

### DIFF
--- a/gui/slick/views/comingEpisodes.mako
+++ b/gui/slick/views/comingEpisodes.mako
@@ -202,7 +202,7 @@ window.setInterval('location.reload(true)', 600000); // Refresh every 10 minutes
 
     <tbody style="text-shadow:none;">
 
-% for cur_result in sql_results:
+% for cur_result in results:
 <%
     cur_indexer = int(cur_result['indexer'])
     run_time = cur_result['runtime']
@@ -297,7 +297,7 @@ window.setInterval('location.reload(true)', 600000); // Refresh every 10 minutes
     <br /><br />
 % endif
 
-% for cur_result in sql_results:
+% for cur_result in results:
 <%
     cur_indexer = int(cur_result['indexer'])
 
@@ -472,7 +472,7 @@ window.setInterval('location.reload(true)', 600000); // Refresh every 10 minutes
         <thead><tr><th>${day.strftime('%A').decode(sickbeard.SYS_ENCODING).capitalize()}</th></tr></thead>
         <tbody>
         <% day_has_show = False %>
-        % for cur_result in sql_results:
+        % for cur_result in results:
             % if int(cur_result['paused']) and not sickbeard.COMING_EPS_DISPLAY_PAUSED:
                 <% continue %>
             % endif
--- a/gui/slick/views/status.mako
+++ b/gui/slick/views/status.mako
@@ -3,6 +3,7 @@
     import sickbeard
     from sickbeard import helpers
     from sickbeard.show_queue import ShowQueueActions
+    from sickrage.helper.common import dateTimeFormat
 %>
 <%block name="scripts">
 <script type="text/javascript">
@@ -111,7 +112,7 @@
                % else:
                <td></td>
                % endif
-               <td>${service.lastRun.strftime("%Y-%m-%d %H:%M:%S")}</td>
+               <td>${service.lastRun.strftime(dateTimeFormat)}</td>
                <td>${service.silent}</td>
            </tr>
            <% del service %>
@@ -159,7 +160,7 @@
                     % else:
                         <td>sickbeard.showQueueScheduler.action.currentItem.priority</td>
                     % endif
-                    <td>${sickbeard.showQueueScheduler.action.currentItem.added.strftime("%Y-%m-%d %H:%M:%S")}</td>
+                    <td>${sickbeard.showQueueScheduler.action.currentItem.added.strftime(dateTimeFormat)}</td>
                     <td>${ShowQueueActions.names[sickbeard.showQueueScheduler.action.currentItem.action_id]}</td>
                 </tr>
             % endif
@@ -191,7 +192,7 @@
                     % else:
                         <td>${item.priority}</td>
                     % endif
-                    <td>${item.added.strftime("%Y-%m-%d %H:%M:%S")}</td>
+                    <td>${item.added.strftime(dateTimeFormat)}</td>
                     <td>${ShowQueueActions.names[item.action_id]}</td>
                 </tr>
             % endfor
--- a/sickbeard/classes.py
+++ b/sickbeard/classes.py
@@ -60,11 +60,11 @@ class AuthURLOpener(SickBeardURLopener):
         # if this is the first try then provide a username/password
         if self.numTries == 0:
             self.numTries = 1
-            return (self.username, self.password)
+            return self.username, self.password
 
         # if we've tried before then return blank which cancels the request
         else:
-            return ('', '')
+            return '', ''
 
     # this is pretty much just a hack for convenience
     def openit(self, url):
@@ -189,7 +189,7 @@ class AllShowsListUI:
                         if searchterm.lower() in name.lower():
                             if 'firstaired' not in curShow:
                                 curShow['firstaired'] = str(datetime.date.fromordinal(1))
-                                curShow['firstaired'] = re.sub("([-]0{2}){1,}", "", curShow['firstaired'])
+                                curShow['firstaired'] = re.sub("([-]0{2})+", "", curShow['firstaired'])
                                 fixDate = parser.parse(curShow['firstaired'], fuzzy=True).date()
                                 curShow['firstaired'] = fixDate.strftime(dateFormat)
 
@@ -203,7 +203,7 @@ class ShowListUI:
     """
     This class is for tvdb-api. Instead of prompting with a UI to pick the
     desired result out of a list of shows it tries to be smart about it
-    based on what shows are in SB.
+    based on what shows are in SickRage.
     """
 
     def __init__(self, config, log=None):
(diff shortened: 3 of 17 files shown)


fernandog:
Will fix this?

https://github.com/SiCKRAGETV/sickrage-issues/issues/976

MGaetan89:
Yes, it should. As the code for the web UI and the API will be the same.

fernandog:
Thanks

MGaetan89:
After working a bit more on this, the `future` command of the API will probably return more data than it actually does. It won't break any existing applications, as it is just extra information.

miigotu:
done yet? lol

MGaetan89:
No, still have to work on it.
I'll let you know when I'm done ;)

MGaetan89:
I noticed that the API uses `%Y-%m-%d %H:%M` as a date time format, while the rest of the application uses `%Y-%m-%d %H:%M:%S`. Any reason for that?

If the API uses `%Y-%m-%d %H:%M:%S` like the rest of the application, I could use `sickrage.helper.common.dateTimeFormat` everywhere, so we are sure that we don't have any inconsistency between displays. See https://github.com/MGaetan89/SickRage/commit/ee36ba77eb743d92cd5d69bcede279ef6fc0a312 for an example with `dateFormat`. What do you think?

(@miigotu don't be afraid of all the commits, I'll squash them once I'm done. But I like to use many small commits, so it's easier to track progress ;))

miigotu:
Pretty sure %Y-%m-%d %H:%M:%S will be ok for the api as well.

MGaetan89:
Thanks. I will update everything.

Just in case, in the API, the `timeFormat` (without seconds) is used in the `show` command as follow:

```python
showDict['airs'] = sbdatetime.sbdatetime.sbftime(dtEpisodeAirs, t_preset=timeFormat).lstrip('0').replace(' 0', ' ')
```

MGaetan89:
The API part is ready. I will start the web part now (still in this PR, so no merge yet ;)).

MGaetan89:
I also squashed everything into one commit for the API.
I'll do the same for the web part when it's done.

fernandog:
@MGaetan89 nice job! thanks

i didn't understand this:
```
I chose the web UI option. However the date sort might need to use airdate rather than localtime. What do you think?
```

MGaetan89:
I don't remember why the point was xD
But the correct implementation (the one I'm working on) [uses `localtime` for sorting](https://github.com/SiCKRAGETV/SickRage/pull/2438/files#diff-f39ebd434e45fd29b08cdc738181b151R14).

miigotu:
it would be either local time or network time converted from airdate. and technically it should sort by whichever time type is set in SR.

miigotu:
Actually, I think everything should be using localtime now that I think about it. Shows are posted on tvdb with their airdate/time set in the timezone of the network right?

If everything in SR used that time converted to local time then the setting in SR could be eliminated, and it would make even searches more reliable.

MGaetan89:
So if I understand correctly, [this sorting map](https://github.com/SiCKRAGETV/SickRage/pull/2438/files#diff-f39ebd434e45fd29b08cdc738181b151R14) is correct, as it uses `localtime`?

miigotu:
yes I think it is correct, without digging too much into the code ofc.

MGaetan89:
OK. We still have time until it is in `master` to figure this out.
I will try to finish this by the end of the week so it can be merged in `develop` for testing. As this is a big change, I would appreciate if this can stay a few days in `develop` before going in `master`.

miigotu:
yes, I will release to master just before thus is merged

MGaetan89:
The Web UI part was easier than I thought. I guess I just did an excellent job preparing the work for the API, that's why it was easy :P
So this is in your hands now :)

MGaetan89:
@fernandog Can you see if this fixed https://github.com/SiCKRAGETV/sickrage-issues/issues/976?

fernandog:
Sure!

fernandog:
@MGaetan89 

In "coming episodes' Continuum appears as "missed" but in the API appears as Today
Now its 23:10 here. Dailysearch already changed to "Wanted"

image

**Missed**
```
/api/*****/?cmd=future&sort=date&type=missed
{
    "data": {
        "missed": [
            {
                "airdate": "2015-09-17", 
                "airs": "quinta 11:00 ", 
                "ep_name": "Bewilderment of Heart", 
                "ep_plot": "The Darkness overtakes the citizens of Vega; William wrestles with the truth about his time in the desert; Arika becomes the victim of her own manipulations; Michael grows concerned about Noma; Alex searches for the Amphora.", 
                "episode": 11, 
                "indexerid": 277462, 
                "network": "Syfy", 
                "paused": 0, 
                "quality": "Custom", 
                "season": 2, 
                "show_name": "Dominion", 
                "show_status": "Continuing", 
                "tvdbid": 277462, 
                "weekday": 4
            }
        ]
    }, 
    "message": "", 
    "result": "success"
}
```

**Today**
```
/api/*****/?cmd=future&sort=date&type=today
{
    "data": {
        "today": [
            {
                "airdate": "2015-09-18", 
                "airs": "sexta 10:00 ", 
                "ep_name": "Power Hour", 
                "ep_plot": "Alec joins Carlos and the VPD; Kiera and Garza form an alliance.", 
                "episode": 3, 
                "indexerid": 258171, 
                "network": "Showcase (CA)", 
                "paused": 0, 
                "quality": "Custom", 
                "season": 4, 
                "show_name": "Continuum", 
                "show_status": "Continuing", 
                "tvdbid": 258171, 
                "weekday": 5
            }
        ]
    }, 
    "message": "", 
    "result": "success"
}
```

miigotu:
@fernandog there is no `Continuum` in the missed you posted, that is a different show `Dominion`.

Confusing `Continuing` with `Continuum` ?

Both the api and the webui use the same exact function and paramteters now so there is no way it can be different.

fernandog:
@miigotu 
that's the issue. Continuum should be in "Missed" not in "Today". Today <> Missed 

Comingepisodes: 2 shows in "Missed"
API: only one show in "Missed"

MGaetan89:
The grouping of episodes on the web is not the same as in the new `ComingEpisodes` class, I guess.
On the web, the grouping logic is in the mako file. I will have a look later to use the same logic on both sides.

MGaetan89:
Even in the make file the logic is not always the same xD
I will unify that and the problem should be fixed.

fernandog:
@MGaetan89 there's something wrong here:

First 3 eps airs in the same time (using network time) but only first one is missed
image

fernandog:
why one datetime is -4 and the other -3 ??
That's why SR it's not marking as WANTED minutes after the airdate

2015-09-21 23:00:28 INFO     DAILYSEARCHER :: Show 2015-09-21 23:00:00-04:00 ends and now it is 2015-09-21 23:00:28.385246-03:00

miigotu:
Seems network timeszones is not taking DST into account?

fernandog:
humm.. could be. but why one is correct (-3) and the other not? Both don't use the same code?

miigotu:
on the indexer it might be set as -3, network timezones might know that network is currently in DST -4.

fernandog:
is this an issue?

fernandog:
@miigotu

still don't get it:

Show ends 2015-09-21 23:00:00-04:00, so it ends 23:00 ? 
and why 23:00 it's not less then 2015-09-21 23:00:28.385246-03:00 ?

miigotu:
Its 23:00 UTC -4 hours, so 7pm

23:00 UTC -3 is 8pm, so no, it isn't less.

fernandog:
but why we need to use the UTC in the comparison?
if we really need to use, shouldn't we have convert airtime to UTC then back to user local time?
otherwise datetimes are wrong in the UI

miigotu:
@fernandog you pointed out the discrepancy, we obviously will look into it and it will be fixed if it is broken.

"but why we need to use the UTC in the comparison?"
Doesnt even make any sense because it is already ISO ( UTC and TZ)

Must we stop everything and fix it NAOW and dig into the code to find you some answer immediately or
can you wait until we have time to look at it and hopefully figure it out? This is all a work in progress.
Explaining it to you wont make it be fixed faster, especially when we don't know the cause yet. When we do you will know right away because it will be fixed, and you will get an update notification. good? coolthx

PR #2441, opened 2015-09-08
Start making providers use params correctly, to fix cachecontrol (cac…

miigotu:
…he/sessions)


miigotu:
Still need to finish making other providers pass params correctly, but this has to be merged to fix other issues.

PR #2444, opened 2015-09-08
Added missing developer info to footer

OmgImAlexis:


fernandog:
let's enabled this for everybody so we can ask users the page load time. related to issue 2723

PR #2445, opened 2015-09-08
Missed imports, fixup conditions

miigotu:


MGaetan89:
So if the user did not configure a username/password, he can not see this?

WebSpider:
I was also wondering about this ... Why the sblogin?

-----Original Message-----

miigotu:
sblogin is True if you have no login configured. Its included so that it is not visible when not logged in.

miigotu:
It is sort of redundant, I missed this:
https://github.com/SiCKRAGETV/SickRage/pull/2445/files#diff-7a62d0eb6a8e5fb4f863e112ea1920b7R216

PR #2467, opened 2015-09-09
Add temp log lines to HDT

fernandog:
Want to check if I'm the only one having this issue


fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/2729

PR #2470, opened 2015-09-10
Remove jQuery image loading and size checking

OmgImAlexis:
This will be added back in once we work out how to make it not kill loading times. Ref: https://github.com/SiCKRAGETV/sickrage-issues/issues/2723

### DIFF
--- a/gui/slick/views/home.mako
+++ b/gui/slick/views/home.mako
@@ -387,6 +387,9 @@ $(document).ready(function(){
           % endif
         });
 
+        // Hides size column for now until we can fix it
+        $('[data-show-size]').hide();
+        $('[data-column="6"]').hide();
 });
 </script>
 </%block>
@@ -778,7 +781,7 @@ $(document).ready(function(){
 
         ## <% show_size = sickbeard.helpers.get_size(curShow._location) %>
         ## <td align="center" data-show-size="${show_size}">${sickbeard.helpers.pretty_filesize(show_size)}</td>
-        <td align="center" data-show-size="0">Disabled!</td>
+        <td align="center" data-show-size="0"></td>
 
         <td align="center">
             <% paused = int(curShow.paused) == 0 and curShow.status == 'Continuing' %>
--- a/gui/slick/views/home.mako
+++ b/gui/slick/views/home.mako
@@ -104,11 +104,6 @@ $.tablesorter.addParser({
     type: 'numeric'
 });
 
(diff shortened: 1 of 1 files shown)


gui/slick/views/home.mako
```diff
@@ -781,12 +776,13 @@ $(document).ready(function(){
             <span class="visible-print-inline">${download_stat}</span>
         </td>
 
-        <% show_size = sickbeard.helpers.get_size(curShow._location) %>
-        <td align="center" data-show-size="${show_size}">${sickbeard.helpers.pretty_filesize(show_size)}</td>
+        ## <% show_size = sickbeard.helpers.get_size(curShow._location) %>
+        ## <td align="center" data-show-size="${show_size}">${sickbeard.helpers.pretty_filesize(show_size)}</td>
+        <td align="center" data-show-size="0">Disabled!</td>
```
MGaetan89: Just hide the column for the moment, instead of showing "Disabled!", no?

OmgImAlexis: I wasn't sure where the hiding was and I didn't know if we'd need to rip out half of the javascript you added. If there's an easier way to do it let me know.

MGaetan89: I guess the easiest way would be to add `class="hidden"` (maybe check that this is the correct Bootstrap class) [here](https://github.com/SiCKRAGETV/SickRage/pull/2356/files#diff-ce131b2d456b1eaed7f0dba1e738d4e6R609), [here](https://github.com/SiCKRAGETV/SickRage/pull/2356/files#diff-ce131b2d456b1eaed7f0dba1e738d4e6R625) and on your "Disabled" cell (which can now be empty).

OmgImAlexis: Alright, still need to have the size check commented out but that should be good.

miigotu:
Hide it?

OmgImAlexis:
@miigotu tried to hide it and it breaks the table layout because of tablesort. For now we could just use `&nbsp` or a space?

miigotu:
`<td />` ?

gui/slick/views/home.mako
```diff
@@ -392,6 +387,9 @@ $(document).ready(function(){
           % endif
         });
 
+        // Hides size column for now until we can fix it
+        $('[data-show-size]').hide();
+        $('[data-column="6"]').hide();
```
OmgImAlexis: @MGaetan89 @miigotu "fixed" it.

MGaetan89: Might work

OmgImAlexis: Well it hides the columns after they've loaded, just not sure if we should hot-fix it like this or actually remove it for now?

PR #2478, opened 2015-09-11
Fix HDT: use bs3 instead of bs4

fernandog:
https://github.com/SiCKRAGETV/SickRage/commit/36fefecb06e74219f68ee34f43fd1b37b0565ce1

### DIFF
--- a/lib/BeautifulSoup.py
+++ b/lib/BeautifulSoup.py
@@ -1,0 +1,2017 @@
+"""Beautiful Soup
+Elixir and Tonic
+"The Screen-Scraper's Friend"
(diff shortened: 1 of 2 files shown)


fernandog:
@miigotu for some reason it's not working anymore (I spent hours debugging that. maybe I made a mistake). Its still getting only the first item in the result.

'tables' variable is ok now
the 'torrents' variable shows only one torrent.
I export the content to a .html and it shows only one.

miigotu:
using bs3 over bs4 made no difference, it just adds bulk to the source. we cant control what page they send back, except by the arguments to the url.

PR #2482, opened 2015-09-12
Adding mako cache cleaner upon startup of new version

WebSpider:
Fills FR SiCKRAGETV/sickrage-issues#2712

### DIFF
--- a/SickBeard.py
+++ b/SickBeard.py
@@ -349,6 +349,16 @@ class SickRage(object):
         if self.consoleLogging:
             print "Starting up SickRage " + sickbeard.BRANCH + " from " + sickbeard.CONFIG_FILE
 
+        # Clean up after update
+        if sickbeard.GIT_NEWVER:
+            toclean = os.path.join(sickbeard.CACHE_DIR, 'mako')
+            for root, dirs, files in os.walk(toclean, topdown=False):
+                for name in files:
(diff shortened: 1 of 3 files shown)


SickBeard.py
```diff
@@ -349,6 +349,16 @@ def start(self):
...
 
+        # Clean up after update
+        if sickbeard.GIT_NEWVER:
+            toclean = os.path.join(sickbeard.CACHE_DIR, 'mako')
+            for root, dirs, files in os.walk(toclean, topdown=False):
+                for name in files:
+                    os.remove(os.path.join(root, name))
+                for name in dirs:
+                    os.rmdir(os.path.join(root, name))
+            sickbeard.NEWVER = 0
```
MGaetan89: GIT_NEWER?

miigotu:
ahaha, wanna PR?

SickBeard.py
WebSpider: Ah. I missed that at 3am :D

Will correct it later today, should be GIT_NEWVER

-----Original Message-----

MGaetan89:
It's set as `int` everywhere. So it needs to be changed everywhere to use `False`properly.

MGaetan89:
https://github.com/SiCKRAGETV/SickRage/pull/2485

PR #2483, opened 2015-09-12
Remove size column js as it doesn't exist anymore.

OmgImAlexis:


fernandog:
@OmgImAlexis I still have a gray column in my /home page. Cleared browser cache and deleted mako folder

image

MGaetan89:
@OmgImAlexis Did you change the column index in the JS?
I think it would have been easier to revert my PR until the DB is updated :P

OmgImAlexis:
@MGaetan89 there's no way to do that as we've changed all of mako, etc. since then that's why I did it like this. You'll have to go through and change all of the js back manually.

miigotu:
I think it is only the column array index now that needs changed from 7 to 6

OmgImAlexis:
@fernandog is that on master or dev? I don't have the issue at all on my dev install.

fernandog:
Dev

OmgImAlexis:
Try now, miiotu merged another PR into dev and that should fix it.

fernandog:
Working. Thanks

PR #2488, opened 2015-09-12
Add Strike Provider

duramato:
This PR adds getstrike.net as a search provider ,a known DHT search engine


miigotu:
I'll add this, but it is not written correctly for SR. I will fix this one when I finish fixing the rest, but I wont be accepting any more until they are all fixed in SR and the PR will have to follow our params/url layout in the future.

duramato:
@miigotu is any provider correctly written atm I can use for future reference?

miigotu:
bluetigers is ok I think, I did alphabetically up to ftdb I think so far.

basically they need to have self.login_params and self.search_params, and no replacement characters in the search url. any uri paramters "search=Show S00X00" needs to be in a dict and passed as params:

in `__init__`:
```python
        self.urls = {
            'base_url': 'https://www.bluetigers.ca/',
            'search': 'https://www.bluetigers.ca/torrents-search.php'
        }

        self.search_params = {
            "c16": 1, "c10": 1, "c130": 1, "c131": 1, "c17": 1, "c18": 1, "c19": 1
        }
```
and in `_doSearch`:
```python
                self.search_params['search'] = search_string

                data = self.getURL(self.urls['search'], params=self.search_params)
```
This creates one cache item for bluetigers, since the url we are passing to cachecontrol is always `'https://www.bluetigers.ca/torrents-search.php'`

Then requests parses the params and passes them with the url like this:

`https://www.bluetigers.ca/torrents-search.php?c16=1&c10=1&c130=1&c131=1&c17"=1&c18=1&c19=1&search=Show+S00X00`

As much as possible any calls to urllib or urllib2 methods need to be avoided.

PR #2489, opened 2015-09-12
Change dereferer.org to nullrefer.com

duramato:
dereferer.org is a lot down ,while nullrefer seems to be way more stable and faster

### DIFF
--- a/sickbeard/__init__.py
+++ b/sickbeard/__init__.py
@@ -772,7 +772,7 @@ def initialize(consoleLogging=True):
 
         CPU_PRESET = check_setting_str(CFG, 'General', 'cpu_preset', 'NORMAL')
 
-        ANON_REDIRECT = check_setting_str(CFG, 'General', 'anon_redirect', 'http://dereferer.org/?')
+        ANON_REDIRECT = check_setting_str(CFG, 'General', 'anon_redirect', 'https://www.nullrefer.com/?')
         PROXY_SETTING = check_setting_str(CFG, 'General', 'proxy_setting', '')
         PROXY_INDEXERS = bool(check_setting_int(CFG, 'General', 'proxy_indexers', 1))
 


miigotu:
@SiCKRAGETV/moderators input?

flightlevel:
Agree @miigotu 
Dereferer hasn't worked for a month for me. Thought it was just a routing issue with my ISP but obviously others experiencing the same thing.

OmgImAlexis:
Can we not have an option in the settings? I've trust dereferer and don't really want to use something I've never used before.

miigotu:
I agree with @OmgImAlexis

WebSpider:
Also, this is a config.ini setting already. This pr just changes default 

-----Original Message-----

OmgImAlexis:
Maybe just add a thing to the FAQ stating that if they're having issues try changing it to this in their `config.ini`?

duramato:
Having this empty ,i believe isnt a good option as the websites will see your domain (or wherever you are hosting SR) nullrefer i just a more stable option and faster btw
Also couchpotato is now using nullrefer as the default

flightlevel:
Completely forgot its already an option in General Configuration > advanced settings.
Leave as is and the user can change if needed.
More people would have complained if it was major, hasn't worked on Australia's biggest ISP for a while http://forums.whirlpool.net.au/forum-replies.cfm?t=2437254

OmgImAlexis:
@flightlevel Bigpond may be "Australia's biggest ISP" but really the only people that use them now are government businesses and old people, not really the demographic we fit with. :stuck_out_tongue_winking_eye:

miigotu:
This is like the third time you tried to change this xD. It works for the majority of users, and is more trusted by the users it works for. If you need a different option set it in your settings.

PR #2492, opened 2015-09-13
move js from home.mako to home.js, switch from fuzzydate to timeago, and remove size column

miigotu:


MGaetan89:
Is there a minified version of `jquery.timeago.js`?

miigotu:
its already tiny =p

PR #2496, opened 2015-09-13
2492 Make directory when not available

jorgb:
Fix for #2495  
Based upon 'develop' branch this time ;-)

### DIFF
--- a/sickbeard/postProcessor.py
+++ b/sickbeard/postProcessor.py
@@ -337,6 +337,8 @@ class PostProcessor(object):
 
             self._log(u"Moving file from " + cur_file_path + " to " + new_file_path, logger.DEBUG)
             try:
+                new_dir_path = os.path.dirname(new_file_path)
+                helpers.make_dirs(new_dir_path)
                 helpers.moveFile(cur_file_path, new_file_path)
                 helpers.chmodAsParent(new_file_path)
             except (IOError, OSError), e:


miigotu:
there are issues with doing it this way. I can explain at least one a little later.

miigotu:
This needs to check if the Show dir exists first. We cannot make directories when the show location might not be set correctly (show moved? archived? disk offline/no mounted?)

jorgb:
That goes beyond my abilities, the code  there needs to be the code that is
executed when "add show" is used as it seems to create a dir just fine
there.

miigotu:
Because the directory must exist. Blindly creating the show directory during post processing is no an option.

jorgb:
So that means it does not get fixed at all?

miigotu:
It isn't broken. You are breaking it by deleting the show folders that SR has made, you said that yourself (even though you linked the wrong issue here).

jorgb:
Never mind then. Happy bug fixing!
- Jorgen

ipso:
What happens if the disk where you store your shows crashed? Do you manually have to re-create the hundreds or thousands of directories before Sickrage will function again?

miigotu:
Hundreds or thousands of Shows? lol.

fernandog:
@jorgb you should check if directory exists first with os.path.exists()

jorgb:
Thanks, but I am done with trying to contribute. To me it feels that if 
the issue I filed is already closed (before helping me making it an 
acceptable PR to merge), that there is no desire to merge it in the end 
anyway.

I filed a feature request as suggested, and trust the original devs, if 
they so desire to implement this, will make it a better fix than I tried 
to do.

With regards,
- Jorgen

fernandog:
@jorgb please try again. everybody here started doing small contributions.

ipso:
I would submit my patch, but Mr. miigotu clearly thinks this is a non issue even though there are numerous requests for it here and in the forums. 

Closing issues before having any real discussion, then laughing at their reasoning is a great way to build a thriving community around a project. Why would anyone bother wasting their time?

fernandog:
@jorgb forgot about this setting:
stop SR, edit your config.ini and set to 1 this

create_missing_show_dirs

jorgb:
Thanks bunches for following up, I will check it out later!

With regards,
- Jorgen

PR #2497, opened 2015-09-14
Ran pep8 against sickbeard/ recursively ..

WebSpider:
PEP8: Blank lines
PEP8: Missing/extra whitespaces
PEP8: Mixing tabs and spaces
PEP8: Compare to 'None' using 'is', not ==
PEP8: Statements ending with semicolon
PEP8: Multiple statements in one line


WebSpider:
If travis fails, I need to rethink this.

MGaetan89:
@miigotu tried to do that too a few days ago. Maybe one PR poet item would make it easier to see what's going wrong?

miigotu:
yeah I did an autopep8 and it makes some problems. the is/is not for sure when comparing against true/false/none becomes an issue for one.

miigotu:
i think we need to do it incrementally, while selecting individual tests for each commit.

PR #2502, opened 2015-09-14
Fixes https://github.com/SiCKRAGETV/sickrage-issues/issues/2774

medariox:
This PR fixes the bug.

### DIFF
--- a/sickbeard/subtitles.py
+++ b/sickbeard/subtitles.py
@@ -76,7 +76,7 @@ def isValidLanguage(language):
     return True
 
 def getLanguageName(language):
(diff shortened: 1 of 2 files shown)


miigotu:
Sucks that we have to scan the video twice to handle this. Maybe the new version of subliminal fixes that limitation.

medariox:
Yes, sadly subliminal.video.scan_subtitle_languages only scans directories (was driving me crazy) and that's why we have to separate the tasks. I think the whole subtitle logic can be improved tho. A version update would be definitely beneficial, let's see if I manage to find some time this week.

PR #2506, opened 2015-09-15
Only select required field in history

MGaetan89:
An other potential performance improvement for people with a lot of shows: only select fields that we use, instead of `SELECT *`.
I'll try to do that everywhere during my refactoring.

### DIFF
--- a/sickrage/show/History.py
+++ b/sickrage/show/History.py
@@ -38,7 +38,7 @@ class History:
         else:
             actions = []
 
-        common_sql = 'SELECT h.*, show_name ' \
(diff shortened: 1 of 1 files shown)


WebSpider:
Yes, this will make a difference in memory usage, since we rely on python to clean up lingering objects, vars, etc 

SQL actions won't be much quicker, unless we adjust indexes to align with queries more.

-----Original Message-----

MGaetan89:
I agree. But I don't trust my SQL knowledges enough to do that :P

PR #2507, opened 2015-09-15
Fixes fuzzydates being used & fixes if checks in js

OmgImAlexis:
Since Sickrage sometimes uses `True` and sometimes uses `1` we can't be sure which one will be present on the meta tag so we need to use an array to check for either of them.

### DIFF
--- a/gui/slick/js/new/home.js
+++ b/gui/slick/js/new/home.js
@@ -282,14 +282,17 @@ $(document).ready(function(){
         sortAppend: [[2,0]]
     });
 
-    if ($("#showListTableShows").find("tbody").find("tr").size() > 0)
+    if ($("#showListTableShows").find("tbody").find("tr").size() > 0){
         $.tablesorter.filter.bindSearch( "#showListTableShows", $('.search') );
+    }
 
-    if($('meta[data-var="sickbeard.ANIME_SPLIT_HOME"]').data('content') == 'True')
-        if ($("#showListTableAnime").find("tbody").find("tr").size() > 0)
+    if(['True', 1].indexOf($('meta[data-var="sickbeard.ANIME_SPLIT_HOME"]').data('content')) >= 0){
+        if($("#showListTableAnime").find("tbody").find("tr").size() > 0){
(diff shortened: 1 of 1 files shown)


fernandog:
working for me, thanks!

what's the range to be in "In about a month" ?

OmgImAlexis:
Bigger than 30, less than 45.
````
 var words = seconds < 45 && substitute($l.seconds, Math.round(seconds)) ||
        seconds < 90 && substitute($l.minute, 1) ||
        minutes < 45 && substitute($l.minutes, Math.round(minutes)) ||
        minutes < 90 && substitute($l.hour, 1) ||
        hours < 24 && substitute($l.hours, Math.round(hours)) ||
        hours < 42 && substitute($l.day, 1) ||
        days < 30 && substitute($l.days, Math.round(days)) ||
        days < 45 && substitute($l.month, 1) ||
        days < 365 && substitute($l.months, Math.round(days / 30)) ||
        years < 1.5 && substitute($l.year, 1) ||
        substitute($l.years, Math.round(years));
````

fernandog:
less equal or just less?

OmgImAlexis:
Less, I think over 45 days converts to months.

MGaetan89:
Why not cast the value to bool and then check if it's true? That's how it is fine in python.

OmgImAlexis:
@MGaetan89 or maybe we could just set one or the other in Python instead of randomly choosing one. @miigotu any ideas why some of these are `1` and some are `True`?

MGaetan89:
@OmgImAlexis I'm doing that in my refactoring when I see `0` and `1` instead of `False` and `True`. But I can't guarantee I won't miss any.

OmgImAlexis:
Okay, for now I'll leave it as is. Once it's refactored let me know and I'll switch this over to `Boolean("True")`.

MGaetan89:
Sounds to good to me.

WebSpider:
Better yet: We should convert the rest of the code to use True/False. It's one of those PHP habits I need to shake myself as well.

-----Original Message-----

MGaetan89:
Converting the code is a WIP ;=

PR #2508, opened 2015-09-16
Moved comingEpisodes js to own file, added metas block to main

OmgImAlexis:


miigotu:
-.- why would you do coming episodes while comingEpisodes is being refactored by @MGaetan89 ?

Conflicts are your job this time lol!

OmgImAlexis:
All I did was move a bunch of code from it, this should actually help with that since after the first merge all comingEpisodes will be is mako code. It means I can work on cleaning up the js more without getting in his way.

miigotu:
His PR was already open. Most times when 2 PR touch the same file it causes a conflict and needs a manual merge.

fernandog:
@OmgImAlexis sorting by airdate in comingEpisodes/ does not work.

I think we can remove the "Sort By:" and the listbox as we can sort in the table itself.
 what do you think?

OmgImAlexis:
Sounds good. I'll have a play around with it and see if I can get it working again. I think it may be because of where it was getting the sort data.

OmgImAlexis:
@fernandog fixed. https://github.com/SiCKRAGETV/SickRage/pull/2511

PR #2509, opened 2015-09-16
Hide upgrade div on restart

OmgImAlexis:


fernandog:
ow! few lines only. I was trying to use java script to hide the div

OmgImAlexis:
@fernandog haha, hiding things with css is a lot easier. For future referance to add something for a specific page just use the block in the file that has the same name such as `restart.mako`.

fernandog:
thanks!

fernandog:
worked! I will try tomorrow to hide the div when user click "update"

OmgImAlexis:
We can do it when they click update, just use something like this.

````
$('.class_for_update_link_here').on('click', function(){
    $('.upgrade-notification').hide();
});
````

**Comment by fernandog** [comment]
also we can do like CP adding a block page with "SickRage is backing up your data"

I found this but CP looks nicer
http://malsup.com/jquery/block/

```
App.blockPage('Please wait while CouchPotato is being updated with more awesome stuff.', 'Updating');
```

https://github.com/RuudBurger/CouchPotatoServer/blob/develop/couchpotato/core/_base/updater/static/updater.js#L93

OmgImAlexis:
I really don't want to add yet another library for jQuery when it's not needed. Half of the ones we've got now aren't even really needed.
Also keep in mind most of that stuff is in js because of the whole ui being a single page app.

fernandog:
NP. I thought it was some function that only needed to be called.

PR #2510, opened 2015-09-16
No git clean

miigotu:


fernandog:
@miigotu  why not just do 'git clean -f' instead of -df

Not matter the name of folder tracked files wont be removed

```
-d
Remove untracked directories in addition to untracked files
```

miigotu:
because it is not just folders that are being removed and shouldnt be. There are 0 reasons to do a git clean.

fernandog:
Also backups are not being removed as 'backup' folder is in gitignore and user can change as its hardcoded

fernandog:
If there are files being removed we must add to .gitignore

miigotu:
Backups are being removed. It has been reported tons of times. So ok the backup folder is ignored, what about the zip/rar inside?

There is NO REASON we need to git clean. Just dont git clean and it solves it all at once.

miigotu:
What if a user put a folder containing their video library in the SR dir with some randon name? They would lose all of their videos.

gui/slick/views/config_general.mako
```diff
@@ -729,7 +729,7 @@
                             </label>
                         </div>
 
-                        <div class="field-pair">
+                        <div class="field-pair" hidden>
```
MGaetan89: "hidden" is outside of "class" ;)

fernandog:
I understand that. But as explain above: removing the -d it wont remove folders. Just fyi.
The backup feature was added a long time ago and we never had issues but its ok. You dont want git clean. I get that.

PR #2512, opened 2015-09-16
Fix Remove Show Confirmation, keeps jquery format uniform by using class to bind to.

jalim:


miigotu:
@OmgImAlexis is this correct?

jalim:
With the latest Develop branch on my system clicking remove just removes the show without confirmation.

OmgImAlexis:
No, I don't think so. I'm not sure but I think other things use the remove class on anchor, that'll need to be checked. It doesn't matter anyways as this is only for the `deleteShow` page. It should be left as is.

jalim:
It seemed to me that up until recently the remove class had always been used in this way to bind.

OmgImAlexis:
[Checking history](https://github.com/jalim/SickRage/commit/23c1a61c0f2a3ff85279ee25e808828f05d67f4e#diff-9e941ddd870fc24662d5ea3410f78095R44), you're right about that. I'm not sure but I think I may have added that to fix an issue with it not being triggered. If it's working fine then remove it.

jalim:
How about assigning it a less generic class?  Like RemoveShow or something?  At the moment just the remove class is working for me but as you say this might be used somewhere I'm not seeing it and might cause conflicts down the road.

OmgImAlexis:
I completely agree but if you're going to do that you need to actually change it in the mako file as well.

jalim:
Done.

PR #2513, opened 2015-09-16
Use common code to delete a show

MGaetan89:
https://github.com/SiCKRAGETV/sickrage-issues/issues/2567

### DIFF
--- a/sickbeard/webapi.py
+++ b/sickbeard/webapi.py
@@ -35,6 +35,7 @@ from sickrage.media.ShowPoster import ShowPoster
 from sickrage.media.ShowBanner import ShowBanner
 from sickrage.show.ComingEpisodes import ComingEpisodes
 from sickrage.show.History import History
+from sickrage.show.Show import Show
 from sickrage.system.Restart import Restart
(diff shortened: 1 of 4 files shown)


MGaetan89:
Forgot to say something about this.
[In the API](https://github.com/SiCKRAGETV/SickRage/pull/2513/files#diff-bc7f9073a5871410a6bf8a76d4bb6001L2221), if the show can not be removed, it returns the exception message.
[In the web UI](https://github.com/SiCKRAGETV/SickRage/pull/2513/files#diff-357efeabeb6d6921e8196af70c431bebL1518), if the show can not be removed, the exception message is logged, and we display "Unable to delete show: show.name".

I used the second solution for both. Maybe I should just add a log entry with the exception? Or is it better to return the exception message? Currently it is "This show is already queued to be removed".
I was looking for my next target, and sometimes we may have multiple exceptions to handle. So it might be better to return the exception message I guess. What do you think?

miigotu:
either way, as long as they are the same

MGaetan89:
The message will be the same, that's not an issue.
I'll switch to using the exception message then.

PR #2516, opened 2015-09-16
Return exception as error message if deleting a show fails

MGaetan89:
As per https://github.com/SiCKRAGETV/SickRage/pull/2513#issuecomment-140837111 and the following comments.


MGaetan89:
Rebased.

PR #2518, opened 2015-09-17
various cleanups (dont merge, will validate when back from holiday)

WebSpider:
- Typo's.
- Forgotten declarations in __init__
- Incorrect docstring quotations
- Comparisons to None not using 'is' or 'is not'
- Simplifications of nested comparisons

### DIFF
--- a/sickbeard/generic_queue.py
+++ b/sickbeard/generic_queue.py
@@ -41,6 +41,8 @@ class GenericQueue(object):
 
         self.lock = threading.Lock()
 
+        self.amActive = None
+
     def pause(self):
         logger.log(u"Pausing queue")
         self.min_priority = 999999999999
--- a/sickbeard/search.py
+++ b/sickbeard/search.py
@@ -52,7 +52,7 @@ def _downloadResult(result):
     """
 
     resProvider = result.provider
-    if resProvider == None:
+    if resProvider is None:
         logger.log(u"Invalid provider name - this is a coding error, report it please", logger.ERROR)
         return False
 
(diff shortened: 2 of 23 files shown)


sickbeard/search.py
```diff
@@ -592,7 +592,7 @@ def searchProviders(show, episodes, manualSearch=False, downCurQuality=False):
 
                 logger.log(u"Seeing if we want to bother with multi-episode result " + _multiResult.name, logger.DEBUG)
 
-		# Filter result by ignore/required/whitelist/blacklist/quality, etc
+        # Filter result by ignore/required/whitelist/blacklist/quality, etc
```
MGaetan89: This looks wrong

sickbeard/webapi.py
```diff
@@ -272,7 +272,12 @@ class ApiCall(ApiHandler):
     _help = {"desc": "No help message available. Please tell the devs that a help msg is missing for this cmd"}
 
     def __init__(self, args, kwargs):
-        # missing
+
+        self._requiredParams = []
+        self._optionalParams = []
+        self.indexer = None
+        self._missing = []
```
miigotu: This will break webapi I think.

sickbeard/generic_queue.py
```diff
@@ -41,6 +41,8 @@ def __init__(self):
 
         self.lock = threading.Lock()
 
+        self.amActive = None
+
```
fernandog: Should be True or False

fernandog: In this case True

miigotu: Actually should be False (None is ok because we just check if blah.amActive)

self.amActive = True is missing from start of the run method.

fernandog: Yes. Sorry. Ita missing setting true in def run(self, force=False):

miigotu:
Same as my other comment. Max 1 or 2 files per PR for this stuff and go over every single change manually.

WebSpider:
Ok, i will make smaller pr's for this, makes sense!

-----Original Message-----

miigotu:
@WebSpider I've also been working on this, my commits for last night are weeeeeeee.

Any idea how to get the sickbeard.logger wrapper working right to not cause lint errors?

WebSpider:
I've been postponing the logger lint issues as well :D

-----Original Message-----

PR #2522, opened 2015-09-17
Move encodingKludge to sickrage module

MGaetan89:
This is not fixing anything or adding any feature. It just moves the `encodingKludge` in the `sickrage` module.
Some warnings were also addressed as I saw them, but that was not the goal here.

This PR might conflict with https://github.com/SiCKRAGETV/SickRage/pull/2518.


WebSpider:
It will definately clash with my latest docstrings addition, since i did some changes in ek as well :)

miigotu:
technically encoding kludge is a dirty hack and should be removed with full Unicode conversion soonish =¶

MGaetan89:
What does it mean concretely?
How can I do that ?

miigotu:
Its complicated. Basically once we use requests everywhere, requests will automatically decode the response to unicode based on response headers. If we then use codecs.open everywhere and make sure all content is unicode everywhere except on I/O (file output, console logging, post/uri data, database operations) we have no more unicode or foreign language issues.

Its ok to just rebase this and merge, because the unicode sandwich battle is an ongoing war.

MGaetan89:
Rebased :)

MGaetan89:
Would it help to use unicode string (`u'Hello'`) everywhere in the new `sickrage` package instead of just string (`'Hello'`)? Or can it break something to mixed both (I don't think so as it is already the case)?

miigotu:
Yes, unicode() is preferred over str()

WebSpider:
I will look for str() as well when going over all the missing docstrings as well, and see if they can be easily replaced by u()

-----Original Message-----

MGaetan89:
I cans witch to unicode as I move logic into the `sickrage` module. I guess it will be easier like that rather than changing everything in `sickbeard`.

PR #2529, opened 2015-09-19
Move exceptions to sickrage module

MGaetan89:
The `ex() ` function is now in `sickrage.helper.common` and was renamed `exception()`
The exceptions were moved into the `sickrage.helper.exceptions` file.
The following exceptions were renamed:
- `CantRefreshException` ->`CantRefreshShowException`
- `CantRemoveException` -> `CantRemoveShowException`
- `CantUpdateException` -> `CantUpdateShowException`
- `FailedProcessingFailed` -> `FailedPostProcessingFailedException`
- `MultipleDBEpisodesException` -> `MultipleEpisodesInDatabaseException`
- `MultipleDBShowsException` -> `MultipleShowsInDatabaseException`
- `PostProcessingFailed` -> `EpisodePostProcessingFailedException`
- `ShowDirNotFoundException` -> `ShowDirectoryNotFoundException`
- `SickBeardException` -> `SickRageException`

The following exceptions were removed because they were not used:
- `ConfigErrorException`
- `EpisodeNotFoundByAbsoluteNumberException`
- `FailedHistoryMultiSnatchedException`
- `FailedHistoryNotFoundException`
- `FileNotFoundException`
- `LaterException`
- `NewzbinAPIThrottledException`
- `NoShowDirException`
- `WrongShowException`


miigotu:
Why rename ex to exception? Its longer to write and will take getting used to writing it.
Also, why put it in helpers?

WebSpider:
Please don't rename this ... Using Exception fully is a typical Java thing (so is splitting everything in a single file, but we'll get to that :D)

-----Original Message-----

miigotu:
I dont mind submodules like `show` with all show related actions contained inside, however renaming ex to exception is a nogo because one it is used to much that it is obvious what ex means and muscle memory is a bitch. ex is just easier to write, and `exception` could easily be confused with `Exception`.

In fact, some linters will give a bad name error for that last reason.

MGaetan89:
I will change `exception` back to `ex`.

@WebSpider it's true that I'm firstly a Java developer, but I don't think that it is a good practice, in any language, to have file with 3k/5k lines in lines.
If you have an other structure idea, feel free to suggest it. It's easier to change it now when I'm starting rather than later when the `sickrage` module becomes big.

miigotu:
Technically ex is an extension of encodingKludge. It just formats exceptions to a format/encoding that can be printed in a console.

miigotu:
I think they should stay separate though, because ek is bad and will probably go away, ex is necessary for unicode to work properly. ss() from encodingKludge will be modified and kept.

MGaetan89:
Welcome back `ex()` :)

miigotu:
Much better, but shouldnt ex be in either exceptions (because it is used with exceptions, and commonly imported with them) or in encoding.py (because that is what it is doing, encoding an exception message for console/file logging) ?

WebSpider:
Well, that's True, however since Python basically doesn't care how packages are organized, and general coding style (pep8) encourages us to focus on usability instead of rigid rules, and looking at the whole sickbeard code we still have, I'd say we split out into a few packages, which resemble functionality in the app as it currently is:

- anything having to do with queueing/event handling (classes, mutators, related exceptions, threads, etc)
- anything having to do with our ui/webserver/callback functions etc
- anything having to do with us requesting information (this is pretty much handled by the providers, but I think indexers could be carved out as well a bit more as well)
- anything having to do with show/ep matching from indexers/providers
- anything having to do with quality assessment
- anything having to do with maintenance
- and off course some general packages such as helpers that provide OS/DB independency

There probably are a few packages/parts of code I forgot, but I guess the main line is: Focus on functionality, not on fixed boundaries.

Just my 0.02 off course :)

-----Original Message-----

MGaetan89:
@miigotu I will move it to `sickrage.helper.exceptions`.
@WebSpider That's more or less what I started doing, no? I have the `helper` package and the `show` package. I didn't really touch the rest yet, but that' probably what will happen.

MGaetan89:
I moved `ex()`.

PR #2530, opened 2015-09-19
Use unicode instead of string in `sickrage` module

MGaetan89:
I don't know if it was necessary everywhere, but I guess it can hurt, does it?

### DIFF
--- a/sickrage/helper/common.py
+++ b/sickrage/helper/common.py
@@ -16,6 +16,6 @@
 # You should have received a copy of the GNU General Public License
 # along with SickRage.  If not, see <http://www.gnu.org/licenses/>.
 
-dateFormat = '%Y-%m-%d'
-dateTimeFormat = '%Y-%m-%d %H:%M:%S'
-timeFormat = '%A %I:%M %p'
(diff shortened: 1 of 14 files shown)


miigotu:
I dont think it is ok to do it in mass like this. dict keys can have issues between unicode/ascii

MGaetan89:
So I just remove unicode from dict keys?

miigotu:
For sure. But like everything in encoding.py doesnt need to be done too, and most other places in this PR.

Paths, file and show names/info and function returns need to be unicode, but parameters to python built-ins don't

It's too complicated of a transition to do it all at once like this.

MGaetan89:
Ok, I will update this and let you know when it's done.

MGaetan89:
Updated in https://github.com/SiCKRAGETV/SickRage/commit/846903f0df7d72fc23bd65806221bfe01634992c

miigotu:
Im still scared to merge a batch change like this =P

miigotu:
I dont think this is necessary, and it would be hell to fix the conflicts now =P

PR #2533, opened 2015-09-19
Make use of retention

duramato:
Usenet-crawler has 2300 days  retention BUT atm sickrage does not make use of it ,thus doesnt return a lot of old results
This PR fixes that issue


miigotu:
This is incorrect. Sickrage already sets the maxage to one day before the show actually aired. Why would you search any older than necessary?

duramato:
yeah, you are right

PR #2540, opened 2015-09-20
Clear up some linting errors in name_parser

miigotu:


sickbeard/name_parser/parser.py
```diff
@@ -346,19 +347,21 @@ def _combine_results(self, first, second, attr):
         b = getattr(second, attr)
 
         # if a is good use it
-        if a != None or (type(a) == list and len(a)):
+        if a != None or (isinstance(a, list) and a):
```
MGaetan89: Can't this just be `a != None`?
If `a` is `None`, the first part is `False` and the second part is `False` too (`None` is not a `list`).
If `a` is not `None`, the first part is `True` and the rest is not evaluated.

miigotu: Yes that is kinda silly, but I will leave that as is for now as this PR is explicitly for linting fixes.

Code changes and refactoring should be kept separate.

PR #2545, opened 2015-09-20
(do not merge, will test more when not on slow connection) Tidy up code: removing obvious non-fatal errors

WebSpider:
- Removing unused imports
- Removing unreachable code
- Removing trailing semicolons
- Fixing inconsistent indentations
- Fixing where instance attributes defined outside __init__ by adding them to it


miigotu:
This needs to be done one file at a time. I have started doing it 50 times probably as a batch operation/PR and I always end up finding hidden errors the tools create.

PR #2548, opened 2015-09-21
Add public-private var to providers

fernandog:
@miigotu is that it ?

Now we need the template change. @OmgImAlexis Do you know how to do like this?
https://github.com/SiCKRAGETV/SickRage/pull/1960#issuecomment-112254094

### DIFF
--- a/sickbeard/providers/alpharatio.py
+++ b/sickbeard/providers/alpharatio.py
@@ -42,6 +42,7 @@ class AlphaRatioProvider(generic.TorrentProvider):
         generic.TorrentProvider.__init__(self, "AlphaRatio")
 
         self.supportsBacklog = True
+        self.public = False
 
         self.enabled = False
         self.username = None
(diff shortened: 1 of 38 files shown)


OmgImAlexis:
@fernandog yeah, that should be simple enough to do. Did you just want me to check if it's public and change the background colour?

miigotu:
@OmgImAlexis yes for now, we can do it in stages.

Still debating how we should distinguish nzb from torrent as well, and how we should implement the list as a 2 column display. Initial layout would probably have Torrents on the left, and nzb on the right, sorted with public at the top and private below (and colored different based on public/private)

When they are selected they will float to the top of the left column in the selected/user defined order and still be draggable/rearrangeable.

Maybe on color for nzb, one for torrent, and a different shade of those colors designating private? or maybe just an asterisk designating private?

Either way there will need to be a color key to let the user know what the colors mean.

PR #2552, opened 2015-09-21
Pylint fixes in SickBeard.py, fix redefined class in mainDB.py

miigotu:


WebSpider:
Just installed pylint in addition to pycharm linting. Which options do u use for pylint, so it reports stuff we should look at?

(I for one don't care about 100char/line, as my screen are highres anyways, for instance)

-----Original Message-----

miigotu:
If you dont do that import hook you get a lot of errors about "cannot import blah" and then a lot of undefined

In atom I didnt change any defaults, I just skip over the too many branches and too many returns and the logger ones.

For a lot of unnecessary ones I add a directive in the file to disable it for now, and later we can go back through and find ones for refactoring.

PR #2556, opened 2015-09-21
Fix linting errors in logger.py

miigotu:


WebSpider:
Are you taking a per-file approach on the cleanup? If so, I'll take tv*.py and work on it tonight .. 

Finally was able to get a proper test setup going last night :)

-----Original Message-----

miigotu:
@WebSpider sort of, per file is best I think as it is easier to review. 

In my ~/.pylintrc I just added this:

```
init-hook=import sys;sys.path.insert(1, '/home/miigotu/Projects/SickRage');sys.path.insert(1, '/home/miigotu/Projects/SickRage/lib')
```

PR #2557, opened 2015-09-21
Fix missing self.public

fernandog:


fernandog:
@miigotu should I close this? binsearch also needs self.public

PR #2558, opened 2015-09-21
Update "Coming Episodes" to the same splitting on the web and in the API

MGaetan89:
I'm not 100% sure about this yet, so don't hesitate to comment if you see something wrong.

@fernandog

### DIFF
--- a/sickrage/show/ComingEpisodes.py
+++ b/sickrage/show/ComingEpisodes.py
@@ -133,11 +133,11 @@ class ComingEpisodes:
             result['airs'] = str(result['airs']).replace('am', ' AM').replace('pm', ' PM').replace('  ', ' ')
             result['airdate'] = result['localtime'].toordinal()
 
-            if result['airdate'] < today:
+            if ComingEpisodes.is_episode_missed(result):
                 category = 'missed'
-            elif result['airdate'] >= next_week:
+            elif ComingEpisodes.is_episode_later(result):
                 category = 'later'
-            elif result['airdate'] == today:
+            elif ComingEpisodes.is_episode_today(result):
                 category = 'today'
             else:
                 category = 'soon'
@@ -158,3 +158,21 @@ class ComingEpisodes:
             grouped_results[category].append(result)
 
         return grouped_results
+
+    @staticmethod
+    def is_episode_later(data):
+        return data['localtime'].toordinal() >= (date.today() + timedelta(days=7)).toordinal()
+
+    @staticmethod
+    def is_episode_missed(data):
+        air_date = data['localtime']
+        runtime = data['runtime']
+
+        if runtime:
+            air_date += timedelta(minutes=runtime)
+
+        return air_date.toordinal() < date.today().toordinal()
+
+    @staticmethod
+    def is_episode_today(data):
+        return (data['localtime'] - date.today()).days == 0
--- a/gui/slick/views/schedule.mako
+++ b/gui/slick/views/schedule.mako
@@ -3,6 +3,7 @@
     import sickbeard
     from sickbeard.helpers import anon_url
     from sickbeard import sbdatetime
+    from sickrage.show.ComingEpisodes import ComingEpisodes
     import datetime
     import time
     import re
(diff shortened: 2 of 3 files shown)


fernandog:
why not sure?

MGaetan89:
Why I tested it in a "non-calendar" layout, I had one "Missed" section with one episode and one "Sunday" section with 3 missed episodes in it (the background color was red as expected). I don't know if it was like that before or not.

Secondly, on the web UI, `today` is defined as `datetime.datetime.now().replace(tzinfo=network_timezones.sb_timezone)`, while it is simply `date.today().toordinal()` in the API. Might need to have a look at that too.

fernandog:
about today, if airs today and airdate < now, keep as today, otherwise missed. isn't?

sickrage/show/ComingEpisodes.py
```diff
@@ -158,3 +158,21 @@ def get_coming_episodes(categories, sort, group, paused=sickbeard.COMING_EPS_DIS
...
+        runtime = data['runtime']
+
+        if runtime:
+            air_date += timedelta(minutes=runtime)
+
+        return air_date.toordinal() < date.today().toordinal()
+
+    @staticmethod
+    def is_episode_today(data):
+        return data['localtime'].toordinal() == date.today().toordinal()
```
miigotu: This last function is the only one I see that looks wrong at first glance xD.

is_episode_today is checking if the time is at exactly a day it looks like, no hours, no seconds.

I think you want something like:

```python
def is_episode_today(data):
        return (data['localtime'] - date.today()).days == 0
```

MGaetan89:
Thanks for the feedbacks.
I update the `is_episode_today()` method according to @miigotu suggestion.

fernandog:
Should we merge?

MGaetan89:
If you don't see anything wrong, you can.

miigotu:
let's wait, need to review this with rested eyes later

fernandog:
@MGaetan89 

```
Error !

TypeError: unsupported operand type(s) for -: 'datetime.datetime' and 'datetime.date'

<tbody style="text-shadow:none;"> % for cur_result in results: <% if int(cur_result['paused']) and not sickbeard.COMING_EPS_DISPLAY_PAUSED: continue cur_indexer = int(cur_result['indexer'])
/home/osmc/SickRage/sickrage/show/ComingEpisodes.py, line 178:
return (data['localtime'] - date.today()).days == 0
/home/osmc/SickRage/gui/slick/views/comingEpisodes.mako, line 97:
<%
/home/osmc/SickRage/gui/slick/views/layouts/main.mako, line 220:
/home/osmc/SickRage/lib/mako/runtime.py, line 853:
callable_(context, *args, **kwargs)
```

miigotu:
should be `- datetime.now()` or `- datetime.today()` then.

MGaetan89:
I will look again at this after I'm done on the API part. Shouldn't be too long now.

fernandog:
@MGaetan89 shouldn't be ?

return (data['localtime'] - datetime.date.today()).days == 0

MGaetan89:
No, I don't think so. Because I do `from datetime import date` and not `import datetime`.

fernandog:
@MGaetan89 
let me know when you need to test this. sunday/monday are great days to test

MGaetan89:
I will go through all the comment related to this PR tomorrow to update it accordingly. But I probably won't be done before the end of the day. Maybe it will be by Tuesday evening.

fernandog:
NP!

fernandog:
@MGaetan89 thinking about this. We should not check ep status in comingEpisodes.

It user dailysearcher freq is 3 hours, comingepisodes wont  sjow as wanted. CE is like a calendar. Not ep status managment. Agree?

fernandog:
all datetimes should use network_timezones.sb_timezone, isn't?

miigotu:
No, it has to check episode status, because skipped and ignored are not included in the results. And no, network_timezones.sb_timezone has nothing to do with anything other than refreshing from indexer.

The object is to clean up the bad code not just add it right back after we refactor it.

fernandog:
I only want my airdates with my local time. Thats why
We should convert unaired>wanted every full hour then.
Some users like @neoatomic sets dailysearch to 240min. ComingEpisodes will be wrong. 4 hours delayed. 
ComingEpisodes IMO is a calender. Episode status user could check in displayShow.

miigotu:
"I only want my airdates with my local time." no shit, so why would you need to convert anything when the airdates are already in a proper format.

Coming episodes are only for shows you WANT or are UNAIRED. It has always been this way.

fernandog:
Did I asked something diff  ?!
Just said to dont get ep status (unaired/wanted). But is beter to use airdates instead of ep status (as I said before)
Who said something about ignored?!

fernandog:
just to explain in details what I previous said (as It's not a good ideia as it depends on dailysearch freq):
get all WANTED ep status and show in comingEpisodes.
all the UNAIRED would be classified: today, soon, later

will start to post full details of my comments as they are get misinterpreted

miigotu:
wanted is not only coming. they can be missed, soon, or current. everybody else understands how it is supposed to be.

fernandog:
AGAIN:
get all WANTED ep status and show in comingEpisodes.
all the UNAIRED would be classified: today, soon, later

miigotu:
that's how it always has and will work. so why comment to argue something when its already how it is. just don't say anything then.

miigotu:
like how you originally said to not check status, now you say to get the status.

fernandog:
(as It's not a good ideia as it depends on dailysearch freq):

please stop, people are already complaining about how you reply uses

fernandog:
@MGaetan89 everything ok? We didnt hear from you anymore

MGaetan89:
@fernandog Yes everything's ok. Just gonna be very busy this month. So I won't have a lot of time for SickRage until November. I focus my free time on my app until then.

miigotu:
Can this one be updated or did you change this in another PR or plan to....?

We have lots of PR open atm that would be nice to get completed/merged.

fernandog:
@MGaetan89 when you do plan to finish this PR? thanks!

MGaetan89:
As I said a couple of weeks back, I don't have a lot of free time this month.
So I will help working on SickRage next month (so either next week or in two weeks).

MGaetan89:
I rebased this with the latest version of `develop`. Hope I didn't lost anything in the process.
I will try to (finally) finish this.

PR #2561, opened 2015-09-21
Change default DATA_DIR to PROG_DIR/data

WebSpider:
After some discussions in SiCKRAGETV/sickrage-issues#2788 this proposes to change the default DATA_DIR from PROG_DIR to PROG_DIR/data, and adds it to gitignore.

Offcourse, the cmdline switch to specify DATA_DIR is still available, but this way we at least know where user-specific stuff ends up, and we don't have to keep on excluding all kinds of locations. It's all in PROG_DIR/data by default.

Note: Only on new installs!
Also: Previously made backups may not be able to restore, due to changed file locations

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 #  SR User Related   #
 ######################
 cache*/
+data*/
 Logs/
 restore/
 backup*/
@@ -16,6 +17,7 @@ server.key
 ######################
 /tests/Logs/*
 /tests/cache/*
+/tests/data*/
 /tests/sickbeard.db*
 /tests/cache.db*
 /tests/failed.db
--- a/SickBeard.py
+++ b/SickBeard.py
@@ -142,7 +142,7 @@ class SickRage(object):
         sickbeard.MY_FULLNAME = os.path.normpath(os.path.abspath(__file__))
         sickbeard.MY_NAME = os.path.basename(sickbeard.MY_FULLNAME)
         sickbeard.PROG_DIR = os.path.dirname(sickbeard.MY_FULLNAME)
-        sickbeard.DATA_DIR = sickbeard.PROG_DIR + os.sep + "data"
+        sickbeard.DATA_DIR = os.path.join(sickbeard.PROG_DIR, 'data')
         sickbeard.MY_ARGS = sys.argv[1:]
         sickbeard.SYS_ENCODING = None
 
--- a/SickBeard.py
+++ b/SickBeard.py
@@ -142,7 +142,7 @@ class SickRage(object):
         sickbeard.MY_FULLNAME = os.path.normpath(os.path.abspath(__file__))
         sickbeard.MY_NAME = os.path.basename(sickbeard.MY_FULLNAME)
         sickbeard.PROG_DIR = os.path.dirname(sickbeard.MY_FULLNAME)
-        sickbeard.DATA_DIR = sickbeard.PROG_DIR
+        sickbeard.DATA_DIR = sickbeard.PROG_DIR + os.sep + "data"
         sickbeard.MY_ARGS = sys.argv[1:]
         sickbeard.SYS_ENCODING = None
 
--- a/sickbeard/__init__.py
+++ b/sickbeard/__init__.py
@@ -648,7 +648,7 @@ def initialize(consoleLogging=True):
         DEFAULT_PAGE = check_setting_str(CFG, 'General', 'default_page', 'home')
 
         ACTUAL_LOG_DIR = check_setting_str(CFG, 'General', 'log_dir', 'Logs')
-        LOG_DIR = os.path.normpath(os.path.join(LOG_DIR, ACTUAL_LOG_DIR))
+        LOG_DIR = os.path.normpath(os.path.join(DATA_DIR, ACTUAL_LOG_DIR))
         LOG_NR = check_setting_int(CFG, 'General', 'log_nr', 5) #Default to 5 backup file (sickrage.log.x)
         LOG_SIZE = check_setting_int(CFG, 'General', 'log_size', 1048576) #Default to max 1MB per logfile
         fileLogging = True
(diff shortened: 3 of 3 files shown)


sickbeard/__init__.py
```diff
@@ -648,7 +648,7 @@ def initialize(consoleLogging=True):
         DEFAULT_PAGE = check_setting_str(CFG, 'General', 'default_page', 'home')
 
         ACTUAL_LOG_DIR = check_setting_str(CFG, 'General', 'log_dir', 'Logs')
-        LOG_DIR = os.path.normpath(os.path.join(DATA_DIR, ACTUAL_LOG_DIR))
+        LOG_DIR = os.path.normpath(os.path.join(LOG_DIR, ACTUAL_LOG_DIR))
```
miigotu: This is not right

SickBeard.py
```diff
@@ -142,7 +142,7 @@ def start(self):
         sickbeard.MY_FULLNAME = os.path.normpath(os.path.abspath(__file__))
         sickbeard.MY_NAME = os.path.basename(sickbeard.MY_FULLNAME)
         sickbeard.PROG_DIR = os.path.dirname(sickbeard.MY_FULLNAME)
-        sickbeard.DATA_DIR = sickbeard.PROG_DIR
+        sickbeard.DATA_DIR = sickbeard.PROG_DIR + os.sep + "data"
```
miigotu: This is not right, always use os.path.join

miigotu:
Im not sure this is necessary, the problem was fixed with https://github.com/SiCKRAGETV/SickRage/commit/b7edb8a61473b140f002bea11174af47ab738bbb

WebSpider:
I see the issue was resolved, but should we really create database files, config files etc in our own root?

It's cleaner practise to just store all install-specific info in a seperate dir.

Let me know what you think.

-----Original Message-----

miigotu:
yeah its normal practice, but our source is not normal since we always track it with git. so we always know what is a source file and what is not. this makes the pros of not using the source dir for data mute, as the git index is what groups the source, not the dir.

WebSpider:
However we do still have the possibility for update method = source, which is an untracked zip file, right?

-----Original Message-----

miigotu:
@WebSpider I guess we can merge this after fixing the 2 things I commented on.

I am only reluctant about this because I was looking into a way to reliably decide where the data should actually go by default, and I didn't want to change it twice.

IE: On linux, if running as a real user with a home dir use ~/.config/sickrage/, if running as root or a user without a home dir use /var/sickrage or something similar

On windows it would be a little different ofc.

miigotu:
@WebSpider you alive with this one?

WebSpider:
Yeah I'm picking up on SR again after settling back home tonight!

-----Original Message-----

sickbeard/__init__.py
```diff
@@ -1622,7 +1622,7 @@ def save_config():
     new_config['General']['config_version'] = CONFIG_VERSION
     new_config['General']['encryption_version'] = int(ENCRYPTION_VERSION)
     new_config['General']['encryption_secret'] = ENCRYPTION_SECRET
-    new_config['General']['log_dir'] = ACTUAL_LOG_DIR if ACTUAL_LOG_DIR else 'Logs'
+    new_config['General']['log_dir'] = ACTUAL_LOG_DIR if ACTUAL_LOG_DIR else 'logs'
     new_config['General']['log_nr'] = int(LOG_NR)
```
fernandog: why logs and not Logs?

miigotu:
@WebSpider figured you were still on vaca =P had fun?

WebSpider:
Got back Friday, had family stuff today.
Was great fun. Did I mention I love ouzo? :)

-----Original Message-----

sickbeard/__init__.py
WebSpider: For the same reason the python dirs are called lib, sickbeard and sickrage, instead of Lib, SickBeard and SickRage: It's just good practice to do all lowercase in dirnames when being cross platform/ cross-fs.

Besides that, it's in a number of LFS best practices that I know of, and I'm sure some others that I don't know of.

-----Original Message-----

.gitignore
```diff
@@ -1,6 +1,7 @@
 #  SR User Related   #
 ######################
 cache*/
+data*/
 Logs/
```
fernandog: need to lowercase here then

WebSpider:
@miigotu shouldn't this be a packager decision, per distro/os?

Gentoo fs layout is very different from Ubuntu, is very different from slackware.

Imo the reason to do this, is to avoid user confusion if they're running from source or GIT. We just inherited the way SB does this, which was something I've been frowning upon from the start tbh ;)

-----Original Message-----

.gitignore
WebSpider: Got it. Will correct that in a few hrs.

-----Original Message-----

miigotu:
yes Im not a fan of how it is atm either. either way the backup/restore is fixed at least lol. figured you needed to get home and more into it before you got to testing your PR's xD.

fernandog:
SB does /data now

adtwomey:
So the best way to get sickrage back working is do a new installs. As the git pull just stop sickrage from working

WebSpider:
@adtwomey what do you mean? This PR hasn't been merged yet, and it would not change anything on existing installs, it just changes the default locations on new install.

.gitignore
WebSpider: Default logdir is inside DATA_DIR, so it is already covered by the gitignore of data/
This doesn't need changed

.gitignore
```diff
@@ -16,6 +17,7 @@ server.key
 ######################
 /tests/Logs/*
```
miigotu: and here xD.

WebSpider: it's in there so we dont delete all files from people who havent moved yet..
In tests, logdir is also inside DATA_DIR, so in tests/data/logs, which is .gitignores as per tests/data*

miigotu:
So everything here is good now?

WebSpider:
Yup. Tested it in an old install, and a new install. Old locations keep being used and dont get purged, new locations are used when doing a clean install.

PR #2565, opened 2015-09-22
update xthor add quotes to search string

raver2046:
add quotes to search string for better results in searching.

### DIFF
--- a/sickbeard/providers/xthor.py
+++ b/sickbeard/providers/xthor.py
@@ -47,7 +47,7 @@ class XthorProvider(generic.TorrentProvider):
         self.cj = cookielib.CookieJar()
 
         self.url = "https://xthor.bz"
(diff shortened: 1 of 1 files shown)


miigotu:
Not really the correct solution, But I will merge it for now until I can rework the params.

raver2046:
Yes but it works.

miigotu:
When my kid doesnt listen I could kick him in the face and it would work, but is it the correct way?

raver2046:
i try to do with my poor knowledge dude

miigotu:
not being mean there, just making a point. If you want to learn, when someone who knows says "its not the best way" a good response would be "OK. which way would be better?" instead of "but it works".

That's why I said I would merge it, so it would work until I could make it use the better process later and maybe ping you to show you then.

PR #2566, opened 2015-09-22
Move the js from mako to .js files

OmgImAlexis:


OmgImAlexis:
@MGaetan89 does this look okay? I know I broke a few things with my last few commits.

MGaetan89:
Quickly looked at it, and it looks good to me.

OmgImAlexis:
By the way I'm redoing a lot of the home.js logic for the sorting functions and moving to using data elements so if you need to change anything with that you may need to wait till after I've finished this.

MGaetan89:
I'm on the API builder for the moment, so you can have fun on the home ;)
If we are not planning anything there, we won't block each other.

OmgImAlexis:
All's good, just thought I'd let you know. :+1:

OmgImAlexis:
@fernandog or @miigotu when you get the chance can this please be merged into develop.

miigotu:
tested everything you actually changed? filterRow and table sorting?

OmgImAlexis:
@miigotu yep, as far as I can tell it all works fine. The only problem I had was the reset button and I've fixed that.

miigotu:
Ok, merged another release just before adding this, because we just fixed a hella stuff in hotfixes yesterday morning.

OmgImAlexis:
Alright. :+1:

miigotu:
Going to work on merging /js/*.js and /js/new/*.js ?

OmgImAlexis:
Yes, first all js from mako files is to be moved into `/js/new/*.js` and then I'm going to lint all of the old stuff and merge everything into single files per page along with common files like the tablesorting parsers. Once @MGaetan89 gets the api finished and we get the new parts up and going I should be able to remove the `<%block name="metas">` as well.

miigotu:
Btw, the last round was a big speed improvement for me, good job.

OmgImAlexis:
Woo! Guessing a lot of that's from linting. That's another reason why I want to get the js out of the mako files. My linter goes crazy thinking .mako files are pure HTML and thinks the `<%block>` sections are weird regex that aren't closed so I can't actually lint the js. Once it's in it's own file it's fine.

PR #2568, opened 2015-09-22
Really fixes SiCKRAGETV/sickrage-issues#2774

medariox:
This time for real.

Please note:
Some embedded subtitles are not being correctly recognized by subliminal (actually enzyme: https://github.com/Diaoul/enzyme) and will be tagged as "und" (undefined), in that case the correct subtitle will be downloaded. There is nothing we can do to solve this, I guess it depends on how the subtitle has been embedded.

### DIFF
--- a/sickbeard/tv.py
+++ b/sickbeard/tv.py
@@ -1433,6 +1433,17 @@ class TVEpisode(object):
         """Look for subtitles files and refresh the subtitles property"""
         self.subtitles = subtitles.subtitlesLanguages(self.location)
 
+    def getWantedLanguages(self):
+        languages = set()
+        for language in frozenset(subtitles.wantedLanguages()).difference(subtitles.subtitlesLanguages(self.location)):
+            languages.add(subtitles.fromietf(language))
+        self.refreshSubtitles()
+        return languages
+
+        if not languages:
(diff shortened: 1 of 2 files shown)


sickbeard/tv.py
```diff
@@ -1433,6 +1433,17 @@ def refreshSubtitles(self):
...
         self.subtitles = subtitles.subtitlesLanguages(self.location)
 
+    def getWantedLanguages(self):
+        languages = set()
+        for language in frozenset(subtitles.wantedLanguages()).difference(subtitles.subtitlesLanguages(self.location)):
+            languages.add(subtitles.fromietf(language))
+        self.refreshSubtitles()
+        return languages
+
+        if not languages:
```
miigotu: this is unreachable.

medariox: It is reachable, it is called here: languages = self.getWantedLanguages();

miigotu: read it again.

```python
return languages
if not languages:
    logger.log.....
```
Lines 1443-1445 need moved to right after `languages = self.getWantedLanguages();`

medariox: Oh, you are right of course. Will open a new PR with the changes.

PR #2572, opened 2015-09-23
update libertalia

raver2046:
if login is unsuccessful this is an error


fernandog:
if we change this to error users will submit issues. we can't do that

miigotu:
Unsuccessful login is not an error, @raver2046 you could make it logger.WARNING

PR #2574, opened 2015-09-23
Fixes "Snatched (Proper) 1080p WEB-DL" not fitting on a single line

OmgImAlexis:
The reason for the massive amount of additions and deletion is my editor "fixed" the indenting to use 4 spaces instead of the tab character.

### DIFF
--- a/gui/slick/css/style.css
+++ b/gui/slick/css/style.css
@@ -4,139 +4,139 @@ fonts
 /* Open Sans */
 /* Regular */
 @font-face {
-	font-family: 'Open Sans';
+    font-family: 'Open Sans';
 
-	src: url('fonts/OpenSans-Regular-webfont.eot');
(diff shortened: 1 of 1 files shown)


fernandog:
The future column widget in comingEpisodes could prevent this too. Indexer column is useless now. Ep name could be hidden too IMO

OmgImAlexis:
@fernandog some people like those columns, no need to remove them. What do you mean "future column widget in comingEpisodes / displayShow"?

fernandog:
Not remove. But add the button 'select columns' like in the /home page.

miigotu:
@OmgImAlexis let me translate. On ComingEpisodes, there is no column selector.

OmgImAlexis:
Oh.. Yeah we should probably add them.

PR #2576, opened 2015-09-23
Move js and lint

OmgImAlexis:
@MGaetan89 I think table parsing should be moved to a generic file so we don't have duplicates of it, what do you think?


MGaetan89:
If the logic is the same, I agree, yes.

PR #2579, opened 2015-09-23
Fix page url on shazbat.tv

duramato:


sickbeard/providers/shazbat.py
```diff
@@ -76,7 +77,7 @@ def __init__(self, provider):
 
     def _getRSSData(self):
 
-        rss_url = self.provider.url + 'rss/recent?passkey=' + provider.passkey + '&fname=true'
+        rss_url = self.urls['base_url'] + 'rss/recent?passkey=' + provider.passkey + '&fname=true'
```
miigotu: `self.provider.urls['base_url']`

duramato:
@miigotu done ,but why is it that way?

miigotu:
Because it is not in the provider class, it is in the TVCache class for the provider. To access the provider attributes from the cache class in each provider you must go through `self.provider.`

duramato:
oh i see thanks for the info
also @miigotu,the rebase its done :)

PR #2581, opened 2015-09-23
New API builder

MGaetan89:
Here is a preview of the new API builder so you can test it if you want.
I am still working on it, so don't merge it yet!

For the moment, I am using a new `apiBuilder2.mako`, but it will replace the existing `apiBuilder.mako` file in the end.
The command parameters are not taken into account right now. So only command without parameters or with only optional parameters work as expected.

Here is the points that I still have to do:
- [x] Use parameters value when making an API call
- [x] Restore debug/profile/jsonp options
- [x] Support image commands
- [x] Support dark theme
- [x] Add a link to the API builder in *Config* > *General* > *Interface*
- [x] Merge `apiBuilder2.mako` into `apiBuilder.mako`
- [x] Move JS from `apiBuilder2.mako` into `apibuilder.js`
- [x] Add a search bar to quickly access a command


MGaetan89:
First update of the PR:
- The dark theme is now supported. It can be improved, but it works. Now the user keep the same interface/theme between the web UI and the API builder
- I added a link in *Config* > *General* > *Interface* that points to the API builder
- I added a search box to quickly access an API command. It's still missing the scroll to the command documentation, because I have an offset issue...

MGaetan89:
I added back the "Debug", "Profile" and "JSONP" options.
Can someone tell what the debug option does or how I can test if it works? Even by [looking at the code](https://github.com/SiCKRAGETV/SickRage/blob/master/sickbeard/webapi.py#L130) I don't get it :/

@miigotu?

miigotu:
I think debug is for "actually make the call to SR", otherwise it is supposed to be just building the api call and return dummy data. Maybe that wasnt working as it used to since tornado replaced cherry py from the comment.

OmgImAlexis:
@MGaetan89 can you tag me once this is fully merged, I want to lint all of the new js you've added.

MGaetan89:
@miigotu if it's not working, do I remove it or do I keep it? I guess we can keep it, but how can I generate dummy data?

@OmgImAlexis sure, no problem.

miigotu:
@MGaetan89 actually now that I think about it, debug might have been to enable the logger in cherrypy. I think you can just remove it.

MGaetan89:
I will remove it help.
I don't see the harm in logging them, at least in debug. As long as we have a log when the call fails, it's good.

MGaetan89:
I removed the debug option https://github.com/MGaetan89/SickRage/commit/f99cb0301dd6afe920a05ac37e1467daecdbc637

miigotu:
Soon done this one? xD

MGaetan89:
It was not an easy task, but I'm almost done.
Just need to finish the parameter management.

miigotu:
About that, the old style paramters will still work right? I know we already established that it should, just makng sure because I know there are a lot of people that still use those other apps.

OmgImAlexis:
@MGaetan89 you alright with me redoing your js after this? I want to lint it and then change the vanilla javascript to jQuery to help with cross browser compatibility.

MGaetan89:
@miigotu What do you mean with "old style parameters"?
@OmgImAlexis Yes, of course. I'll let you know once this PR is done.

miigotu:
maybe I misunderstood your last comment about parameters.

MGaetan89:
I meant that I still have to take the parameters value when making an API call (`indexerid`, `season`, ...). Otherwise some API commands can't be used.

MGaetan89:
This should be good to merge.
Now, instead of `2 * number_of_shows` SQL queries, I only need one.

It's not perfect yet. Here is a list of things that can be improved:
- When selecting a command from the search bar, we should scroll to the show
- When changing the `indexerid` value, the season/episode lists should be reseted
- We should not send the request if one of the required parameter is missing

@OmgImAlexis if you want to check the JS, you're free to go now.

neoatomic:
Nice work @MGaetan89 !

OmgImAlexis:
@MGaetan89 once this is merged I'll lint it, @miigotu tag when once it's merged please.

fernandog:
@OmgImAlexis

OmgImAlexis:
@fernandog cheers.

fernandog:
@
@MGaetan89
Nice work!

Why future and not coming episodes?

```
Future
Get the coming episodes
```

MGaetan89:
@fernandog Thanks.
The command is called `future`, I didn't changed anything about that. We will be able to change that in the next API version.

fernandog:
Ok. NP. Thanks

PR #2582, opened 2015-09-23
Don't format JSON before sending the response

MGaetan89:
With this PR, the JSON response of every API commands will no longer be formatted. This save some data when using the API (and maybe improve speed, as there is no formatting to do).
For my list of shows, I have to downloaded 35ko before this PR, and 19ko after.
If someone is calling an API command directly from the browser and want a formatted response, there are a lot of good browser extensions that do that (and way better than what we currently do).
The new API builder will format the result on the client side, so it won't be affected.

### DIFF
--- a/sickbeard/webapi.py
+++ b/sickbeard/webapi.py
@@ -92,7 +92,6 @@ result_type_map = {
 class ApiHandler(RequestHandler):
     """ api class that returns json results """
     version = 5  # use an int since float-point is unpredictable
-    intent = 4
 
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


MGaetan89:
We could also enable gzip-compression (if it is not already used, I didn't check), but it might have side-effects maybe.

miigotu:
We already  use gzip compression =P. This PR just makes it not tidy (indent) the json output right? That will even improve it for the client side.

MGaetan89:
Yes, it only removes the indentation from the response.

PR #2585, opened 2015-09-24
Update pkg_resources.py - Sickrage won't start on Openelec and Audo with RPi2 and others.

p00chie:
Sickrage won't start on Openelec and Audo with RPi2 and others.

See:
http://openelec.tv/forum/80-sabnzbd-suite/78152-unable-to-bring-up-sickrage-page-after-autoupdate?start=45


miigotu:
I will accept this if you remake the PR to develop. We do not accept PR directly to master.

PR #2587, opened 2015-09-24
Fixes ${sbRoot}  in js

OmgImAlexis:


OmgImAlexis:
@MGaetan89 you okay with changing `sbRoot` to `srRoot` considering you're moving the python stuff from `sickbeard` to `sickrage`?

MGaetan89:
Yeah, no problem. In python, the variable is set in multiple places in `webserve`. So it should be easy to change on that part.
It will need a bit more work in the mako files I guess. Maybe we can have both for a while until we are sure the migration is complete.

OmgImAlexis:
Yeah that's fine, I'll make sure to only use `srRoot` in all of the `/js/new*.js` files on my next lint PR and add it in the `main.mako` where the js `sbRoot` is.

PR #2589, opened 2015-09-24
Image API commands now return the correct Content-Type

MGaetan89:
This PR is here to help https://github.com/SiCKRAGETV/SickRage/pull/2581 so it does not mix too many things.

The goal is to make image API commands (currently, `show.getbanner`, `show.getfanart`, `show.getnetworklogo` and `show.getposter`)  return the correct `Content-Type`, instead of `text/html`.
This way, the new API builder can automatically detect image response and display them, so we don't have to manage [a list of commands](https://github.com/SiCKRAGETV/SickRage/blob/master/gui/slick/js/apibuilder.js#L12) that return an image.

### DIFF
--- a/sickbeard/webapi.py
+++ b/sickbeard/webapi.py
@@ -109,8 +109,9 @@ class ApiHandler(RequestHandler):
 
         # set the output callback
         # default json
-        outputCallbackDict = {'default': self._out_as_json,
-                              'image': lambda x: x['image'],
+        outputCallbackDict = {
+            'default': self._out_as_json,
+            'image': self.out_as_image,
         }
 
         accessMsg = u"API :: " + self.request.remote_ip + " - gave correct API KEY. ACCESS GRANTED"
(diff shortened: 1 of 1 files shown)


miigotu:
Great, very important to not mix different changes!

fernandog:
NZB360 stopped working.
Will see which commit did that.

fernandog:
Also DroidBeard.
Its possible SickRageConnect too

@Thraxis 
https://github.com/Thraxis/SickRageConnect

MGaetan89:
Do you have any log or something that could help?

fernandog:
No. Lets wait then reply my bug reports to their app

MGaetan89:
If it is breaking all the applications that are using the API, maybe we should revert this.
Unless they can give us useful information about why it fails so I can adapt this PR accordingly.

MGaetan89:
Just in case: http://nzb360.com/forums/viewtopic.php?f=6&t=599

MGaetan89:
I think I know why it's broken. I set the callback to `self.out_as_image`, but the method is actually `self._out_as_image`. Will fix that in a minute.

PR #2590, opened 2015-09-25
ntpath is windows only. Some systems override this module and it will…

miigotu:
… work ok, others do not (syno?)

@neoatomic @fernandog this might be the source of the config/cache restore issues, or at least a part of it. ntpath WILL work SOMETIMES on SOME SYSTEMS.

@neoatomic could you check if this fixes the syno restore problems?

### DIFF
--- a/SickBeard.py
+++ b/SickBeard.py
@@ -295,15 +295,11 @@ class SickRage(object):
         os.chdir(sickbeard.DATA_DIR)
 
         # Check if we need to perform a restore first
-        try:
-            restoreDir = os.path.join(sickbeard.DATA_DIR, 'restore')
-            if self.consoleLogging and os.path.exists(restoreDir):
-                if self.restoreDB(restoreDir, sickbeard.DATA_DIR):
-                    sys.stdout.write("Restore: restoring DB and config.ini successful...\n")
-                else:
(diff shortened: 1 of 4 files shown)


fernandog:
nice! that's odd because it works in my win7

miigotu:
@fernandog ntpath is for Windows. os.path will use ntpath or posixpath depending on the machine it is on.  I think abeloin did this for possible network paths, but it might fail due to that.

fernandog:
but autobackup won't be a NT path isn't? maybe could be an issue for manual backup

miigotu:
Found the other bug, for config and db's. Adding it to this commit in a sec.

miigotu:
It will be whatever path type is for that OS.

Restore of db and config was only working when running in console btw ^

miigotu:
cherry-picked to master. This is a big fix, no more explaining how to manually restore!

fernandog:
thanks!

neoatomic:
@miigotu The browse function of the backup/restore is broken. Can not navigate to a folder anymore for selecting a file for backuping or restoring. Only manually entering the path is possible.

fernandog:
After this PR ? We moved some JS around in another PR
@OmgImAlexis

OmgImAlexis:
@neoatomic open an issue, it's likely a js issue. No point in discussing it here.

PR #2591, opened 2015-09-25
Add more values to history limit

fernandog:
@duramato


MGaetan89:
I don't know if 10 is really useful.

PR #2592, opened 2015-09-25
Fix history limit and save on change limit

fernandog:
This will fix the first loading of history page when you have a history limit <> 100

@duramato


fernandog:
@OmgImAlexis

OmgImAlexis:
@fernandog looks alright. Don't really know python though, I only handle the mako and js stuff.

PR #2597, opened 2015-09-25
Fixes SiCKRAGETV/sickrage-issues/issues/2954

miigotu:


MGaetan89:
What do the `[]` you added?

miigotu:
It is the same as adding a back tick, it makes sure sql parses it as a string and not an int

sickbeard/db.py
```diff
@@ -281,7 +281,7 @@ def tableInfo(self, tableName):
         :param tableName: name of table
         :return: array of name/type info
         """
-        sqlResult = self.select("PRAGMA table_info(%s)" % tableName)
+        sqlResult = self.select("PRAGMA table_info(`%s`)" % tableName)
```
miigotu: @MGaetan89 this does the same thing

MGaetan89:
Ok, I didn't know the `[]` solution, only the back ticks. Any reason mix the two in this PR?

miigotu:
Well, I originally had all backticks, but everywhere else in SR uses the brackets. brackets in table_info just feels retarded to me so I left that one backticks.

MGaetan89:
I understand. Thanks for the info. It's good to learn something new everyday :)

PR #2602, opened 2015-09-26
Moved more js to own files

OmgImAlexis:


gui/slick/views/displayShow.mako
```diff
@@ -540,8 +493,13 @@ $(document).ready(function(){
                     ${file_size}
                 % endif
             </td>
+            ## <% date = (sbdatetime.sbdatetime.sbfdate(sbdatetime.sbdatetime.convert_to_setting(network_timezones.parse_date_time(epResult['airdate'], show.airs, show.network))), 'never')[int(epResult['airdate']) == 1] %>
+            <% date = sbdatetime.sbdatetime.convert_to_setting(network_timezones.parse_date_time(epResult['airdate'], show.airs, show.network)) %>
```
OmgImAlexis: @MGaetan89 any chance we could get some date/time formatting functions added in python so we don't have to do this every time?

miigotu: Why did you remove the "never" cases?

OmgImAlexis: Oops, I thought I had added that back in. Give me a sec.

gui/slick/js/new/config_postProcessing.js
```diff
@@ -0,0 +1,2 @@
+$('#config-components').tabs();
```
MGaetan89: I think we can get rid of all the `.tabs()` calls by using `data-toggle="tab"`.
http://getbootstrap.com/javascript/#tabs

gui/slick/js/new/config_subtitles.js
```diff
@@ -0,0 +1,13 @@
+$(document).ready(function() {
```
MGaetan89: We do you have this only here (and still some JS outside a few lines below)?

gui/slick/views/displayShow.mako
MGaetan89: Yes, I agree about having a common function to format the date.
I'm always finding something else to do on SickRage, so it's hard to have a prioritized list.

gui/slick/js/new/config_subtitles.js
OmgImAlexis: For now it's just what was in the mako files, I'm slowly removing this. I'll redo the name and stuff for this PR in a bit.

gui/slick/js/new/config_postProcessing.js
OmgImAlexis: Sounds good.

OmgImAlexis: Just checked and I think we still need it but I can move them all to a single file.

PR #2603, opened 2015-09-26
Fix info icon on custom naming legend

VinceVal:
Change class from 'icon-info-sign' to 'glyphicon glyphicon-info-sign'.


miigotu:
Good fix, didnt even know there were supposed to be icons there.

PR #2607, opened 2015-09-26
Fix not being able to set status to compound archived qualities

miigotu:
Fix leak when testing "if int(status) in statusStrings"


miigotu:
@fernandog should fix setting archived qualities.

PR #2608, opened 2015-09-26
Fix subfolder not created for new shows

medariox:
And possibly fixes the "Could not refresh subtitles" problem.


medariox:
https://github.com/SiCKRAGETV/SickRage/pull/2609

PR #2610, opened 2015-09-26
Update pkg_resources.py

p00chie:
As discussed here
https://github.com/SiCKRAGETV/SickRage/pull/2585/


miigotu:
this is already in develop

miigotu:
oops my bad. hard telling where this comes from on my phone.

PR #2614, opened 2015-09-26
Updated 6box logos & added ALL brazilian networks logos

duramato:


miigotu:
Merging. However, `gui/slick/images/network/hbo latin america.png` might need changed to work for both light and dark theme.
@neoatomic input?

duramato:
@miigotu it doesnt seem that bad (one of the only shows i could find lol)
image

neoatomic:
@miigotu Indeed, would have been better to just duplicate the already existing HBO logo.
You need to be careful that the logos work on both themes. (and its tempting not to test.:) )
In the past we also used withe logo's but i converted all of them to color. Think we need to stick to some standards. Especially with the eye on a future new theme. Black/withe is (generally) a no go...

PR #2615, opened 2015-09-27
Multiple JS fixes and refactoring

OmgImAlexis:


miigotu:
Much easier to follow! Still sbRoot in js so this is broken atm, but add that and squash the sbRoot commits together and I think you should merge this as one PR, and start from there for the next.

Makes for a nice sized PR and you dont have to worry as much about getting out of sync or needing rebase.

OmgImAlexis:
@miigotu alright, just changed all of the js ones too. We should be okay now, if there's any issues sbRoot still exists so it shouldn't crash anything. My dev install works fine on all pages.

OmgImAlexis:
How's that?

miigotu:
Great, should merge now?

OmgImAlexis:
Yep.

PR #2618, opened 2015-09-27
Removed fuzzydate and library

OmgImAlexis:


WebSpider:
Why is fuzzy date being removed?
It seems to be all the hype now to show relative dates instead of absolute dates on the internet

miigotu:
@WebSpider it was replaced with timeago a while back (which does the same thing)

PR #2623, opened 2015-09-27
Lint new API Builder

OmgImAlexis:
@MGaetan89 next time you make changes to this could you move everything into the body so we validate HTML spec. Nothing should be outside of it.

### DIFF
--- a/gui/slick/js/apibuilder.js
+++ b/gui/slick/js/apibuilder.js
@@ -1,6 +1,6 @@
-$(document).ready(function () {
+$(document).ready(function() {
     // Perform an API call
-    $('[data-action=api-call]').on('click', function () {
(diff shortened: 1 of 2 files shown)


MGaetan89:
@OmgImAlexis I don't understand what you mean here.

OmgImAlexis:
@MGaetan89 in apibuilder you've got a bunch of tables, etc. after the closing `</body>`, it means the page isn't valid HTML and I want to try to get the whole project validated, that's why we've been linting everything.

MGaetan89:
No, check again. The only thing I have after `</body>` is `</html>`.
The rest if `<%def>` mako commands. Mako removed them when rendering the page.

PR #2627, opened 2015-09-27
Use mako global variables instead of calling `sickbeard.*`

MGaetan89:
Some global mako/template variables were defined in `webserve`.
This PR use them instead of calling `sickbeard.*` directly.


SiCKRAGEB0T:

Thanks!

miigotu:
@MGaetan89 I kinda like moving the other way (use sickbeard.* instead). sickbeard is imported anyways This way we might be able to eliminate some of those vars.

On the other hand, if you go the route of this PR then sickbeard should not be imported and just vars from sickbeard should be passed as arguments to render.

Thoughts?

MGaetan89:
I agree with your first idea to remove those vars.
I just tried to unify everything with this PR:

miigotu:
Lets sit on this one then and maybe get some more input.

OmgImAlexis:
I'd much rather see everything to go `sickbeard.VAR` as they're seen as constants and leave just the variables within that files as normal vars.

miigotu:
@MGaetan89 @OmgImAlexis lets go the other route then, use sickbeard.THE_VAR wherever possible instead of passing it in. This makes sure the true value is always used rather than whatever it was the last time the local variable was set. We need the true value always in mako if we ever hope to use ajax.

PR #2636, opened 2015-09-28
Dynamically get torrent trackers instead of hard coded ones

duramato:
The current way works fine,even if the hardcoded ones dont match the torrent ones (as it will fallback to DHT) ,but this it gets the actual ones
Also added torrent name to magnet file so that it gets added to the client with a name and not the hash (its just cosmetic as it gets the name after it handshakes to other peers)


SiCKRAGEB0T:

Thanks!

PR #2640, opened 2015-09-28
Redo most of kat provider, Also make it use params so cachecontrol wo…

miigotu:
…rks better. Fixes SiCKRAGETV/sickrage-issues/issues/3006

### DIFF
--- a/sickbeard/providers/kat.py
+++ b/sickbeard/providers/kat.py
@@ -20,13 +20,12 @@
 from __future__ import with_statement
 
 import traceback
-import urllib
 import re
 import datetime
+import xmltodict
 
 import sickbeard
-import generic
-
+from sickbeard.providers import generic
 from sickbeard.common import Quality
 from sickbeard import logger
 from sickbeard import tvcache
@@ -34,7 +33,6 @@ from sickbeard import helpers
 from sickbeard import db
 from sickbeard import classes
 from sickbeard.show_name_helpers import allPossibleShowNames, sanitizeSceneName
-from unidecode import unidecode
 
 
 class KATProvider(generic.TorrentProvider):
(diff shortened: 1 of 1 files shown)


sickbeard/providers/kat.py
```diff
@@ -66,81 +76,92 @@ def imageName(self):
...
+
+                            info_hash = item['torrent:infoHash']
+                            #link = item['link']
+
                         except (AttributeError, TypeError, KeyError):
                             continue
 
-                        if mode != 'RSS' and (seeders < self.minseed or leechers < self.minleech):
+                        # Dont let RSS add items with no seeders either -.-
+                        if not seeders or seeders < self.minseed or leechers < self.minleech:
```
fernandog: @miigotu no seeders or seeders < self.minseed ? if user ser minseed to 10, a torrent with 9 seeds won't be added to RSS

miigotu: It shouldnt be added to RSS, because the user has said he doesnt want torrents with less than 10 seeders. With that other fix RSS adds torrents with 0 seeders.

fernandog: @miigotu also we won't know that torrent was discarded in dailysearcher RSS. will be hard to debug that IMO

fernandog:
@miigotu why not IF mode = 'RSS' and seed = 0 then continue?

miigotu:
We would get items with less seeds than the user configured then. RSS is no different than an episode or season search. It must follow the same rules.

miigotu:
And you get the same log line saying it was discarded whether it is rss or not.

fernandog:
so what do you think changing the discard message from debug to info ?

fernandog:
wait. can't do that as we process the same rss item multiple times. damm

miigotu:
no, lol. It is right just the way it is.

PR #2642, opened 2015-09-29
Finalize extratorrent provider

miigotu:
@duramato

### DIFF
--- a/sickbeard/providers/extratorrent.py
+++ b/sickbeard/providers/extratorrent.py
@@ -20,9 +20,9 @@ import re
 import traceback
 import datetime
 import sickbeard
-import generic
 import xmltodict
 
+from sickbeard.providers import generic
 from sickbeard.common import Quality
 from sickbeard import logger
 from sickbeard import tvcache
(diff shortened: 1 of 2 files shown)


SiCKRAGEB0T:

Thanks!

sickbeard/providers/extratorrent.py
```diff
@@ -59,9 +58,6 @@ def __init__(self):
     def isEnabled(self):
         return self.enabled
 
-    def imageName(self):
-        return 'extratorrent.png'
```
duramato: Why did you remove this?

miigotu: @duramato it isnt needed as long as you name the file right.

That class is derived from `sickbeard.generic.TorrentProvider`, which is derived from `sickbeard.generic.GenericProvider`

If you look in `sickbeard.generic.GenericProvider` here:
https://github.com/SiCKRAGETV/SickRage/blob/master/sickbeard/providers/generic.py#L97

You see these three methods:
```python
   def getID(self):
        return GenericProvider.makeID(self.name)

    @staticmethod
    def makeID(name):
        return re.sub("[^\w\d_]", "_", name.strip().lower())

    def imageName(self):
        return self.getID() + '.png'
```

This takes the provider name, replaces anything that isn't a letter or number, strips whitespace from the ends, and lowercases it and then appends `.png`

Why override it if we dont need to?

PR #2650, opened 2015-09-29
Use enzyme to detect embedded subtitles instead of subliminal

medariox:
Fixes https://github.com/SiCKRAGETV/sickrage-issues/issues/3024, https://github.com/SiCKRAGETV/sickrage-issues/issues/2936, https://github.com/SiCKRAGETV/sickrage-issues/issues/2874. Probably also https://github.com/SiCKRAGETV/sickrage-issues/issues/2965.

Credits to @Diaoul for the original code.

### DIFF
--- a/sickbeard/subtitles.py
+++ b/sickbeard/subtitles.py
@@ -24,8 +24,9 @@ from sickbeard import db
 from sickrage.helper.common import dateTimeFormat
 from sickrage.helper.encoding import ek
 from sickrage.helper.exceptions import ex
+from enzyme import MKV
+from babelfish import Error as BabelfishError, Language, language_converters
 import subliminal
-import babelfish
 import subprocess
 
 subliminal.cache_region.configure('dogpile.cache.memory')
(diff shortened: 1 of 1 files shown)


Diaoul:
This is not the proper fix. You should use str for paths, not unicode in python2. The issue is that python getsize tries to automatically encode your unicode into str using what python think is the best encoding for filesystem which is ascii and it fails to do that due to special characters.

If you know the filesystem encoding, I suggest you encode the path *before* you pass it to subliminal. It should do the job as subliminal can work with native strings since 1.1.0 (*not released yet*, see commit https://github.com/Diaoul/subliminal/commit/79fb0442dd30ab4207f4a52fae9c3e8efb1eb451).

I suggest you read the following: https://github.com/mitsuhiko/click/blob/master/docs/python3.rst

medariox:
Thanks for the heads up!

You are surely right, the path should be a str not unicode. Anyway, we do not need scan_video in this context, as we are just looking for embedded subtitles and using e.g. GuessIt would result in an unnecessary additional resource consumption. I think this is an acceptable fix for now.

NicoLeOca:
@medariox 
With this fix, should we expect a better accuracy in language recognition or not?

medariox:
@NicoLeOca 
No, this doesn't change anything on that front.

NicoLeOca:
@medariox @Diaoul 
I got this results then:
image

Where are the issues?
SR and I'll be patient until the whole subtitles thing is updated along with subliminal 1.1
or should I raise issues under enzyme repo?

PR #2652, opened 2015-09-29
Add DVDMux & BDMux as valid SDDVD qualities

duramato:
as suggested here http://feathub.com/SiCKRAGETV/SickRage/+9

DVDMux ,BDMux are the same as a DVDRip and BDRip ,but with muxed/embebed subs

### DIFF
--- a/sickbeard/common.py
+++ b/sickbeard/common.py
@@ -270,7 +270,7 @@ class Quality:
             ret = Quality.SDTV
         elif checkName([r"web.?dl|webrip", r"xvid|x26[45]|h.?26[45]"], all) and not checkName([r"(720|1080)[pi]"], all):
             ret = Quality.SDTV
-        elif checkName([r"(dvdrip|b[rd]rip|blue?-?ray)(.ws)?.(xvid|divx|x26[45])"], any) and not checkName([r"(720|1080)[pi]"], all):
+        elif checkName([r"(dvd(rip|mux)|b[rd](rip|mux)|blue?-?ray)(.ws)?.(xvid|divx|x26[45])"], any) and not checkName([r"(720|1080)[pi]"], all):
             ret = Quality.SDDVD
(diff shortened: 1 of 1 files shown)


miigotu:
aka if we add this then I would start getting subbed or language muxed releases downloaded? Most of our users don't want foreign languages.

duramato:
they already would download ,ist the same case as x265,if they dont whant them then add the language code to ignered list ,like SWE,ITA,RU,etc

miigotu:
what do you mean they already would download? what do they get detected as now?

WebSpider:
It should at lease be configurable, imo. I'm not a fan of non-original audio either, that's what subs are for.

-----Original Message-----

duramato:
@miigotu unknown ,also most muxed releases already are detected normally as they only have the langage code in the name, like Survivor's Remorse S02E02 VOSTFR HDTV

duramato:
@WebSpider unless you configure a ignored word ,it will also download they SR is ,this just adds correct recognision

WebSpider:
Yeah I do have quite some blacklisted words, that's true

-----Original Message-----

miigotu:
Unknown was not originally included in `All` quality setting for exactly this reason, and soon will not be included again.

We need to update the default ignored words again btw. I have it working nice on my main install, but it took quite awhile to get it to be ok.

duramato:
@miigotu Ignored words exist for this reason 
thake this release ```Hannibal.s01e01.WEB-DLRip.XviD.Rus.Eng.BaibaKo.tv``` unless it you have rus added to the ignored words SR will download and detect the quality

miigotu:
It should be in line 286 and 288 also then, lets try and be uniform.

duramato:
@miigotu if you dont mind ,ill do the blu ray part on another pr

PR #2654, opened 2015-09-29
Fix for when seeders/Leechers is "---" instead of a number

duramato:
fixes https://github.com/SiCKRAGETV/sickrage-issues/issues/3028


miigotu:
that won't work. you can't compare ints and strings.

you need helpers.tryInt

duramato:
@miigotu like this ?```
                        seeders = helpers.tryInt(item['seeders']) or 0
                        leechers = helpers.tryInt(item['leechers']) or 0

fernandog:
helpers.tryInt(item['seeders'],0)

duramato:
https://github.com/SiCKRAGETV/SickRage/pull/2655

PR #2655, opened 2015-09-29
Fix for when seeders/Leechers is "---" instead of a number

duramato:
2nd attempt

### DIFF
--- a/sickbeard/providers/extratorrent.py
+++ b/sickbeard/providers/extratorrent.py
@@ -123,8 +123,8 @@ class ExtraTorrentProvider(generic.TorrentProvider):
                         info_hash = item['info_hash']
                         url = item['enclosure']['@url']
                         size = int(item['enclosure']['@length'] or item['size'])
-                        seeders = int(item['seeders'])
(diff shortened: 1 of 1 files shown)


miigotu:
can't check right now because I'm on my phone, but is helpers imported?

duramato:
@miigotu yeah ```from sickbeard import helpers```

miigotu:
ok, merging. next time just make sure there is a space after `,`

formatting matters =P

duramato:
@miigotu does it make the code not work?

miigotu:
No, but it will fail build tests later once we start linting in travis builds, so it is a good idea to start getting in the habit of pep8 (except line length if its ridiculous)

duramato:
@just got this 
```    title = item['title']
TypeError: string indices must be integers```
the same fix ,would work for that wont?

miigotu:
no. title is not an int. I guess you need a try that has a continue when it causes an exception.

miigotu:
@duramato better to reference the issue from the PR description (editted for you)

PR #2658, opened 2015-09-30
Fix SiCKRAGETV/sickrage-issues/issues/2769

fernandog:
All settings values from webserve can have 3 states: None, 0(False) or 1(True)

webserve do all checks to see if there's any diff value in the selected shows. If all values are the same then setting value is None

when we bool(value) and value is None it become false. Which is wrong.
This PR fixes it

@OmgImAlexis do you have another solution other than compare == 1 / == 0  / == None ?

### DIFF
--- a/gui/slick/views/manage_massEdit.mako
+++ b/gui/slick/views/manage_massEdit.mako
@@ -83,34 +83,24 @@
     </div>
 </div>
 
-% if anyQualities + bestQualities:
-<% isSelected = ' selected="selected"' %>
-<% isEnabled = isSelected %>
-<% isDisabled = isSelected %>
-% if archive_firstmatch_value:
-    <% isDisabled = '' %>
-% else:
-    <% isEnabled = '' %>
-% endif
 <div class="optionWrapper clearfix">
 <span class="selectTitle">Archive on first match</span>
     <div class="selectChoices">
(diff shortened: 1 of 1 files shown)


fernandog:
I tried to change the "Root Directories *" div but I couldn't. I think there's a better way to show instead of using "==>"

Also if orig_root_dir_ is the same as new_root_dir_ them don't show the later.

OmgImAlexis:
@fernandog are you talking about js or python in the mako files? In js we can do something like this.

````
var variable_name = "None";
parseInt(variable_name) ? true : false; // false

var variable_name = "1";
parseInt(variable_name) ? true : false; // true

var variable_name = "0";
parseInt(variable_name) ? true : false; // false
````
Then we can just need to make sure we only get one of those as a response.

Then instead of `variable_name === "True"` we can do `(parseInt(variable_name) ? true : false) === true)`. That'll convert the variable into true/false and then check if it equals true.

fernandog:
@OmgImAlexis python! No JS there

fernandog:
@miigotu comments?

miigotu:
This looks fine. Was a bit wonky understaning that archive_first_match thing with the wrong check for qualities. (should only have checked for bestQualities originally)

I think this is how it should be.
There should be a # Note: in the code that says the possible values.

PR #2659, opened 2015-09-30
Fix SiCKRAGETV/sickrage-issues/issues/2905

fernandog:
@miigotu this way instead of return 'error' we return the error message.

### DIFF
--- a/sickbeard/versionChecker.py
+++ b/sickbeard/versionChecker.py
@@ -164,10 +164,10 @@ class CheckVersion:
                     logger.log(u"We can't proceed with the update. New update has a old DB version. It's not possible to downgrade", logger.ERROR)
                     return False
                 else:
-                    logger.log(u"We can't proceed with the update. Unable to check remote DB version", logger.ERROR)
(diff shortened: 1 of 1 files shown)


sickbeard/versionChecker.py
```diff
@@ -164,10 +164,10 @@ def db_safe(self):
                     logger.log(u"We can't proceed with the update. New update has a old DB version. It's not possible to downgrade", logger.ERROR)
                     return False
                 else:
-                    logger.log(u"We can't proceed with the update. Unable to check remote DB version", logger.ERROR)
+                    logger.log(u"We can't proceed with the update. Unable to check remote DB version. Error: %s" % result, logger.ERROR)
                     return False
-            except:
-                logger.log(u"We can't proceed with the update. Unable to compare DB version", logger.ERROR)
+            except Exception as e:
+                logger.log(u"We can't proceed with the update. Unable to compare DB version. Error: %s" % repr(e), logger.ERROR)
```
MGaetan89: What's the difference between `repr` and `ex`? Why use one rather than the other?

miigotu:
`repr` shows everything in that class, `ex` just shows the message

PR #2661, opened 2015-09-30
Restores support to ignore embedded subs, '?' will now be replaced by…

medariox:
… default langauge (see conditions)

'?' will only be replaced with the default language if:
- You have set only one default language
- Only one result for subtitles has been found
- That result is 'und' (undefined)

### DIFF
--- a/sickbeard/subtitles.py
+++ b/sickbeard/subtitles.py
@@ -96,27 +96,28 @@ def subtitlesLanguages(video_path):
     embedded_subtitle_languages = set()
 
     # Serch for embedded subtitles
-    if video_path.endswith('mkv'):
-        with open(video_path, 'rb') as f:
-            mkv = MKV(f)
-            if mkv.subtitle_tracks:
-                for st in mkv.subtitle_tracks:
-                    if st.language:
-                        try:
-                            embedded_subtitle_languages.add(Language.fromalpha3b(st.language))
-                        except BabelfishError:
-                            logger.log('Embedded subtitle track is not a valid language', logger.DEBUG)
+    if not sickbeard.EMBEDDED_SUBTITLES_ALL:
+        if video_path.endswith('mkv'):
+            with open(video_path, 'rb') as f:
+                mkv = MKV(f)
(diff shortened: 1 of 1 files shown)


NicoLeOca:
I can understand why you did that but that's really a shame; we loose information here.

miigotu:
@NicoLeOca no information is lost there. If the sub is und it is und and there is no way to find out what language it is.

NicoLeOca:
@miigotu 
There is a bit of misunderstanding here.
I never complain about having a "?".
I prefer to have it too, at least I know there are embedded subtitles.

But having the language detected properly would be better. I would avoid SR to download additionnal subtitles, most likely less accurate.
My aim is to have the best accuracy of subtitles; because Plex Mobile Sync burns subtitles into the video to sync.

that's why I made so much investigations.
I am not pleased to piss you all off...

and in my investigation to understand how the subtitles detection works, I am stuck here:
https://github.com/SiCKRAGETV/SickRage/pull/2650

NicoLeOca:
@miigotu 
As a second option,
If we want to keep tricking the system, we should consider the embedded subtitles as being in the same language as the audio (and not as the subtitles default language)
As SR rejects realeases including "subs" and so on, the "foreign" embedded subtitles are avoided.

miigotu:
"As SR rejects realeases including "subs" it does not, unless you tell it to.

NicoLeOca:
anyway, if we don't aim at improving the language detection, I am OK with this commit then

miigotu:
und is und thats all there is to it.

NicoLeOca:
is it because enzyme doesn't parse it properly? because mkv info does recognize the language.
if it's due to enzyme, diaoul said to submit an issue there

miigotu:
@NicoLeOca I cant tell you that info without having the file.

NicoLeOca:
@miigotu 
If I pass them to you, would you agree to test?

miigotu:
@NicoLeOca when I get time sure, email me a link. miigotu@gmail

PR #2665, opened 2015-09-30
Change comingEpisodes keys to be aligned with API

fernandog:
@MGaetan89 agree? I checked with ShowsRage


OmgImAlexis:
That's what it was before, right?

fernandog:
didn't understand...

OmgImAlexis:
Don't worry, I was thinking of another section.

MGaetan89:
@fernandog What's the point of asking my opinion if you merge before I answer? XD

The change if fine for me. The only comment I have (had, since I refactored that part), is that "soon" and "later" is not really clear. I guess "soon" should rather be "next week" or something like that (only in the web UI, otherwise it breaks the API).

PR #2666, opened 2015-09-30
Fixes SiCKRAGETV/sickrage-issues/issues/3039

fernandog:


miigotu:
There is no change in that logic.

fernandog:
@miigotu why? first we need to check if is safe to update, if is safe then we start backup, if backup runs ok, then we update

miigotu:
It already does that.

fernandog:
not in the correct order

miigotu:
You didnt change any order.
```diff
def run_backup_if_safe(self):
-        return self.safe_to_update() is True and self._runbackup() is True
+        if self.safe_to_update() is True:
+            if self._runbackup() is True:
+                return True
+        return False
```

python will not process the second statement after an `and` keyword if the first statement is false, instead it returns false. If the first statement is true and the second statement is false, it still returns false.
If and only if both statements are true will it return true.

The code you changed it to is evaluated EXACTLY the same, but with more processor instructions (so it is less efficient)

fernandog:
Ok. So my my backup started if my show update queue were full? Any clue?

Are you there thats true with multi-threading?

miigotu:
It wasn't failed because of a full show queue. ANY request locks its own session in cache/sesssions. It is possible you had a previously locked session and just need to clean the cache/sessions

fernandog:
deleted cache/mako. don't have the error BUT the safe checks are not running

I only have:
2015-09-30 12:50:53 INFO     Thread-28 :: Config backup successful, updating...
2015-09-30 12:46:48 INFO     Thread-28 :: Config backup in progress...

But not this logs. We should have them.
logger.log(u"We can proceed with the update. New update has same DB version", logger.DEBUG)
logger.log(u"We can proceed with the update. Post-Processor is not running", logger.DEBUG)
logger.log(u"We can proceed with the update. Shows are not being updated", logger.DEBUG)

PR #2697, opened 2015-10-02
Syncthing uses prefix for sync files, not extension/suffix

miigotu:


SiCKRAGEB0T:

Thanks!

WebSpider:
Why use syncthing as a prefix, and make the suffix configurable?
Making prefix configurable as well would be more consistent

miigotu:
@WebSpider I agree to a point.

This is the only known prefixing syncfile afaik, though. So whether we implement a configurable prefix is something we can discuss.

Should we really add another configuration that no one will use except for one possible prefix? A user configurable setting would save from future maintenance only if another prefix pops up, and it adds to more settings to overwhelm new users.

At least for an immediate fix for the requesting user this was sufficient for now, regardless of what we decide to do.

PR #2739, opened 2015-10-04
Move 'Archive on first match' closer to 'Quality' in editShow

fernandog:
@miigotu @OmgImAlexis 

If user have only 'anyQualities' and then add 'bestQualities' , the 'archive on first match' will only appear in the next page load after user 'submit' changes, isn't?

### DIFF
--- a/gui/slick/views/editShow.mako
+++ b/gui/slick/views/editShow.mako
@@ -65,6 +65,11 @@ This will <b>affect the episode show search</b> on nzb and torrent provider.<br 
 <%include file="/inc_qualityChooser.mako"/>
 <br />
 
+<b>Archive on first match:</b>
+<input type="checkbox" name="archive_firstmatch" ${('', 'checked="checked"')[show.archive_firstmatch == 1]} /><br>
+(check this to have the episode archived after the first best match is found from your archive quality list)</br>
+<br />
+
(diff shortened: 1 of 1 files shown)


miigotu:
I might be wrong but I dont think that if needs to be around  it

fernandog:
We can remove IF and add JS that check if user has anyQualities and

gui/slick/views/editShow.mako
```diff
@@ -115,13 +122,6 @@ This will <b>affect the episode show search</b> on nzb and torrent provider.<br
 (check this if you wish to use the DVD order instead of the Airing order. A "Force Full Update" is necessary, and if you have existing episodes you need to move them)
 <br/><br/>
 
-% if anyQualities + bestQualities:
-<b>Archive on first match:</b>
-<input type="checkbox" name="archive_firstmatch" ${('', 'checked="checked"')[show.archive_firstmatch == 1]} /><br>
-(check this to have the episode archived after the first best match is found from your archive quality list)</br>
```
miigotu: this note is not valid, archive on first match is going to work for initial qualities. the check for any+best isn't needed, just show it as an option Always

fernandog: Ok

fernandog:
ready to merge

PR #2742, opened 2015-10-05
Added bower support

OmgImAlexis:
I'd like to move over to bower so we don't have libraries installed that're out of date, etc. this will also help with SiCKRAGETV/sickrage-issues/issues/2502


miigotu:
Need to document here how this is used, so that the other developers know what it is for, no matter what happens.

OmgImAlexis:
Alright, what kind of documentation is needed? By the way currently it's not doing anything I'm just adding the required libraries for when we switch to bower as we've got a lot of custom stuff that just isn't needed anymore.

OmgImAlexis:
The way we use [Bower](http://bower.io/) is that all js libraries are installed using `$ bower install <package_name> --save`, that installs everything the packages need to work into `/gui/slick/js/bower_components/` so we can minify and concatinate all of the js libraries into a single file for production use.

To search for a package just use `$ bower search <package_name>`, then verifiy the package is the correct one you want and run `$ bower install <package_name> --save` to save it to the `bower.json` file.

OmgImAlexis:
Wiki page added https://github.com/SiCKRAGETV/sickrage-issues/wiki/How-to-setup-a-development-enviroment

PR #2748, opened 2015-10-05
Add more history page limits

duramato:


SiCKRAGEB0T:

Thanks!

MGaetan89:
Instead of adding new options every time, why not use an input with its type set to `number`?

OmgImAlexis:
This probably isn't a good idea since this may crash SR for users with Raspberry Pis and honestly shouldn't be needed. If you need more logs with using less in cli or change the number in the address bar. @fernandog @miigotu please close.

fernandog:
@OmgImAlexis I tested 'all' in my rpi and not an issue. runned fast (like 20s). As it's an option we can accept.

@miigotu ?

duramato:
@OmgImAlexis if it can crash ,then they shouldnt use
same goes for the ALL setting

OmgImAlexis:
I guess as long as it's not going to crash any low powered devices and like in that case like @MGaetan89 said we should have an input box instead.

miigotu:
I think it should be a hard set limit and have buttons to switch pages like 0-100, 100-200, etc.
But thats just me.

OmgImAlexis:
@miigotu we could add that to the bottom of the page, like left and right arrow.


<-- 100                              300 -->

duramato:
@OmgImAlexis i think that would be good as an addition to the way it is now

fernandog:
I like low value: 10 and 25
I dont mind really old history

miigotu:
@OmgImAlexis thats what I was thinking exactly, but probably at the top of the logs too.

PR #2756, opened 2015-10-05
Fix bug causing redownload of download episode if quality is below th…

tene2:
…e highest possible quality (even if archive on first match).

In theory downloaded episode in "good enough quality" should be automatically archived, however this change prevent episode to be redownloaded if they're still in downloaded.


duramato:
@tene2 you need to PR to the develop branch

PR #2764, opened 2015-10-06
Fix bug causing redownload of download episode if quality is below th…

tene2:
…e highest possible quality (even if archive on first match)

In theory downloaded episode in "good enough quality" should be automatically archived, however this change prevent episode to be redownloaded if they're still in downloaded.


miigotu:
I see another bug there and Id like to rework that section now that I see it, but I will merge this first.

tene2:
Cool. If you explain maybe I can help.

I was basically trying to understand where this "archive on first match" magic should happen, without success so far.

PR #2779, opened 2015-10-07
Removed TVRage network and unused logos.

neoatomic:
Now only TheTVDB logo's are left. Totally saving 1.1MB
Used this recent list of network names. http://pastebin.com/pAsqN2Dm

Still almost 600 TheTVDB logos are missing. http://pastebin.com/LsRvt2NA


SiCKRAGEB0T:

Thanks!

WebSpider:
Reading that we are missing 600 logos made me think: Is it really a good solution to include the logos in a theme, or would a GitHub repo that we cache a local copy of be better? (such as white/blacklist and network times)

OmgImAlexis:
@WebSpider they should still be included but I think it'd be a better idea to have them out of the theme directory so just in `gui/image/`. The same goes for other images that aren't theme specific.

WebSpider:
Do we really need 600+ binary files sitting in the repo that a typical user may use only, say, 20 of, for decorational purposes?

OmgImAlexis:
@WebSpider maybe we could zip them and have a command to unzip? @MGaetan89 was looking at adding them all to a sprite sheet so we only had a single file instead of 600+.

neoatomic:
The majority of those 600 are exotic networks that are not worth adding.
The best solutions would be if thetvdb added those logos to there API. They are currently working on a new API, so might be an idea to see if they are interested in that.

PR #2789, opened 2015-10-08
Added srRoot to apibuilder

OmgImAlexis:
Added srRoot to apibuilder otherwise it doesn't work at all if you've got a web_root set.

### DIFF
--- a/gui/slick/js/apibuilder.js
+++ b/gui/slick/js/apibuilder.js
@@ -5,7 +5,7 @@ $(document).ready(function() {
         var profile = $('#option-profile').is(':checked');
         var targetId = $(this).data('target');
         var timeId = $(this).data('time');
-        var url = $('#' + $(this).data('base-url')).text();
+        var url = srRoot + $('#' + $(this).data('base-url')).text();
         var urlId = $(this).data('url');
 
         $.each(parameters, function (index, item) {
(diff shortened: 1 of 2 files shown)


SiCKRAGEB0T:

Thanks!

MGaetan89:
Do you really need to set all those meta and include two JS files just to set a variable?
That looks kind of counter-effective for me.

OmgImAlexis:
@mgaetan89 this was just a quick fix. I'm redoing the JS and mako files this/next week. For now I've just copied all the stuff from main.mako

MGaetan89:
It's a bit hard to go twice over the same (already done) file for such a
small change. But why not.

OmgImAlexis:
@MGaetan89 I'm looking at splitting the mako files up more this will also mean you can use a layout file for apibuilder and only include the actual body, etc. in it like the other files.

MGaetan89:
I could have done it with the main.mako file, no?
I tried to to keep the apibuilder file simple as it's a single page.

OmgImAlexis:
@MGaetan89 yes and no, I'm guessing you'll want to header different in the apibuilder, I'm still looking into the best way to change that. I'm assuming we can just do the same as the blocks for scripts, etc. but add content into the original block. So if you don't have a block for header in your child view it'll keep the "default", if you do it'll override it.

PR #2828, opened 2015-10-11
Anime: Added leading zero search string for single digit episode numbers

HenrikPoulsen:
I noticed that SickRage cannot find episodes for shows in for example NyaaTorrents if the episode number is 1-9.
An example i used is Lucifer comet. Currently sickrage will generate the url http://www.nyaa.se/?sort=2&term=Comet+Lucifer+1&cats=1_0&page=rss&order=1 for this show. And as you can see that returns no results.
So I made it add an additional search string when the episode scene number is single digit so it will still use the old search string as well as the following one http://www.nyaa.se/?sort=2&term=Comet+Lucifer+01&cats=1_0&page=rss&order=1 which gives all results.
I left the original search string since not all groups use a leading zero.

### DIFF
--- a/sickbeard/providers/generic.py
+++ b/sickbeard/providers/generic.py
@@ -603,7 +603,13 @@ class TorrentProvider(GenericProvider):
                 ep_string += str(ep_obj.airdate).replace('-', '|') + '|' + \
                         ep_obj.airdate.strftime('%b')
             elif ep_obj.show.anime:
-                ep_string += "%i" % int(ep_obj.scene_absolute_number)
+                scene_number = int(ep_obj.scene_absolute_number)
+                # If the episode number is 1-9 we add a leading zero since the episode number usually starts with a 0
+                # in that case so we search for both variations to maximize our results
+                if scene_number < 10:
+                    ep_string_leading_zero = "%s0%i" % (ep_string, scene_number)
+                    search_string['Episode'].append(ep_string_leading_zero)
+                ep_string += "%i" % scene_number
             else:
                 ep_string += sickbeard.config.naming_ep_type[2] % {'seasonnumber': ep_obj.scene_season,
                                                               'episodenumber': ep_obj.scene_episode}


SiCKRAGEB0T:

Thanks!

miigotu:
We cannot add another search string like that. You must come up with a better solution.

This way it would hit the providers twice for each exception, effectively doubling the strain the entire userbase of SR puts on nyaa.

HenrikPoulsen:
Is it more acceptable then to simply change it to always pad it with a zero for single digit episodes? It is pretty much standard to always name anime episodes that way, only kept it to still cover the corner cases

miigotu:
I would check nyaa for majority of results, see if they are 1, 01, or 001. I would have thought 001 was most prevalent.

fernandog:
why not remove episode at all? SR will parse all results and only fetch the episode you want

HenrikPoulsen:
Just searching 1 gives 16k hits (you also get lots of false positives for Season 1 etc), 01 gives 86k, 001 gives 1.8k.

fernandog:
@miigotu @HenrikPoulsen can you confirm it accepts OR using "|" ?

http://www.nyaa.se/?sort=2&term=Comet+Lucifer+01|02&cats=1_0&page=rss&order=1

episodes 01 and 02 - just to show both results return

HenrikPoulsen:
I can take a look. It's not just Nyaa which has this issue. I have the same problem with a NewzNab provider, so one would need to check most providers which have anime to see if they support it. If it works then I agree that's a nicer way to do it

fernandog:
We can add self.allowOR

miigotu:
@fernandog I want the search_options to be a dict of options, because there are more than just OR settings for search_strings.

fernandog:
Even for OR in the search_setting itself?

Comet+Lucifer+01|02

miigotu:
yes

fernandog:
@miigotu but for now can't we add the | to search both 01 and 1 ?

miigotu:
@fernandog how would you do that when we use a shared _get_episode_search_strings.

Instead of workarounds we need to do it properly.

HenrikPoulsen:
How about using the same _get_season_search_strings as tokyotoshokan.py? It seems to result in better search strings anyway and it also adds the leading zero.

miigotu:
@HenrikPaulsen the search strings are being normalized and moved out if individual providers. Adding it to nyaa is counter productive.

HenrikPoulsen:
OK good to know. Thanks :)

PR #2838, opened 2015-10-11
Replace SSL Error with a url with information. Disable issue submissi…

fernandog:
…on of such errors

@miigotu can I do line 152?


miigotu:
You don't need to do the skip SSL part. It isn't in the error viewer anymore

fernandog:
@miigotu done! thanks

PR #2888, opened 2015-10-15
Added Scene Quality option, making sure all codecs are detected

medariox:
Won't use regex for 10 elements. Please close if not okay with that.

### DIFF
--- a/sickbeard/tv.py
+++ b/sickbeard/tv.py
@@ -2118,19 +2118,19 @@ class TVEpisode(object):
             for codec in codecList:
                 if codec in name.lower():
                     found_codec = codec
-                    if found_codec is codecList[0]:
+                    if found_codec == codecList[0]:
                         found_codec = 'XviD'
-                    elif found_codec is codecList[4] or codecList[7]:
+                    elif found_codec == codecList[4] or found_codec == codecList[7]:
                         found_codec = codecList[1]
-                    elif found_codec is codecList[9]:
-                        found_codec = codecList[2]
-                    elif found_codec is codecList[5] or codecList[8]:
+                    elif found_codec == codecList[6]:
                         found_codec = codecList[3]
+                    elif found_codec == codecList[5] or found_codec == codecList[8]:
+                        found_codec = codecList[2]
 
             if found_codec:
-                return found_codec
+                return " " + found_codec
             else:
-                return ""
+                return " " + codecList[1]
 
         def release_group(show, name):
             if name:
--- a/sickbeard/tv.py
+++ b/sickbeard/tv.py
@@ -2112,6 +2112,26 @@ class TVEpisode(object):
                 name = helpers.remove_non_release_groups(helpers.remove_extension(name))
             return name
 
+        def release_codec(name):
+            codecList = ['xvid', 'x264', 'x265', 'h264', 'x 264', 'x 265', 'h 264', 'x.264', 'x.265', 'h.264']
+            found_codec = None
+            for codec in codecList:
+                if codec in name.lower():
+                    found_codec = codec
+                    if found_codec is codecList[0]:
+                        found_codec = 'XviD'
+                    elif found_codec is codecList[4] or codecList[7]:
+                        found_codec = codecList[1]
+                    elif found_codec is codecList[9]:
+                        found_codec = codecList[2]
+                    elif found_codec is codecList[5] or codecList[8]:
+                        found_codec = codecList[3]
+
+            if found_codec:
+                return found_codec
+            else:
+                return ""
+
         def release_group(show, name):
             if name:
                 name = helpers.remove_non_release_groups(helpers.remove_extension(name))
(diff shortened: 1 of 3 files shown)


SiCKRAGEB0T:

Thanks!

sickbeard/tv.py
```diff
@@ -2177,7 +2200,7 @@ def release_group(show, name):
             '%AB': '%(#)03d' % {'#': self.absolute_number},
             '%XAB': '%(#)03d' % {'#': self.scene_absolute_number},
             '%RN': release_name(self.release_name),
-            '%RG': rel_grp[relgrp],
+            '%RG': rel_grp[relgrp].upper(),
```
miigotu: ????

medariox: %RG would just output the group name as it is in the file name (doesn't matter if capitalized or not), but the capitlized %RG should output the capitalized group name (like the other parameters). Now the user has the option to output it capitalized %RG or %rg lowercase.

miigotu:
release group is supposed to stay as it came. `IMMERSE` is not the same as `iMMERSE`

fernandog:
agree with miigotu

medariox:
If all torrents were perfectly tagged, there wouldn't be any need for all of this. Sadly some torrents come as `iMMERSE` other as `IMMERSE` and some others as `immerse`. So what do you suggest? Maybe add an option to keep the "orginal" (aka as it comes from the file name)?

miigotu:
Those are all different people uploading them. Real ones, untampered with, are iMMERSE.

The distinction needs to be left there.

miigotu:
Kinda strange you started this because you didnt want to lose the codec information but are ok with putting incorrect codec info or incorrect group info.

medariox:
Hm? What are you talking about again?
If someone uploads a release like this:
`my.tv.show.112.hdtv-immerse`
SR should find a way to correctly parse the above file name to a proper release name.
e.g. `%S.N.S%0SE%0E.%RT.%SQ.N-%RG` would be like a proper scene release name, except for the group name, which will stay lowercase, breaking the whole point of the name pattern:
`My.Tv.Show.S01E12.HDTV.x264-immerse`
Since any other patter can be either capitalized (%RG) or lowercase (%rg), I don't see why it shouldn't be possible for the group info. It just doesn't make any sense to not add the option.
If for you having `IMMERSE` instead of `iMMERSE` is having "incorrect group info", I'm perfectly fine with that. The codec is only "incorrectly" added when there is no way of telling the real codec of the releases (aka it is not in the file name), e.g. `my.tv.show.112.hdtv-immerse` - why don't you understand that?

fernandog:
Im all for give users options. But default should be the default

miigotu:
The point of the name pattern is to PRESERVE information.
If the codec is not in the file name to begin with is should not be added to the result at all.

You cannot force naming to be absolutely 100% conforming or normalized by falsifying information and blatantly lying in the output. Making assumptions or changing the information that is already there does not fit in the equation.

Normalizing show names, season and episode numbers, and even codec and source is OK because they have an indisputably CORRECT format.

For release groups, unless you are going to add a map that includes EVERY release group in the world so that you can format them as is their proper name you cannot change them. These are PROPER NOUNS mixed with the complication of a trademark.

Have you not thought about why it wasnt uppercased or lowercased in the code to begin with? This is why.

fernandog:
The original issue isnt the SR renaming was not considering code even if it exist in file name?

miigotu:
Yes the original issue was codec missing. Ok add a way to get the codec (accurately, and normalize it correctly). But dont add false information or go changing other things in the same PR that are the way they are for a reason, especially when they are unrelated changes.

medariox:
@fernandog @miigotu 
I can add an option to use the capitalized group name (%CRG instead of %RG for example), but I won't change the adding of x264 when there is no codec in the file name, because I get many, many torrents that are like that and Scene Quality pattern will try to make it scene standard, even when some crazy uploader decides that they can strip information from the file name. If you are not okay with that, please close the PR.

fernandog:
Cant you PR the fix only for now??

medariox:
@fernandog 
It has no point as miigotu won't take it and I'm tired of adding features nobody wants, so I will just keep it private for myself.

miigotu:
@fernandog he wants to add x264 blindly to any file we cannot know the codec of.

AKA:
(release name)
some.show.s01e01.hdtv.wmv
becomes:
Some.Show.S01E01.SDTV.x264.wmv

That is the most unacceptable thing about this PR.

fernandog:
@medariox why don't you check the torrentname and/or folder if codec is there? And if is missing in filename add it. no matter if is x264 or x265 or h264 or whatever

fernandog:
@miigotu are you ok with that change?

medariox:
@fernandog 
I can try that, not sure if possible tho. Anyway, I'm doing it only if it gets approved as I have no time to waste anymore.

miigotu:
@fernandog Ill just do it another time. If it is missing in file name its just missing.

fernandog:
but if torrentname has codec, there's no point not adding. the uploader forgot to add after re-upload the torrent/nzb

PR #2919, opened 2015-10-19
Should fix SiCKRAGETV/sickrage-issues/issues/3273

ROpdebee:
Should fix SiCKRAGETV/sickrage-issues/issues/3273
Would use the enddate of previous episodes in the same segment if that segment was already defined. Caused issue when one episode already aired, unaired ones would also show up as overdue.

### DIFF
--- a/gui/slick/views/schedule.mako
+++ b/gui/slick/views/schedule.mako
@@ -207,6 +207,12 @@
         continue
 
     run_time = cur_result['runtime']
+    cur_ep_airdate = cur_result['localtime'].date()
+
+    if run_time:
+        cur_ep_enddate = cur_result['localtime'] + datetime.timedelta(minutes = run_time)
+    else:
+        cur_ep_enddate = cur_result['localtime']
 %>
     % if 'network' == sort:
         <% show_network = ('no network', cur_result['network'])[bool(cur_result['network'])] %>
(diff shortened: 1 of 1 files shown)


SiCKRAGEB0T:

Thanks!

gui/slick/views/schedule.mako
```diff
@@ -233,11 +233,13 @@
...
         % endif
     % elif 'date' == sort:
-        <% cur_ep_airdate = cur_result['localtime'].date() %>
+        <%
+            cur_ep_airdate = cur_result['localtime'].date()
+            cur_ep_enddate = cur_result['localtime'] + datetime.timedelta(minutes = run_time)
+        %>
 
         % if cur_segment != cur_ep_airdate:
             %if run_time:
```
fernandog: @ROpdebee what is the point of this here as you moved line? You should move up too

OmgImAlexis: @fernandog it's still on  LN#236 but they added `cur_ep_enddate = cur_result['localtime'] + datetime.timedelta(minutes = run_time)` after it.

ROpdebee: @fernandog Otherwise it would only set the enddate for the first episode in the segment (ie Today, Missed, ...), and then further down the code it would use the enddate of the first episode of that segment for all episodes in that segment. Thus, when one episode finished airing, all others for that day would get marked in red as well, even when they weren't aired yet, because their enddate would not get set correctly (because that part would get skipped because the day header was already calculated) Only happened on banner/poster view and when sorted on date, other view modes weren't affected.
However I now notice this introduces some code redundancy so let me factor that out first...

miigotu:
@ROpdebee Only issue I have with this is that code is moving out of there into a shared class for webapi and webui, which is why we didn't bother touching this until @MGaetan89 was finished his PR. We've also been avoiding multiple changes to the interface as much as possible until bootstrap4 is merged back to mainline.

Doing work twice is no fun

ROpdebee:
Oh, didn't realize you guys were rewriting this. As far as I can see from @MGaetan89 's PR, it'll fix this issue as well. Ah well, maybe this could serve as a temporary fix, if necessary?

miigotu:
Will have to hear from both @MGaetan89 and @OmgImAlexis for their opinion about it before I could merge, since it effects both of their current tracks. Has already been some of a challenge to get 2 trains on the same track without colliding.

I just want to avoid adding any unnecessary work for either of them because they both already have a heap of stuff they are working on.

That being said, if they give the go ahead :+1:

MGaetan89:
I didn't look too much at the changes, but if someone else manages to fix the bug that I should be fixing, I'm fine with it :P
My PR is already late on `develop`, so I have some work to do on it anyway. I am fine with merging this now if it fixes the issue. I will integrate it in my PR when I have time again (in about 2 weeks).

Just cross-referencing so I don't forgot about this when I work on my PR again https://github.com/SiCKRAGETV/SickRage/pull/2558

miigotu:
@OmgImAlexis ?

PR #2927, opened 2015-10-19
Fix SNATCHED_PROPER and SNATCHED_BEST not changing status to DOWNLOADED

fernandog:
At the first scan, the file and file size are detected. As snatched_proper and snatched_best  status wasn't checked, file size were updated in DB and in the next scan line 721 ("elif not same_file ... ), the if is false as SR detects the same_size = True so ep status stayed forever as SNATCHED_PROPER/BEST

This is the real fix for SiCKRAGETV/sickrage-issues/issues/1074

Working without issues for this show that only downloaded proper and never changed status to downloaded.
```
2015-10-19 10:53:24 INFO     SHOWQUEUE-REFRESH :: Show Les Revenants S02E06 used to be Snatched (Proper) (720p HDTV) but a file exists with quality 720p HDTV so I'm setting the status to Downloaded (720p HDTV)
2015-10-19 10:53:24 INFO     SHOWQUEUE-REFRESH :: Show Les Revenants S02E05 used to be Snatched (Proper) (720p HDTV) but a file exists with quality 720p HDTV so I'm setting the status to Downloaded (720p HDTV)
2015-10-19 10:53:23 INFO     SHOWQUEUE-REFRESH :: Performing refresh on Les Revenants
```

To reproduce the issue, just add this show and enable provider T411
@duramato can you test/confirm, please?

### DIFF
--- a/sickbeard/tv.py
+++ b/sickbeard/tv.py
@@ -726,28 +726,18 @@ class TVShow(object):
 
                 newStatus = None
 
-                # if it was snatched and now exists then set the status correctly
-                if oldStatus == SNATCHED and oldQuality <= newQuality:
-                    logger.log(u"STATUS: this ep used to be snatched with quality " + Quality.qualityStrings[
-                        oldQuality] + u" but a file exists with quality " + Quality.qualityStrings[
-                                   newQuality] + u" so I'm setting the status to DOWNLOADED", logger.DEBUG)
+                # if it was any snatched staus and now a file exists then set the status correctly
+                if oldStatus in (SNATCHED, SNATCHED_PROPER, SNATCHED_BEST) and oldQuality <= newQuality:
                     newStatus = DOWNLOADED
 
-                # if it was snatched proper and we found a higher quality one then allow the status change
-                elif oldStatus == SNATCHED_PROPER and oldQuality < newQuality:
-                    logger.log(u"STATUS: this ep used to be snatched proper with quality " + Quality.qualityStrings[
-                        oldQuality] + u" but a file exists with quality " + Quality.qualityStrings[
-                                   newQuality] + u" so I'm setting the status to DOWNLOADED", logger.DEBUG)
-                    newStatus = DOWNLOADED
-
-                elif oldStatus not in (SNATCHED, SNATCHED_PROPER):
+                elif oldStatus not in (SNATCHED, SNATCHED_PROPER, SNATCHED_BEST):
                     newStatus = DOWNLOADED
 
                 if newStatus is not None:
                     with curEp.lock:
-                        logger.log(u"STATUS: we have an associated file, so setting the status from " + str(
-                            curEp.status) + u" to DOWNLOADED/" + str(Quality.statusFromName(file, anime=self.is_anime)),
-                                   logger.DEBUG)
+                        logger.log(u"Show %s S%02dE%02d used to be %s but a file exists with quality (%s) so I'm setting the status to %s" %
+                        (self.name, season, episode,statusStrings[curEp.status], Quality.qualityStrings[newQuality], 
+                        statusStrings[Quality.compositeStatus(newStatus, newQuality)]), logger.INFO)
                         curEp.status = Quality.compositeStatus(newStatus, newQuality)
 
             with curEp.lock:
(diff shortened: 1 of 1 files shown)


SiCKRAGEB0T:

Thanks!

fernandog:
wondering if is better to do a DB check for episodes with status snatched_proper/best and size not null and set size to null. So in the next show refresh SR will detect episodes as 'downloaded'

fernandog:
@miigotu no comments about this fix?

duramato:
@fernandog i add the show ,enable t411 and do what next?

fernandog:
@duramato let SR snatch episodes. Add POLY to required words. Will snatch proper in the first time
Mark ep 5 or 6 as wanted

fernandog:
But dont apply my fix yet!

miigotu:
So what about when either of these are None? `self.season, self.episode` 
 Also `record is not dirty`  is a DEBUG line, which any db admin in the world will understand (it isnt for anyone else)

You're again mixing changes, and changing things that don't need changed.

fernandog:
Easy. Just revert that.
Lots of issues user open because they see 'record not dirty' but its ok, will revert
Also will add or 0

So the fix itself is ok?

miigotu:
Im not sure about the "fix" itself. The code is expecting the file to exist there.

fernandog:
reverted the dirty and added "or 0"

yes, the file exist but status is snatched_proper/best in DB. and if we don't go into the if, new_status = None and we only change status "if newStatus is not None"

miigotu:
I think it intentionally didnt touch snatched_proper because snatched proper you can have an existing file, AND be in snatched_proper state waiting for a file to be finished downloading in your client. This change would set the file back to downloaded before the proper was downloaded and processed.

With this change a show refresh will break propers

miigotu:
Only PP should be changing from snatched_proper or snatched_best to downloaded.

fernandog:
that code will only change to downloaded IF the file "PROPER" exists in show folder and status is snatched proper

PP runs the same code (tv.py - makeEpFromFile)
I said to refresh just to simplify the test

fernandog:
there's no other code that converts snatched to downloaded. only that one

miigotu:
It sure does, in `postProcessor.process` it calls `new_ep_quality = self._get_quality(ep_obj)` which checks SNATCHED_PROPER and then later sets it to downloaded `cur_ep.status = common.Quality.compositeStatus(common.DOWNLOADED, new_ep_quality)`

fernandog:
well, so it's an issue only with refreshDir

The file will be inside the folder with PP or manually moved.
If PP moved, there's no issue.
If manually moved, it stays in snatch proper forever (because user can choose to not enable PP)

miigotu:
If you dont have PP enabled you should be using nzbtomedia or autoProcessTV scripts which would call PP. Otherwise you need to have your script handle snatched_proper correctly and update the status via webapi.

Breaking something for everyone for an edge case where someone can use SR in a way that it was not intended to be used sounds like a good idea?

fernandog:
can you tell me a case that breaks?
if user are running PP, they are not moving files, so no issue
Also a lot other people don't use PP or nzbtomedia. they manually move.

miigotu:
Ive very clearly explained the case that it breaks, for every user who has an episode in snatched_proper status and the full show update happens at 3am it would end up set as downloaded if the episode was previously downloaded (a non-proper).

This would immediately leave an orphaned download in your PP folder, and a non-proper marked back to downloaded. There is also a potential for a race condition here if the file were to be removed from the client in the meantime before the next proper search, because the exact same proper would be added again. And the cycle repeats.

The change simply cannot be allowed to break SR for 95% of our users just to make a few users not have to process their files properly (or at least handle manually setting the status after they MANUALLY move files)

fernandog:
@miigotu 
you didn't read this:

also the same_file variable handles the last snatch with the existing file in folder
So it will only convert to download the last snatched file

fernandog:
did you look the old code? It's there, converting snatched proper to downloaded. but there is a typo in the if. missing the =
Why don't you accept it's an error?

If it wasn't supposed to convert to snatch_proper, this code wouldn't be here:

```
                elif oldStatus == SNATCHED_PROPER and oldQuality < newQuality:
                    logger.log(u"STATUS: this ep used to be snatched proper with quality " + Quality.qualityStrings[oldQuality] + " but a file exists with quality " + Quality.qualityStrings[newQuality] + " so I'm setting the status to DOWNLOADED", logger.DEBUG)
                    newStatus = DOWNLOADED
```

miigotu:
The missing = is intentional! I keep telling you that. It cannot allow converting snatched_proper of the same quality, only for a redownload that happened to be a proper.

checkQualityAgain is not true unless the end file name has changed, and the same_file code is not used in the first if. The file location would not change between a snatched and snatched_proper.

You still dont understand this code.

fernandog:
but what if my FIRST download was a PROPER?

PR #2938, opened 2015-10-20
Catch URLError when Kodi cant be reached for notifications

miigotu:


SiCKRAGEB0T:

Thanks!

Koal4:
I think this  issue is related to my issue which i reported a few weeks ago and was closed by one of you teammembers.

After this update I noticed that the error messages where less then before but it is still generating errors when it cant detect my Kodi client when the notification option "log errors when unreachable ?" is turned off.

Errors before update;
http://imgur.com/bOROoxx

Errors after this update:
http://i.imgur.com/Fny9Znf.png

Notification options turned on:
http://i.imgur.com/DcpanaC.png

miigotu:
@Koal4 warnings are not errors, they are warnings.

Koal4:
Errors or warnings, what is the difference botg make the user aware that something is wrong ;)

Anyway, for some reason the issue/bug I found is that since this code has applied that I do get lessafter this change it spams less warning messages so so something has 

Atm i have a strong feeling that the issue i'm trying to address is not coming through unfortunately. What I'm trying to say and also in the other topic (which has been closed for the wrong reasons) is that the Kodi notification option "log errors when unreachable ?" is broken.

When turned on this option should log warning / error messages in the log.
When turned off this option should ignore those warning / error messages in the log.

Problem now is that it works like the option is always turned on while it isn't. In my scenario it keeps on spamming warnings/errors that my Kodi client is unreachable when SR grabs new shows while my HTPC is offline( my HTPC isn't turned on always). By deselecting this option it ignore those warnings/errors and keeps the warning log "clean" so when you have real warnings/errors they are easily noticed then. This has worked fine in the past untill some updates in the weekend from  3 - Oct.

I hope it's a bit clear now otherwise it's very easy to reproduce
Turn on Kodi notifications if you haven't (see my screenshot for some of the options)
Turn off your Kodi machine (You can also insert a random IP which isn't used by Kodi) and press test connection . Let it test for a few seconds and refresh the page and you will notice the warnings appear whenever you have that option "log errors when unreachable ?" turned off.

PR #2943, opened 2015-10-21
Try to fix encoding error in error messages

fernandog:
Error: u"error ('Connection aborted.', gaierror(-2, 'Nome ou servi\\xc3\\xa7o desconhecido'))"

helpers.py are already using %r
`logger.log(u"Connection error to getURL %s Error: %r" % (url, ex(e)), logger.WARNING)`

`ss: Converts string to Unicode, fallback encoding is forced UTF-8.`

So why we are doing str() before ss() ?

```
def ss(var):
    """
    Converts string to Unicode, fallback encoding is forced UTF-8

    :param var: String to convert
    :return: Converted string
    """

    var = _to_unicode(var)
```

### DIFF
--- a/sickrage/helper/exceptions.py
+++ b/sickrage/helper/exceptions.py
@@ -36,7 +36,7 @@ def ex(e):
                 fixed_arg = ss(arg)
             else:
                 try:
-                    fixed_arg = u'error %s' % ss(str(arg))
+                    fixed_arg = u'error %s' % ss(arg)
                 except Exception:
                     fixed_arg = None
 


SiCKRAGEB0T:

Thanks!

miigotu:
Because the arg is not str or unicode, it is an int, which doesnt matter if you run str() on it because numbers always are ascii.

miigotu:
This PR will cause exceptions in ss()

fernandog:
so from where the "Nome ou servi\xc3\xa7o desconhecido" comes ?

miigotu:
From it falling back to utf-8. Do not change anything to fix that one problem, we tried to fix that before and succeeded but it broke encoding for everyone else. We also tried to fix it one time before that and broke everyone's database. Leave it alone.

duramato:
@miigotu wab "ftfy"?

PR #2953, opened 2015-10-22
Schedule/ComingEpisodes was not considering snatched_best and snatche…

fernandog:
…d_proper

so episodes with that status appeared as WANTED in /schedule

@miigotu @MGaetan89


fernandog:
shouldn't ignore SKIPPED also?!

fernandog:
@miigotu done

PR #2957, opened 2015-10-22
Fix SiCKRAGETV/sickrage-issues#3416

ROpdebee:
Also select UNAIRED episodes whose airdates are in the past but have not yet changed status because of timezone differences

### DIFF
--- a/sickrage/show/ComingEpisodes.py
+++ b/sickrage/show/ComingEpisodes.py
@@ -19,7 +19,7 @@
 import sickbeard
 
 from datetime import date, timedelta
-from sickbeard.common import IGNORED, Quality, WANTED
+from sickbeard.common import IGNORED, Quality, WANTED, UNAIRED
 from sickbeard.db import DBConnection
 from sickbeard.network_timezones import parse_date_time
 from sickbeard.sbdatetime import sbdatetime
@@ -108,9 +108,9 @@ class ComingEpisodes:
             'AND s.indexer_id = e.showid '
             'AND airdate < ? '
             'AND airdate >= ? '
-            'AND e.status = ? '
+            'AND e.status IN (?,?) '
             'AND e.status NOT IN (' + ','.join(['?'] * len(qualities_list)) + ')',
-            [today, recently, WANTED] + qualities_list
+            [today, recently, WANTED, UNAIRED] + qualities_list
         )
 
         results = [dict(result) for result in results]


SiCKRAGEB0T:

Thanks!

fernandog:
If they dont appear then itsn an error in sql or something.

Can you show a example before/after screenshot?

ROpdebee:
Check the issue for the before screenshot. Issue here was that the airdate in the db is set to the airdate (just the date, not the time) of the network's timezone (so, for the south park ep it was 21/10), but when converted to local time it would air on 22/10 at 4am. The db select only selected the episodes of which the airdate was >= today and episodes which aired in the past but were WANTED. UNAIRED episodes with a network timezone in the past would be skipped until they actually aired and got set to WANTED.
Screenshot after fix (kinda had to cheat with the system time to reproduce the issue of yesterday)
screen shot 2015-10-22 at 00 55 48

fernandog:
you are changing only display. are you sure SR is searching for them?

fernandog:
@ROpdebee maybe need to change this:

https://github.com/SiCKRAGETV/SickRage/blob/develop/sickbeard/dailysearcher.py#L55-L58

ROpdebee:
SR shouldn't search for them yet, that all works fine, there's no issue there. The problem is that some episodes disappear from the schedule display because the database selection doesn't include those episodes because their airdate is stored based on the airdate from the indexer (which is the network time in the US, for example) but they get selected from the database through comparing the airdate in the db with the local date (in a different timezone, so a different day). I'll try explaining a bit more, bear with me

ROpdebee:
So let's say I'm in Belgium, and it's currently 11pm on Wed. I have South Park active in my show list, it's set to air on Wed at 10pm in the US. It's not set to wanted yet because, when converted to my timezone, it should air on Thu at 4am. This is the situation in the before screenshot. Everything is correct.

Now, the database stores this airdate as Wed, which it got from theTVDB, but on Thu from midnight until 4am (local time in Belgium) the episode will not get selected from the database because it only selects episodes with airdate >= today's date (**local** time, so thursday) and WANTED episodes aired in the recent past. But the episode we're talking about here hasn't yet aired, so it's not yet set to wanted, and doesn't get selected from the database. Once we're past 4am local time, it'll get set to WANTED and it will display normally, but between midnight and the airdate it won't get displayed (but doesn't need to be wanted).

Hope this cleared it up somehow, if it doesn't, then don't break your head over it, it's a rare and hard to reproduce issue.

PR #2994, opened 2015-10-26
WIP Get response text from clients

fernandog:
@miigotu 
Haven't tested yet. Only changed transmission for now.
What do you think about this solution?

### DIFF
--- a/sickbeard/clients/generic.py
+++ b/sickbeard/clients/generic.py
@@ -34,11 +34,6 @@ class GenericClient(object):
 
         logger.log(
             self.name + u': Requested a ' + method.upper() + ' connection to url ' + self.url + ' with Params= ' + str(
-                params) + ' Data=' + str(data if data else 'None')[0:99] + (
-            '...' if len(data if data else 'None') > 200 else ''), logger.DEBUG)
-
-        logger.log(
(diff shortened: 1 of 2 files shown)


SiCKRAGEB0T:

Thanks!

fernandog:
https://github.com/SiCKRAGETV/sickrage-issues/issues/1382

fernandog:
maybe if _request fail we should check here too

```
        if self._request(method='post', data=post_data):

            if self.response.json()['result'] == "success":
                return (True, self.response.json()['result'])
            else:
                return (False, self.response.json()['result']
        else:
            return (False, "error")
```

miigotu:
clients.generic is broken with this PR

fernandog:
Its WIP. What about the solution?

miigotu:
It changes a hella a lot more than it needs to. Only _add_torrent_file and _add_torrent_uri needs changed, to return the status_code instead of true or false.

fernandog:
But the others requests can fail too and return others messages

fernandog:
@miigotu should I continue changing other clients?

miigotu:
No, this doesnt fix the real problem, and this PR is completely fubar.

fernandog:
ok. how to fix this then? know a way?

PR #3017, opened 2015-10-28
Added the ability to configure a providers minimum cache time setting.

gsly:
Added the ability to configure a providers minimum cache time setting.
Changed ThePirateBay and KickAssTorrents providers to use this new
configuration attribute.

### DIFF
--- a/gui/slick/views/config_providers.mako
+++ b/gui/slick/views/config_providers.mako
@@ -419,6 +419,23 @@ $('#config-components').tabs();
                         </div>
                         % endif
 
+                        % if hasattr(curTorrentProvider, 'mincachetime'):
+                        <div class="field-pair">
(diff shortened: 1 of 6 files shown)


SiCKRAGEB0T:

Thanks!

duramato:
@gsly what if a user sets it to 1?
It will pwn the server, wont it?

gsly:
@duramato I never tried setting it to 1, but I think it only comes into play when the daily search happens. Probably should have a minimum value on the user interface.

duramato:
@gsly when need to provent that,maybe and logic if it is less than 10m default to 10min ?

miigotu:
We cannot allow it to be less than what we have set. The mintime is set per provider requirements and at their request. We absolutely cannot allow the user to lower what we have,

PR #3019, opened 2015-10-29
Lint sbdatetime and fix some problems where returning within finally …

miigotu:
…block can swallow exceptions

allow setting eps to wanted with daily searcher while still airing
make sure to check local time for comparing air times in daily search
Check that show.airs and show.network before trying to parse_date_time
Improved file timestamp modifying code

### DIFF
--- a/sickbeard/dailysearcher.py
+++ b/sickbeard/dailysearcher.py
@@ -79,19 +79,15 @@ class DailySearcher():
                 logger.log(u"ERROR: expected to find a single show matching " + str(sqlEp['showid']))
                 continue
 
-            try:
-                end_time = sbdatetime.sbdatetime.convert_to_setting(network_timezones.parse_date_time(sqlEp['airdate'], show.airs,
-                                                             show.network)) + datetime.timedelta(
-                    minutes=helpers.tryInt(show.runtime, 60))
-                #Keep this for future debug
-                #logger.log(u"Show %s ends at %s and now it is %s. Runtime is %s and airs %s" % (show.name, end_time, curTime, show.runtime, show.airs),logger.DEBUG )
+            if show.airs and show.network:
+                # This is how you assure it is always converted to local time
+                air_time = network_timezones.parse_date_time(sqlEp['airdate'], show.airs, show.network).astimezone(network_timezones.sb_timezone)
 
-                # filter out any episodes that haven't aried yet
-                if end_time > curTime:
+                # filter out any episodes that haven't started airing yet,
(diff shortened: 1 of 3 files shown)


sickbeard/dailysearcher.py
```diff
@@ -79,19 +79,15 @@ def run(self, force=False):
...
-                # filter out any episodes that haven't aried yet
-                if end_time > curTime:
+            if show.airs and show.network:
+                # This is how you assure it is always converted to local time
+                air_time = network_timezones.parse_date_time(sqlEp['airdate'], show.airs, show.network).astimezone(network_timezones.sb_timezone)
+
+                # filter out any episodes that haven't started airing yet,
+                # but set them to the default status while they are airing
+                # so they are snatched faster
+                if air_time > curTime:
```
fernandog: @miigotu why did you changed the logic here?

"set them to the default status while they are airing": so as soon as DAILYSEARCHER set to WANTED, SEARCHQUEUE will immediately starts to search for them while show is still airing. Makes no sense to make to searchs while show hasn't finished yet.

miigotu:
It does make sense. So what if it sets it up to "runtime" early, it will get the episode faster this way.

fernandog:
My concern is hitting providers 40-60 min before shows finished aired. Also there's a 3-5 minutes to episodes reaches trackers. 

The runtime early it's a rare thing. We must use the official airtime. 

It snatch really faster with dailysearcher to 10min. I my experience it will always snatch 5-10 after it **reaches** the provider

I would add this as an option like "Search before airtime" - default off
https://github.com/SiCKRAGETV/sickrage-issues/issues/2805

duramato:
@miigotu i agree with fernando on this one, what could be done is set it to wanted like 5~10 min before it airs, so that
Imagnine it ends at 22:10
daily search runs at 22:09
It probably wont catch the release
BUT
if it is set to wanted at like 22:05 it will catch it 

but if a ep has a runtime of 1h, there isnt any need to have it wanted at 21:10 ,when it will finish airing at 22:10 
Also have it as a option could be a good option too

miigotu:
@duramato SB and other forks set it to wanted about 24 hours in advance. There is no reason we cannot set it to wanted at air time instead of end time so that episodes are snatched faster and the schedule and backlog updates properly.

Also, the code previous to this was incorrect and sometimes compared datetime.datetime.now() in local timezone with airdate in the network timezone and set the show to wanted hours before or hours after the show aired.

This code change does nothing but make it steady and predictable and always works. I'm not going to change it.

PR #3024, opened 2015-10-29
Always check scene exceptions when parsing files or names to shows/ep…

miigotu:
…isodes

@pkoutsias


miigotu:
@pkoutsias just gonna merge this, can'tm find any potential problems. If it needs reverted later so be it.

PR #3035, opened 2015-10-30
Improved BDRip recognition and indentation, removed WEBRip support (a…

medariox:
…s SDDVD can never be WEBRip)

### DIFF
--- a/sickbeard/common.py
+++ b/sickbeard/common.py
@@ -139,16 +139,16 @@ class Quality:
                       FULLHDBLURAY: "1080p BluRay"}
 
     sceneQualityStrings = {NONE: "N/A",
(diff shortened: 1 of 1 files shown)


duramato:
@medariox and what will happen to sddvd releases?

medariox:
@duramato
Nothing will change to SDDVD releases, they will be recognized as usual. This is mostly just cosmetic and removes useless code for the "scene quality" naming pattern.

duramato:
I mean for the naming of sddvd?

medariox:
Nothing changed! Just if you have "scene quality" as naming pattern it won't try to check "WEBRip" in the file name. This has no influence at all on SDDVD naming or any other naming pattern.

PR #3041, opened 2015-10-31
Added www.newpct.com provider

CristianBB:
Added www.newpct.com provider, prepared to seek series with Spanish
dubbing.
I've added two new options on provider's config:

- For spanish torrents: This option allows to make show searchs from
this provider ONLY if "show's config => Info Language" is marked as
Spanish. In this way we avoid search from this supplier for shows that
we have configured in VOs or other languages.

- Append identifier: Append an identifier to every episode snatched by
this provider. Usefull in combination with "Required Words" on show's
configuration if you want to download certain shows only from this
provider. In this way, if we have other providers, we avoid to snatch
torrents from other providers to prevent that episodes are downloaded in
other languages, so it is necessary that "Required Words" on show's
configuration are the same that we define here.

I hope that the changes are of your pleasure :)

CristianBB


SiCKRAGEB0T:

Thanks!

miigotu:
Few issues I have with this one.

First, why override _get_episode search strings? This can surely be avoided somehow, and it is highly discouraged to override the search strings. You will have to to a replace on the search string in _doSearch for both searching and the results. temporada and capitulo just wont do to match in post processing.

Second, not only does this provider provide size which is not parsed into the item result, the provider does not provide seeds, leechers, hash, which makes itm highly undesirable as a provider in sickrage

CristianBB:
Hi miigotu,
In response to your questions:

- After many test with www.newpct.com search, it does not correctly process the search by episode number (always returns all results to be reported regardless of the number of episode). you can do the test by entering "The strain" and "the strain 1x01", which returns results have nothing to do with the informed chapter ... In fact I had to implement an algorithm to form a chain type: show name + season + episode so SickRage can recognize results returned (newpct shows are typed as: The Strain - Season 1 [HDTV] [Cap.109] [Spanish Castilian] and none of the implemented algorithms can recognize it).

- I have checked and I can include the process of file size as the page returned as results, but it does not report the number of seeds and leechers so that function could not be implemented ...

It really is a shame, but the Spanish trackers have a very particular way of naming the series, so I added the option to only use this provider in the series that you want to download in Spanish to avoid an extra proccess.

miigotu:
If it doesnt make a different whether 1x1 or w/e is there then there is no reason to override _get_episode_search strings right?

CristianBB:
Ok course, 
I just thought there would be less chances of getting invalid results if removed the rest of the search string

Allow me a few minutes and will remove the code _get_episode_search

CristianBB:
Deleted,
I'll try to parse the torrent size this weekend, please tell me anything.

I have pending a change to add a new regular expression on file "regexes" that I'd like to discuss with you, to parse correctly filenames with format: "Show Name - Random text. [Quality]. [Cap.102].Random text"

miigotu:
Adding a new regex would be the best solution I think, even though I hate touching the regex that usually work ok as is =P

CristianBB:
Hi miigotu,
I've been working hard on that, let me tell you:

- Finally I've implemented the new regexe to process correctly show names with format: "Show Name - Random text. [Quality]. [Cap.102].Random text"
It's working OK on position that I've used (cause of scoring system), even parsing files with autoprocess and manual proccess.

- I've added torrent's size.

- There is a change that you should review, on file  "sickbeard/providers/generic.py", in "findSearchResults" code, when it's parsing the showname I've changed:
myParser = NameParser(False)
To 
myParser = NameParser(showObj=show)

Original code causes that NamePaser compile all regexes available including anime  (cause show object was not being informed). Is there any reason to do this? That algorithm sometimes returns the result of an anime regexe even if show analyzed is not anime. Please check if my change is ok, otherwise I should have to overwrite "findSearchResults" on newpct provider.

Anything please let me know :)

Regards,
Cristian.

sickbeard/providers/newpct.py
```diff
@@ -0,0 +1,201 @@
...
+            size = size * 1024
+        elif modifier in 'MB':
+            size = size * 1024**2
+        elif modifier in 'GB':
+            size = size * 1024**3
+        elif modifier in 'TB':
+            size = size * 1024**4
+        return size
+
+    def _processTittle(self, title):
```
miigotu: mispelled =P

Is this quality borking absolutely necessary?

sickbeard/providers/newpct.py
```diff
@@ -0,0 +1,201 @@
...
+
+from __future__ import with_statement
+
+from sickbeard import logger
+from sickbeard import tvcache
+from sickbeard.providers import generic
+from sickbeard.common import USER_AGENT
+from sickbeard.bs4_parser import BS4Parser
+from sickbeard import scene_exceptions
+from ctypes.test.test_sizes import SizesTestCase
```
miigotu: unused import

sickbeard/providers/newpct.py
```diff
@@ -0,0 +1,201 @@
...
+# along with SickRage.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import with_statement
+
+from sickbeard import logger
+from sickbeard import tvcache
+from sickbeard.providers import generic
+from sickbeard.common import USER_AGENT
+from sickbeard.bs4_parser import BS4Parser
+from sickbeard import scene_exceptions
```
miigotu: unused import

sickbeard/providers/newpct.py
```diff
@@ -0,0 +1,201 @@
...
+#
+# SickRage is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+#  GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with SickRage.  If not, see <http://www.gnu.org/licenses/>.
+
+from __future__ import with_statement
```
miigotu: not supporting python 2.6, so no need to import with from future

sickbeard/providers/generic.py
```diff
@@ -384,7 +384,7 @@ def findSearchResults(self, show, episodes, search_mode, manualSearch=False, dow
 
             # parse the file name
             try:
-                myParser = NameParser(False)
+                myParser = NameParser(showObj=show)
```
miigotu: This is good, we should always pass the show object to NameParser if we have it.

sickbeard/providers/newpct.py
CristianBB: Ups O_O, fixed

CristianBB:
Sorry for that, fixed =)

miigotu:
Ok, everything looks pretty ok now, just conflicts and rebase. I can do that manually later.

CristianBB:
Ok miigotu, if you need me to do something let me know.

PR #3070, opened 2015-11-03
Make sure airdates are > 1 (really 693595) in webapi to stop erroring…

miigotu:
… when unaired episodes are asked for info

Fixes https://github.com/SiCKRAGETV/sickrage-issues/issues/3512


miigotu:
@MGaetan89 slight change here, not sure if it affects ShowsRage, in that if the episode hasnt aired the string in the json will be "Never" instead of "", just like webserve/mako

PR #3089, opened 2015-11-05
Show release name in history (detailed view)

fernandog:
I can't make column Time and Episode not wrap. I only want to warp "Release" name 
@OmgImAlexis help please?

@miigotu

### DIFF
--- a/gui/slick/js/new/history.js
+++ b/gui/slick/js/new/history.js
@@ -6,7 +6,7 @@ $(document).ready(function(){
             if(isMeta('sickbeard.HISTORY_LAYOUT', ['detailed'])){
                 return {
                     0: function(node) { return $(node).find('time').attr('datetime'); },
-                    4: function(node) { return $(node).find("span").text().toLowerCase(); }
+                    6: function(node) { return $(node).find("span").text().toLowerCase(); }
                 };
             } else {
                 return {
@@ -21,7 +21,7 @@ $(document).ready(function(){
             if(isMeta('sickbeard.HISTORY_LAYOUT', ['detailed'])){
                 return {
                     0: { sorter: 'realISODate' },
-                    4: { sorter: 'quality' }
+                    6: { sorter: 'quality' }
                 };
             } else {
                 return {
--- a/gui/slick/views/history.mako
+++ b/gui/slick/views/history.mako
@@ -56,6 +56,7 @@
             <tr>
                 <th class="nowrap">Time</th>
                 <th>Episode</th>
+                <th>Release</th>
                 <th>Action</th>
                 <th>Provider</th>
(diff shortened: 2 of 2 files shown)


SiCKRAGEB0T:

Thanks!

gui/slick/js/new/history.js
```diff
@@ -6,7 +6,7 @@ $(document).ready(function(){
             if(isMeta('sickbeard.HISTORY_LAYOUT', ['detailed'])){
                 return {
                     0: function(node) { return $(node).find('time').attr('datetime'); },
-                    4: function(node) { return $(node).find("span").text().toLowerCase(); }
+                    6: function(node) { return $(node).find("span").text().toLowerCase(); }
                 };
```
fernandog: @OmgImAlexis why I need to change to 6 instead of 5 ? I only added one column

fernandog:
compact mode is broken

gui/slick/js/new/history.js
fernandog: @alisonatwork  would you please help me with this? Also JS is not rendering compact mode with this PR.

alisonatwork:
I think the reason why the table cells are wrapping is because there is no CSS selector that matches the table cells. We could globally nowrap time by adding a new style ```time { whitespace: nowrap; }```, alternatively we could add classes to all of these cells. If you look at the show list table, there are tons of classes all over the place on the ```<td>``` elements, most of them are ```nowrap```.

Best practice would actually be to use semantic classes on our table cells, so not ```nowrap``` but actually classes that described the content (e.g. ```col-airdate``` which already has a nowrap style attached to it in style.css).

Regarding the mobile/compact view, this looks fine to me on Windows phone and Chrome snapped to 1/4 of my desktop monitor, although you do end up with a sideways scrollbar, which is a bit crappy for mobile. We could probably fix this by adding a responsible class on the table itself to set the max-width in pixels or something.

If you want to wrap the filenames, which have dots in them, that is a bit harder. By default it will only wrap on spaces and dashes, so you will need to force that column skinnier and maybe have ```text-overflow: ellipsis``` to basically cut it off when it gets too long. I am not really a CSS expert so i can't solve it tonight :) But this might get you started.

gui/slick/js/new/history.js
alisonatwork: This 6/5 thing totally mystifies me, it doesn't make sense. I need to look at this when it's not almost midnight :D

alisonatwork:
On a side note, if we care about massively huge tables, we should probably style using colgroup/col elements instead of td elements. This is more of a problem for websites you hit up over the web and not on the LAN, but it might be worth considering if we are going to re-style all the columns anyway.

PR #3091, opened 2015-11-05
Fix anime parsing

Noktam:
Sometime, a show is parsed without knowing that's an anime.
That can lead to errors, for exemple : 
Show.Name.102.Source.Quality.Etc-Group 

If parsed as ALL_REGEX it will be validate as a TV show S01E02
If parsed as Anime, it will be validate as a Anime with absolute number : Episode 102

So, when i parse something as ALL_REGEX, if the result is a show who have is_anime == true, it will restart the parsing as anime.

Because most of the time, even if the TV episode number is wrong, it will at least match the episode name.

I just made a little change into regex for the no_season to have a better match of the episode name (for some case when the name start with [XXXX]Show_Name, to not include [XXXX] in the show name.

### DIFF
--- a/sickbeard/name_parser/regexes.py
+++ b/sickbeard/name_parser/regexes.py
@@ -187,7 +187,7 @@ normal_regexes = [
      # 01 - Ep Name
      # 01 - Ep Name
      r'''
-     ^((?P<series_name>.+?)(?:[. _-]{2,}|[. _]))?    # Show_Name and separator
+     ^((\[(.*?)\])[. _-])?((?P<series_name>.+?)(?:[. _-]{2,}|[. _]))?    # Show_Name and separator
      (?P<ep_num>\d{1,3})                             # 02
      (?:-(?P<extra_ep_num>\d{1,3}))*                 # -03-04-05 etc
      \s?of?\s?\d{1,3}?                               # of joiner (with or without spaces) and series total ep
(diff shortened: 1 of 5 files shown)


SiCKRAGEB0T:

Thanks!

sickbeard/name_parser/regexes.py
```diff
@@ -187,7 +187,7 @@
      # 01 - Ep Name
      # 01 - Ep Name
      r'''
-     ^((?P<series_name>.+?)(?:[. _-]{2,}|[. _]))?    # Show_Name and separator
+     ^((\[(.*?)\])[. _-])?((?P<series_name>.+?)(?:[. _-]{2,}|[. _]))?    # Show_Name and separator
```
miigotu: Missing a * here. and two sets of extra parens, double check.
`(\[.*?\][. _-]*)?`

so 
`^(\[.*?\][. _-]*)?((?P<series_name>.+?)(?:[. _-]{2,}|[. _]))?`

sickbeard/name_parser/regexes.py
```diff
@@ -426,6 +426,11 @@
      r'''
      ^(?P<ep_ab_num>\d{3,4})(-(?P<extra_ab_ep_num>\d{3,4}))?\.\s+(?P<series_name>.+?)\s-\s.*
      '''),
+     
+    ('anime_MangaOff',
+     # 003. Show Name - Ep Name.ext
+     # 003-004. Show Name - Ep Name.ext
+     ur'(?P<series_name>.+?)[ ._-]+(E(?P<ep_ab_num>\d{1,4})[ ._-]+)(.*)FANSUB[ ._-]+VOSTFR(.*?)([ ._-](\d{1,4}[pi])[ ._-])?(.*)[ ._-](?P<release_group>.*)'),
```
miigotu: forcing this to only match vostfr... thats a little unilateral

Noktam: Well, i'm far from a regex expert, so i just wanted to catch some shitty release name on mangaoff.cc without risking to mess with the rest.

CristianBB:
Hi @Noktam 
Same problem using ALL_REGEX occurs with no anime shows. I tried to fix it passing original show object to NameParser (generyc.py) but as a result we obtain a lot of invalid results cause of how SR validates every result from providers (there is an explain of how SR compare results on https://github.com/SiCKRAGETV/SickRage/commit/41c26bc27ef2c831f56d5b7ed4fdaa2942869ab4 ).
There is a small change on my fork (https://github.com/SiCKRAGETV/SickRage/compare/develop...CristianBB:develop ) that should fix that problem and only should parse each show with correct regex (anime regex for animes, normal regex for normal shows). Would be nice if you could prove it and tell me if everything works ok, Could you make me that favour? I think this would be a more "generic" way to fix our problems with each type of show. I want no make the Pull Request until we're sure that works properly before that changes destroys something :P 

Regards,
Cristian.

miigotu:

PR #3098, opened 2015-11-05
add aria-haspopup to force ie11/edge to emulate hover on these elemen…

alisonatwork:
…ts, also add superfluous span so that bootstrap-hover-dropdown correctly attaches the events

Okay, i think i've finally got it this time. I've added aria-haspopup="true", which makes IE11 and Edge emulate a hover on touchscreens, so the menu pops up properly. See also http://www.w3.org/TR/wai-aria/states_and_properties The one weird thing is i had to add a superfluous span around the text inside the anchor for the anchors where there wasn't already a span or image. I have a sinking feeling this might be a bug/quirk in bootstrap-hover-dropdown, but trying to figure out what is going on there is a pretty dark hole. Tested on every browser i could get my hands on, in portrait, landscape and mobile, and it seems to work great everywhere.

Note this will probably conflict with my other pull request.

### DIFF
--- a/gui/slick/views/layouts/main.mako
+++ b/gui/slick/views/layouts/main.mako
@@ -109,7 +109,7 @@
                 <div class="collapse navbar-collapse" id="bs-example-navbar-collapse-1">
                     <ul class="nav navbar-nav navbar-right">
                         <li id="NAVhome" class="navbar-split dropdown${('', ' active')[topmenu == 'home']}">
-                            <a href="${srRoot}/home/" class="dropdown-toggle" data-toggle="dropdown" data-hover="dropdown">Shows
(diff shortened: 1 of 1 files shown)


miigotu:
Conflicts =P

alisonatwork:
Gimme a sec ;)

alisonatwork:
OK looks good. When you guys do the bootstrap 4 update, gimme a ping, because it might be that some of these hacks are not needed any more. In particular, i think the aria command should ideally be added by bootstrap and not by us. The superfluous span has something to do with events that i am not enough of a JS guru to understand. I tried moving around a bunch of toggle/hover activations, but what worked on one browser would break on the other one and this seems to be the only one that works everywhere.

PR #3108, opened 2015-11-06
Convert StatusStrings to a dict subclass

labrys:
@miigotu This is a fairly significant change, so should probably get some extra TLC before merging as it will eventually deprecate the old way of doing status strings. Would appreciate some further testing and input before merging this.  I also foresee doing something similar with Quality. 

As it currently exists, though, its should mimic the current `StatusStrings`.  Keys that can't convert to int  raise `ValueError` to match current functionality, but I would like to change that to `KeyError`.   Testing `key in dict` which should now work correctly, except for a key of `None` which raises a `TypeError` (it should check for None as a possible key).

### DIFF
--- a/sickbeard/common.py
+++ b/sickbeard/common.py
@@ -22,6 +22,7 @@ import operator
 import platform
 import re
 import uuid
+from UserDict import UserDict
 
 from random import shuffle
 
@@ -512,7 +513,7 @@ qualityPresetStrings = {SD: "SD",
                         ANY: "Any"}
 
 
-class StatusStrings(dict):
+class StatusStrings(UserDict):
     """
     Dictionary containing strings for status codes
 
@@ -526,23 +527,13 @@ class StatusStrings(dict):
     # todo: Switch from raising ValueError to a saner KeyError
     # todo: Raise KeyError when unable to resolve a missing key instead of returning ''
     # todo: Make key of None match dict() functionality
-    def __init__(self):
-        """
-        Initialize status strings
-        """
-        self[UNKNOWN] = "Unknown"
-        self[UNAIRED] = "Unaired"
-        self[SNATCHED] = "Snatched"
-        self[DOWNLOADED] = "Downloaded"
-        self[SKIPPED] = "Skipped"
-        self[SNATCHED_PROPER] = "Snatched (Proper)"
-        self[WANTED] = "Wanted"
-        self[ARCHIVED] = "Archived"
-        self[IGNORED] = "Ignored"
-        self[SUBTITLED] = "Subtitled"
-        self[FAILED] = "Failed"
-        self[SNATCHED_BEST] = "Snatched (Best)"
-        self.statusStrings = dict(self)  # for backwards compatibility
+
+    @property
+    def statusStrings(self):  # for backwards compatibility
+        return self.data
+
+    def __setitem__(self, key, value):
+        self.data[int(key)] = value  # make sure all keys being assigned values are ints
 
     def __missing__(self, key):
         """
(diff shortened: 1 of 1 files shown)


SiCKRAGEB0T:

Thanks!

miigotu:
I'm going to look over this carefully, because it's a pretty major change and this is a fairly load bearing core class.

labrys:
Exactly, that's why I made sure to call you out for it.

labrys:
Doing a similar change to `Quality` should improve the performance for `StatusString` too, since it could then be streamlined further, however, one change at a time :P  It's been running fine so far for me, across 4 installs, but that's too limited of a test for such a large change.

labrys:
Initial tests with `-mtimeit` shows the new version is slightly faster, but as small as the dict is, its pretty much inconsequential (approx 0.003 usecs difference in my tests)

duramato:
@labrys every second counts :P ,well microseconds
Good job tho ;)
keep up the good work

labrys:
Actually it looks like there may be a better way to do this, among other things want to try to avoid check for `in self.keys()`.  I'll probably play around with it some more this weekend.  As always, comments and suggestions are welcome :)

sickbeard/common.py
```diff
@@ -512,39 +512,80 @@ def statusFromName(name, assume=True, anime=False):
...
+    def __missing__(self, key):
+        """
+        If the key is not found, search for the missing key in qualities
+
+        Keys must be convertible to int or a ValueError will be raised.  This is intentional to match old functionality until
+        the old StatusStrings is fully deprecated, then we will raise a KeyError instead, where appropriate.
+        """
+        key = int(key)  # This will raise a ValueError if the key can't be converted to int
+
+        if key in self.keys() + Quality.DOWNLOADED + Quality.SNATCHED + Quality.SNATCHED_PROPER + Quality.SNATCHED_BEST + Quality.ARCHIVED:
```
miigotu: we already know it isnt in self.keys() here, or __missing__ wouldnt have been called

labrys:
Actually that check is to catch valid strings as keys.   For example '1' (Unaired) would fail without the check since the key is 1.

labrys:
Switching to `UserDict` instead of `dict` cleans this up nicely.  I'll test this some more and check performance, but so far this looks like a much better way to go.

miigotu:
This scares me, it's gonna be a day or two before I get the time to test it in an interpreter.

labrys:
No problem.

miigotu:
so this seems to be ok, really should raise a KeyError when an invalid status is attempted (there should never be a case where it happens)

labrys:
Yeah I only left it as `ValueError` in case something was expecting it.

miigotu:
It returns `''` atm if you pass a number not in the dict =P

```python
import sys;sys.path.insert(1,'lib')
from sickbeard.common import statusStrings
print statusStrings[999999]
''
```

labrys:
That was intentional to match the previous implementation.  

I'm currently switching over to raise `KeyErrors` instead of `ValueErrors` and will make missing keys raise a `KeyError` as well.  In addition it will be raising `TypeErrors` when trying to assign a value to an invalid key.

If testing for 'key in dict' with an invalid type it was throwing TypeErrors.  It shouldn't throw anything.

More importantly though, if they key is a really large number, it gets caught in a recursive loop.  Changing from `int(key)` to `long(key)` prevents the recursion.

labrys:
I should have a new version ready shortly, just putting it through some tests first.

miigotu:
k, Ill look at it after the cowboys game later =P

labrys:
@miigotu Changes are ready. PR #3155 

A (really really ugly) test for it is below... (and i mean ugly)
```python
from common import statusStrings as ss

key_errors = [
    122, '122', None,
    'Elephant', 'Mouse', 'Frog',
    90198520984509128409, '90198520984509128409',
    (1, 'Armadillo', )
]

type_errors = (
    'Elephant',
    (1, 12),
    {},
    [],
    set()
)

print 'Prints the data'
print ss.statusStrings
print ss.data
print
print 'Prints the appropriate string'
print (ss[1], ss['1'])
print (ss[112], ss['112'])
print
print 'Evaluates to true'
print (ss.has_key(1), ss.has_key('1'))
print (ss.has_key(112), ss.has_key('112'))
print (1 in ss, '1' in ss)
print (112 in ss, '112' in ss)
print
print 'Evaluates to false'
print (ss.has_key(122), ss.has_key('122'))
print (ss.has_key(90198520984509128409), ss.has_key('90198520984509128409'))
print (ss.has_key(None), ss.has_key(str(None)), ss.has_key('Elephant'))
print (122 in ss, '122' in ss)
print (90198520984509128409 in ss, '90198520984509128409' in ss)
print (None in ss, 'None' in ss, 'Elephant' in ss, (1, 2, 3) in ss, [12, 'frog', 'kangaroo'] in ss)
print
print 'Raise KeyErrors for %s' % str(key_errors)
for ke in key_errors:
    try:
        ss[ke]
    except KeyError:
        print "KeyError raised for ss[%s]" % str(ke)
    except Exception:
        print "Failed!"
print
print 'Raise TypeErrors for %s' % str(type_errors)
for te in type_errors:
    try:
        ss[te] = 'test'
        print "!!! This shouldn't have worked! %s" % str(te)
    except TypeError:
        print "TypeError raised for ss[%s] = 'test'" % str(te)
print
print 'Assignment works fine'
ss[None] = 'Success!'
print ss[None]
ss[122] = 'Cornflakes'
print (ss[122], ss['122'])
ss['122'] = 'Cheerios'
print (ss['122'], ss[122])
ss['896798797970897907'] = [1, 23, 41, 'Cowboys']
print (ss['896798797970897907'], ss[896798797970897907])

print
print 'Evaluates to True'
print (896798797970897907 in ss, None in ss, '122' in ss)
print (ss.has_key('896798797970897907'), ss.has_key(None), ss.has_key('122'))
print
print 'Print the data again'
print ss.statusStrings
print ss.data
print ss.viewkeys()
print ss.viewitems()
print ss.viewvalues()
```

labrys:
Time to sleep now :sleeping: :zzz:

PR #3155, opened 2015-11-08
Fix recursion in status strings and raise appropriate KeyErrors...

labrys:
…TypeErrors

-Switched to long() conversion to avoid recursion when numeric key > int.  
-Raising appropriate TypeErrors and KeyErrors.
-Added views

### DIFF
--- a/sickbeard/common.py
+++ b/sickbeard/common.py
@@ -516,69 +516,92 @@ qualityPresetStrings = {SD: "SD",
 class StatusStrings(UserDict):
     """
     Dictionary containing strings for status codes
-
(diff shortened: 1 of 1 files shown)


labrys:
Tested it against master and it appears to work well.  Dev is too broke for me to test... shows don't display anything since the encoding hack was removed.

miigotu:
We should probably add some strong unit testing for this

fernandog:
@miigot what about this one? merge or need further tests/changes?

labrys:
I've been running this with no issues for a while now, so in my opinion it should be fine to merge.

I'm working on streamlining it, making it Python 3+ compatible, and adding unit tests if you want to wait, though.

fernandog:
Nice!

PR #3164, opened 2015-11-09
Encode long type as int

gaganpreet:
Adding torrents with file sizes greater than `sys.maxint` to rtorrent fails with "return code undefined" when using SickRage with Python 2.7 (on 32-bit systems especially) because the bencode module didn't encode the long data type. The torrent would get added to rtorrent, but SickRage wouldn't get the correct hash from bencode making it seem that adding the torrent failed. This fix encodes the long type the same as int.

Possibly also fixes SiCKRAGETV/sickrage-issues#1105

### DIFF
--- a/lib/rtorrent/lib/bencode.py
+++ b/lib/rtorrent/lib/bencode.py
@@ -42,8 +42,10 @@ _py3 = sys.version_info[0] == 3
 
 if _py3:
     _VALID_STRING_TYPES = (str,)
+    _VALID_INT_TYPES = (int,)
 else:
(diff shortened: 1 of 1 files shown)


SiCKRAGEB0T:

Thanks!

miigotu:
Shouldn't this be done upstream? Really not sure where ours comes from and too lazy to look =P

Either way I will merge it, I see no real harm

WebSpider:
This was added to the TPB fork about two years ago by @jkaberg ... Trying to trace it a bit further

PR #3169, opened 2015-11-10
Assure we are converting args to unicode where the data enters the ap…

miigotu:
…plication, as soon as possible. The asserts can be removed after all unicode problems are fixed

### DIFF
--- a/sickbeard/helpers.py
+++ b/sickbeard/helpers.py
@@ -1912,3 +1912,42 @@ def getDiskSpaceUsage(diskPath=None):
             return pretty_filesize(st.f_bavail * st.f_frsize)
     else:
         return False
+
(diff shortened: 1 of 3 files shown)


miigotu:
@echel0n if we need it more global we already has def _to_unicode(var): in sickrage.helper.encoding

miigotu:
also, why cant we:

```python
from sickrage.helper.encoding import _to_unicode
def oswalk(top, topdown=True, onerror=None, followlinks=False):
    for _path, _dirs, _files in os.walk(top, topdown, onerror, followlinks):
        yield  _to_unicode(_path), [_to_unicode(_dir) for _dir in _dirs], [_to_unicode(_file) for _file in _files]
```

PR #3170, opened 2015-11-10
Only use the exception logger for ERROR level (maybe we should use lo…

miigotu:
…gger.warn() too xD)

Fixes https://github.com/SiCKRAGETV/sickrage-issues/issues/3609


fernandog:
thanks!

PR #3179, opened 2015-11-13
Change pretty name to S01E01 format

fernandog:
@echel0n 
This affects log lines that shows 1x01 and users think it's not searching using S01E01 pattern


OmgImAlexis:
This should really use the format they have selected then? What about the people that have it renaming it 1x01 format? They'll now think it's incorrect?

PR #3182, opened 2015-11-15
Convert statusstrings to a NumDict

labrys:
##### Code Changes:
- Change from UserDict to NumDict and add unittests for NumDict and new statusstrings
- Streamline common tests and PEPs

This fixes issues with recursion in the UserDict implementation of statusstrings, adds the unit tests required, is Python 3 compatible, and NumDict can be further leveraged to streamline common.Quality in the future.


labrys:
Replaces PR #3155

PR #3184, opened 2015-11-17
Save multiple plex servers

disrani:
Multiple plex servers can be provided in the web UI and the test also uses them. But due to the bug only one server would get saved.

### DIFF
--- a/sickbeard/webserve.py
+++ b/sickbeard/webserve.py
@@ -4673,7 +4673,7 @@ class ConfigNotifications(Config):
         sickbeard.PLEX_NOTIFY_ONSUBTITLEDOWNLOAD = config.checkbox_to_value(plex_notify_onsubtitledownload)
         sickbeard.PLEX_UPDATE_LIBRARY = config.checkbox_to_value(plex_update_library)
         sickbeard.PLEX_HOST = config.clean_hosts(plex_host)
-        sickbeard.PLEX_SERVER_HOST = config.clean_host(plex_server_host)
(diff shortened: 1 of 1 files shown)


sickbeard/webserve.py
```diff
@@ -4673,7 +4673,7 @@ def saveNotifications(self, use_kodi=None, kodi_always_on=None, kodi_notify_onsn
         sickbeard.PLEX_NOTIFY_ONSUBTITLEDOWNLOAD = config.checkbox_to_value(plex_notify_onsubtitledownload)
         sickbeard.PLEX_UPDATE_LIBRARY = config.checkbox_to_value(plex_update_library)
         sickbeard.PLEX_HOST = config.clean_hosts(plex_host)
-        sickbeard.PLEX_SERVER_HOST = config.clean_host(plex_server_host)
+        sickbeard.PLEX_SERVER_HOST = config.clean_hosts(plex_server_host)
         sickbeard.PLEX_SERVER_TOKEN = config.clean_host(plex_server_token)
```
OmgImAlexis: @miigotu would the tokens not need to use `clean_hosts` too?

miigotu: I think it would take more than just that change to make it work for multi-host, should be tested.

disrani: All my Plex servers got updated after a recent download with this change. Also the APIs which take in the host to update Plex already expect a comma separated list.

disrani: And the UI also already mentions that multiple hosts can be provided. The "Test" button tests all the servers too.

miigotu:

PR #3188, opened 2015-11-24
Revert using is not because it should be not equal

sickraged:
`is` is for identity checks not for checking equality.


labrys:
While it is true that is should be used when comparing equality, comparisons to singletons such as None, True and False should be done using is / is not.  

##### From [PEP 0008](https://www.python.org/dev/peps/pep-0008/#programming-recommendations):

Comparisons to singletons like None should always be done with is or is not , never the equality operators.

Also, beware of writing if x when you really mean if x is not None -- e.g. when testing whether a variable or argument that defaults to None was set to some other value. The other value might have a type (such as a container) that could be false in a boolean context!

PR #3192, opened 2015-11-24
Fix readme.md warning typo

HLFH:


labrys:
Duplicate #3190

Arcanemagus:
Not really, this one fixes two typos in that message while the other only fixes one.

labrys:
true, but the other one has a funny subject line :)

PR #3200, opened 2015-12-03
Fix apostrophe show name search

freedomisforhippies:
For shows containing apostrophe or "'s", the search is also made with those removed from the name. Seems to result in better results...


Arcanemagus:
The PR should also be rebased against develop, although it looks like @echel0n has diverged the branches so I'm not sure what is going on there...

PR #3203, opened 2015-12-05
Removed nxtgn provider

calvin--:
The tracker NextGen (nxtgn.biz) has shut down.

https://nxtgn.biz
https://twitter.com/nxtgnorg


echel0n:
Thanks

PR #3204, opened 2015-12-05
libs: moving libs into requirements.txt

satreix:


satreix:
Why did you close that one without merge or explaination ? :cry:

Arcanemagus:
I'm guessing it was because this was against the master branch instead of dev, but it's @echel0n so who really knows.

Arcanemagus:
You might be interested in the fork _all_ of the other developers moved to: https://github.com/SickRage/SickRage

It's actually active unlike this codebase :wink:.

Wally73:
that isn't @echel0n 's branch is that fake one from former devs

satreix:
I heard roge devs were doing stuff on there own yes. @echel0n, was this PR closed only because it was not on the _dev_ branch?

satreix:
Do you have any reason to keep this projects requirements in _/libs_ instead of using `pip` with a _requirements.txt_ file to pull the required dependencies from PyPI ?

I would like to clear out any unnecessary code from the repository. :smile:

PR #3214, opened 2015-12-22
Fix API returning error on help

tresni:
Attempting to pass `args` and `kwargs` without the preceeding `*` && `**` is broken. If you don't have it then they are just additional arguments that get passed to the special `*args` variable of the called function.  This was causing `kwargs` to get mangled into `args` and then leading to an error about how it couldn't parse the next param.  You can see this by doing `/api/[yourkey]/?cmd=show.addnew&help=1`.

I am sure this was causing other errors too, but help was the first one I saw :)

### DIFF
--- a/sickbeard/webapi.py
+++ b/sickbeard/webapi.py
@@ -702,8 +702,8 @@ class CMD_Help(ApiCall):
     def __init__(self, application, request, *args, **kwargs):
         # required
         # optional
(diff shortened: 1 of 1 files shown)


darklow:
@tresni Could this be related to this bug I reported here as well?
Something with a kwargs and wrong arguments:
https://sickrage.tv/forums/forum/help-support/bug-issue-reports/32368-webapi-show-getposter-not-working-returns-error

PR #3223, opened 2015-12-30
Set the `gui_root` inside setUp_test_web_server.

sickraged:


Djang0:
I fixed this in https://github.com/Djang0/SickRage/commit/fba190cb92d5e4885ab3f7ee40c5ad9e00a2debb as well ;)

PR #3224, opened 2015-12-30
Fix Allowed and Preferred Quality selections

mpcaddy:
Lists now get passed through so that all selected qualities from the Allowed and Preferred selections will now get saved instead of just the first selected item in each

### DIFF
--- a/sickbeard/webserve.py
+++ b/sickbeard/webserve.py
@@ -217,7 +217,8 @@ class WebHandler(BaseHandler):
     def async_call(self, function, threadName):
         threading.currentThread().name = threadName
         try:
-            kwargs = recursive_unicode({k: self.request.arguments[k][0] for k in self.request.arguments if len(self.request.arguments[k])})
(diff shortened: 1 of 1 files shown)


Djang0:
Hi,
If you want travis check to succeed, you should refer to this commit : https://github.com/Djang0/SickRage/commit/fba190cb92d5e4885ab3f7ee40c5ad9e00a2debb
and merge it to your own fork, then re submitting the pull request.

neclimdul:
:+1: Fixes saving.

Doesn't fix edit not reflecting the saved values but that can be a follow up.

JBFUK:
Fixed two issues for me:

1. Quality selection saving
2. Add existing shows would not add multiple shows, it would add the first and quit back to the shows listing.  Now behaves as expected.

Thanks.

echel0n:
New update to DEV branch should have this corrected but if not please perform the PR against that as its a entirely new system then what master is using and will being replacing master soon, thanks!

echel0n:
Ok I see this issue is still present so I'm correcting it now and putting it onto dev branch.

PR #3212, opened 2015-12-19
Random 403 : user_agents list was broken

gboddin:
All HTTP(s) requests might be affected :

 - Commas were missing in the user_agents list, python is treating it as a full string
 - As a result, providers/generic.py uses a user_agent which is too long and causes 403 on providers behind cloudflare (for instance)

Adding the missing commas fixes the issue.

PR #3227, opened 2016-01-02
HOTFIX User agent error causing 403s

ograff:
Updated user_agents array with commas that were left out originally
This fixes an issue where some servers were sending 403s because the user agent
was too many chars

#3212 was a PR to fix this but was closed by the author due to Travis Failing


ograff:
@echel0n Are the unit tests expected to work? This hung after "OK"

PR #3231, opened 2016-01-05
Fix thumbnail URLs when they reside in a directory outside the checkout

pcuenca:
The problem was introduced in 69ee32339 as part of the static image file handling.


pcuenca:
This isn't ready yet: it will fail if the web root path configuration is empty. I'll fix it tomorrow.

PR #3233, opened 2016-01-06
Add pip support

satreix:
This setup a _venv_ directory to start using `pip` to manage Python packages instead of doing it by hand. This _venv_ is effectively a py2 venv created with `virtualenv`. It also add a couple lines in SickBeard.py to support this venv on platforms that do not (extend the `PYTHONPATH`).

We should also add that to the documentation:

- setup a venv
  - virtualenv -p python2 venv
  - source venv/bin/activate
  - pip install requirements.txt

This is a working draft to see what the community thinks about the idea, as well as the excution.


echel0n:
new branch has this done already :)

PR #3240, opened 2016-01-10
Updates 122415 fixes

neclimdul:
The beta is _really_ broken. These probably aren't all the right changes, but it got things working.


neclimdul:
one additional thing, on my docker image I have to make sure configobj is installed manually. It might need an entry in requirements.txt

PR #3244, opened 2016-01-22
Upgrading to https scheme

cgoconseils:


echel0n:
lost on what this even is ....

PR #3247, opened 2016-01-26
If a virtualenv is provided for upstart then use it.

sickraged:


echel0n:
Added this PR to latest version, thanks!

PR #3248, opened 2016-01-26
The ordering of providers was being saved incorrectly because 'enable…

sickraged:
…d' providers were simply being placed at the start of the provider priority list.


echel0n:
Enabled providers are always placed first so how does this correct that if anything it appears this will break the ordering, please elaborate more on your fix.

sickraged:
@echel0n what happens is that when providers are saved - the "lowest" priority active provider ends up at the front of the list because of the `providers_reordered.insert(0, curProvider)` call. This change fixes the issue by placing active providers in a separate list and then placing those ahead of inactive providers but with the ordering of the active providers maintained (as well as the ordering of the inactive providers).

PR #3249, opened 2016-02-04
Changed heading to match what the autoProcessTV.py is looking for

danjng:
Changed heading from sickrage to SiCKRAGE so that it matches what autoProcessTV.py is searching for in the lookups area.


echel0n:
PR merged into next version, thanks!

PR #3250, opened 2016-02-14
The ordering of providers was being saved incorrectly because 'enable…

sickraged:
…d' providers were simply being placed at the start of the provider priority list.

What happens is that when providers are saved - the "lowest" priority active provider ends up at the front of the list because of the `providers_reordered.insert(0, curProvider)` call. This change fixes the issue by placing active providers in a separate list and then placing those ahead of inactive providers but with the ordering of the active providers maintained (as well as the ordering of the inactive providers).

### DIFF
--- a/sickrage/core/webserver/views.py
+++ b/sickrage/core/webserver/views.py
@@ -4371,7 +4371,8 @@ class ConfigProviders(Config):
                 zip([x.id for x in sickrage.torrentRssProviderList], sickrage.torrentRssProviderList))
 
         # do the enable/disable
-        providers_reordered = []
+        enabled_providers = []
+        disabled_providers = []
         sorted_providers = sortedProviderDict()
         for curProviderStr in provider_str_list:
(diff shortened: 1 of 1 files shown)


sickraged:
This follows on from 3248. 

cc @echel0n .

sickrage/core/webserver/views.py
```diff
@@ -4381,9 +4382,9 @@ def saveProviders(self, newznab_string='', torrentrss_string='', provider_order=
                 curProvObj = sorted_providers[curProvider]
                 curProvObj.enabled = curEnabled
                 if curEnabled:
-                    providers_reordered.insert(0, curProvider)
```
sickraged: @echel0n this line means that providers which a lower preference become the most preferred when saving providers.

vpara:
Hi there, my sickbird crashes all the time. Can someone help me with this?

PR #3252, opened 2016-03-03
Change url or T411

flemzord:
Update url for T411


bOOm-X:
It will be cool to have that ! t411 has indeed still changed of DNS.
What is wrong with this MR ?

flemzord:
This does not just me, the problem of travis scripts that are in error. cc @bOOm-X

PR #3255, opened 2016-03-05
Fix datetime.now() call in srShowUpdater

lpenz:


echel0n:
Issue corrected in v8.0-beta, thanks!

PR #3257, opened 2016-03-11
Added missing User-agent for tvdb API.

Aeronaut:
This will fix the "No results found, try a different search."...


dgburr:
Can confirm that this fix works

9ler:
Can confirm that this fix works - thanks Aeronaut :)

VietTPham:
Confirm working as well.

cbuckley:
Confirmed working

uilino:
Works!! thanks!!

PR #3258, opened 2016-03-13
Update t411.py

kaintaozia:
hi i made modification to take effect of the change of adress of t411 provider


echel0n:
This has been taken care of on v8.0-BETA, thanks.

PR #3259, opened 2016-03-20
Show only add episode that you already own

markeclaudio:
Discussed here:
http://sickrage.ca/forums/forum/help-support/bug-issue-reports/33434-show-only-add-first-season

Thanks to throwaway994 for the fix


echel0n:
This issue has been corrected via different code changes in our v8.0-BETA, thanks!

PR #3260, opened 2016-03-20
T411 moved from .in to .ch

mouftik:
Now the site has moved from *.in to *.ch so the urls need to be updated as well.


echel0n:
Version 8.0-BETA has this change applied, thanks!

PR #3262, opened 2016-03-25
Fixes for name cache

dgburr:
There are two problems with the current implementation of the name cache:

1. When srNameCache.buildNameCache() is called without arguments it only updates the first show in the list
2. The cache is not stored persistently across restarts

These problems cause backlog searches to fail, since NameParser does not get a result from sickrage.srCore.NAMECACHE.retrieveNameFromCache().  This is particularly troublesome after cache.db gets deleted since the scene_names table will be empty.


echel0n:
I've implemented this on our v8.0 beta, thanks

PR #3264, opened 2016-05-15
Fixed issue with anon_url whereby if the URI began with https:// it ...

niavasha:
added http:// + existing_url, so http://https://domain.com.... This normally
resulted in a broken link in most browsers.

Changed the logic to use re, in the existing style, and check to see whether
http or https first, then if neither, then only adding http://


niavasha:
This was evident on the http://host:port/config/notifications page where there is a link under Plex Media Server for "Finding your account token"

echel0n:
Thanks for this, it has been merged!

PR #3266, opened 2016-06-13
Fix errors in 7.0.22

ebeigarts:
Last night I updated sickrage (hadn't upgraded it since Jan 22) and it stopped working after reboot, first I had problems with `subliminal==2.0.3`, had to downgrade to `subliminal=1.1.1`, then the db upgrade didn't work (db version check and last migration was broken). Manual post-processing had runtime errors.

It now seems to be running normally. I hope you can release `7.0.23` until the next major version gets released. I also tried `develop` branch but couldn't get it to work, it had some problems with `guessit`.

Also after upgrading I had to manually run `Force Full-Update` for all my shows to avoid `Unable to parse the filename ... into a valid show`, I think it has to do something with the last migration adding empty `tmdb_id`.

### DIFF
--- a/sickrage/core/databases/__init__.py
+++ b/sickrage/core/databases/__init__.py
@@ -206,7 +206,7 @@ class Connection(object):
         """
         try:
             if self.hasTable('db_version'):
-                return self.select("SELECT db_version FROM db_version")
+                return self.select("SELECT db_version FROM db_version")[0]["db_version"]
         except:
             return 0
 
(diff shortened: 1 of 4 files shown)


ebeigarts:
Or do I have to use https://github.com/SickRage/SickRage ?

Arcanemagus:
This repo has essentially been dead since @echel0n kicked everyone out, the above is where most of the developers moved.

echel0n:
PR's are done against develop branch so please follow that and re-submit, thanks!

PR #3272, opened 2016-11-20
Develop

domibarton:
Please read and respond to my message…

I'm banging my head here against the table, because the migration from SickBeard to SiCKRAGE is getting really annoying :( I get constant HTTP 500 errors, the WebUI is slow as hell, startup of the SiCKRAGE WebUI takes ages, the indexer is running for 3 days now!

So please, activate my account in the support forums! I never got a confirmation mail!


echel0n:
There are several ways to get support from us and this is not one of them, if your having issues with logging into the forums then send a email to support@sickrage.ca please

PR #3273, opened 2016-11-22
Fix for bug in calling pushbullet methods

cortedeltimo:
The method get_pushbullet_devices() is called on postprocessing while it's defined as part of notifiers. This will cause the follwoing javascript error when the configuration page for pushbullet is open. 

 core.min.js:3 Uncaught TypeError: b.config.postprocessing.get_pushbullet_devices is not a function(…)


echel0n:
Please re-open PR again develop branch

PR #3274, opened 2016-11-22
Fix for pushbullet configuration in webserver

cortedeltimo:
In the pushbullet configuration section, the function get_pushbullet_devices() is called on postprocessing instead of notifications. This is causing the following javascript error:

core.min.js:3 Uncaught TypeError: b.config.postprocessing.get_pushbullet_devices is not a function(…)

Same for load_show_notify_lists()


echel0n:
Thank you for contributing to this project, I've updated the develop branch with the new code.

PR #3275, opened 2016-11-26
Fix for error in _sendPushbullet() due to missing requests lib

cortedeltimo:
The call to self.session.request() cause the following error: 

DEBUG::WEB::Pushbullet authorization failed with exception: 'Traceback (most recent call last):
File "/home/odroid/SiCKRAGE/sickrage/notifiers/pushbullet.py", line 93, in _sendPushbullet
response = self.session.request(method, url, data=data, headers=headers)
AttributeError: \'NoneType\' object has no attribute \'request\'

Changing that to use the srCore.srWebsession.


echel0n:
Thank you for this PR :)

PR #3277, opened 2017-04-13
Update templates to support web_root configuration

gtosh4:
Some of the resources defined in the mako templates are hard coded to use `/` web root which result in many images, css, etc to not be found when utilizing `web_root` configuration.

## Before
before

## After
after


echel0n:
could you please describe your setup to achieve the web_root outside of sickrage ? I want to replicate it on my side as I intend to remove the entire need for web_root variable all together

gtosh4:
I've got a nginx proxy running with:
```nginx
upstream sickrage {       
    server 127.0.0.1:9092;
}
[...]
server {                 
    listen 443 ssl http2;
    server_name saraswati;
[...]
    location /tv {                                                  
        proxy_set_header X-Real-IP $remote_addr;                    
        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header Host $http_host;                           
        proxy_set_header X-NginX-Proxy true;                        
        proxy_http_version 1.1;                                     
        proxy_set_header Connection "";                             
        proxy_pass_header X-Transmission-Session-Id;                
        add_header   Front-End-Https   on;                          
        proxy_bind 127.0.0.1;                                       
                                                                     
        proxy_pass http://sickrage;                                 
    }
}
```

And
```sh
$ grep 'web_root' config.ini
  web_root = /tv
```

I have found additional mako templates which I'm in the progress of fixing. It would be a shame to lose `web_root` as it makes it easier for me to set up the proxy. The idea behind the proxy is to set up https & authentication on the nginx proxy and bind sickrage to `127.0.0.1` only to force going through to proxy to non-local visitors (as is my case where sickrage is running on my media server).

PR #3287, opened 2018-01-27
Add API subdomain to usenet-crawler.com

Elehiggle:
See: https://www.usenet-crawler.com/

"26 January 2018
I have separated the api and website. This should lower the amount of open connections a lot, hopefully making the site more snappy. New url https://api.usenet-crawler.com"


echel0n:
I've merged the change into the develop branch instead, thanks!

PR #3292, opened 2018-11-18
Fixed bad merge, errant spacing

KevinAnthony:
Without this change, the latest release will not start.


echel0n:
Fix being pushed now for this

PR #3293, opened 2018-12-02
Update from yggtorrent.to to yggtorrent.gg

Karpet76:


echel0n:
No need to change the URL as I have code in place that auto-discovers the new URL

PR #3300, opened 2019-08-17
FIx not found forum URL

guilhermebruzzi:
Changed to the link that the website redirect for the Forum


echel0n:
Thanks for catching this, edit will be manually added into the readme.

PR #3304, opened 2020-12-22
Fixes IPTorrent parsers for seeders and leachers

LostDakota:


echel0n:
I'll make the code adjustments and then push them out later this week.

